PR #2018, opened 2026-04-11
chore: add 2 newly passing tests to whitelist (1103)

tokuhirom:
Add roast/S11-modules/importing.t and roast/S11-modules/nested.t


tokuhirom:
Closing — the whitelist commit incorrectly bundled unrelated role-registration changes that broke S06-currying/named.t, S06-currying/slurpy.t, S06-multi/compile-time.t and others. origin/main is clean. The 2 tests (S11-modules/importing.t, S11-modules/nested.t) need a proper investigation rather than ad-hoc patches.

PR #2028, opened 2026-04-13
fix: support custom trait_mod:<is> on class/role declarations

tokuhirom:
## Summary
- Parse `is <lowercase>(...)` on class/role declarations and dispatch to user-defined `multi trait_mod:<is>` candidates via synthesized post-decl calls.
- Persist `Type.HOW does Role[...]` mutations in a class HOW override map so subsequent `Type.HOW` lookups return the augmented metaclass.
- Rank multi candidates by named-parameter type-hierarchy distance before required-named/trait counts so `Str:D :$x` beats `Any :$x!` on a string arg.

Adds `roast/S14-traits/package.t` to the whitelist (4/4 passing).

## Test plan
- [x] `prove -e 'target/debug/mutsu' roast/S14-traits/package.t` (4/4)
- [x] `make test` (3802 tests pass)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/roast-whitelist.txt
+++ b/roast-whitelist.txt
@@ -597,6 +597,7 @@ roast/S14-roles/rw.t
 roast/S14-roles/stubs.t
 roast/S14-roles/submethods-6e.t
 roast/S14-roles/typecheck.t
+roast/S14-traits/package.t
 roast/S14-traits/variables.t
 roast/S15-literals/identifiers.t
 roast/S15-literals/numbers.t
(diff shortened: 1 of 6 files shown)


tokuhirom:
Closing — the multi-dispatch tie-break reorder regressed S11-modules/require.t and S14-roles/composition.t. The S14-traits/package.t fix needs to be redone without altering tie-break ordering.

PR #2088, opened 2026-04-14
fix: validate =begin table Pod blocks for empty content

tokuhirom:
## Summary
- Extend Pod table validation to `=begin table ... =end table` delimited blocks (previously only `=table` paragraph form was validated)
- Makes roast/S26-documentation/07b-tables.t pass all 4 subtests

## Test plan
- [x] `prove -e target/debug/mutsu roast/S26-documentation/07b-tables.t`
- [x] `make test`


tokuhirom:
Closing: my parser-level validation correctly errors on empty =begin table/=end table at parse time, but this regresses roast/S26-documentation/12-non-breaking-space.t which generates an empty table at runtime via EVAL because mutsu's BEGIN phaser at end-of-file does not run before earlier 'my' var initializers (a separate, larger bug). The existing TODO in src/parser/helpers.rs already notes this exact blocker. Marking 07b-tables as too_difficult until BEGIN ordering is fixed.

PR #2122, opened 2026-04-15
Defer roast temp.t (raku itself fails 8/46)

tokuhirom:
## Summary
- Raku itself fails 8 of 46 subtests in roast/S04-blocks-and-statements/temp.t (TEMP block / .TEMP method semantics warn "Use of Nil in string context" and the .TEMP method is never actually invoked on temporization/restoration under current Rakudo).
- Since the test cannot pass on raku, it cannot be made to pass completely on mutsu. Recording the blocker in too_difficult.txt.

## Test plan
- [x] Verified raku fails 8 subtests on this file.


tokuhirom:
Closing stale PR — changes already covered by later merges or too conflicted to rebase.

PR #2123, opened 2026-04-15
chore: defer S10-packages/require-and-use--dead-file.t

tokuhirom:
## Summary
- raku itself fails this roast test: current Rakudo aborts after 2 of 18 planned subtests with "Cannot convert string to number", both failing.
- Since it cannot pass on raku, it cannot pass on mutsu. Record in too_difficult.txt per workflow.

## Test plan
- [x] Ran `raku roast/S10-packages/require-and-use--dead-file.t` and confirmed it fails on raku.


tokuhirom:
Closing stale PR — changes already covered by later merges or too conflicted to rebase.

PR #2140, opened 2026-04-16
fix: improve attribute trait handling, type-object guards, and accessor metadata

tokuhirom:
## Summary
- Parse `is rw`/`is readonly`/`is required` on grouped `has` declarations (`has ($.x, $.y) is rw`)
- Detect unknown `is`/`will` traits on attributes and throw `X::Comp::Trait::Unknown` with proper metadata (type, subtype, declaring)
- Guard against modifying instance attributes (`$!attr`) when the invocant is a type object (no instance)
- Implement `.rw` method on Sub/Routine values so `A.^lookup("x").rw` returns the correct boolean
- Make `^lookup` find public attribute accessors and return Sub with correct `is_rw` flag
- Preserve itemized arrays (`$[...]`) in attribute initialization via `.new()`
- Use `coerce_attr_value_by_sigil` in `dispatch_bless` for consistent attribute value coercion

Reduces roast/S12-attributes/instance.t failures from 17 to 11 (6 non-TODO tests fixed). Remaining failures are deep architectural issues (accessor returns copy so push/type-enforcement don't persist, private attribute backdoor via EVAL, scalar container semantics).

## Test plan
- [x] `make test` passes (all 471 test files, 3802 subtests)
- [x] `make roast` passes (buf.t failure is pre-existing)
- [x] `cargo fmt` applied

### DIFF
--- a/TODO_roast/S12.md
+++ b/TODO_roast/S12.md
@@ -9,6 +9,7 @@
 - [ ] roast/S12-attributes/delegation.t
 - [ ] roast/S12-attributes/inheritance.t
 - [ ] roast/S12-attributes/instance.t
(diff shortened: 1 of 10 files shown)


tokuhirom:
Closing — will re-read diff and re-implement on fresh branch if needed.

PR #2157, opened 2026-04-27
fix: pass all 181 subtests in roast/S03-operators/range.t

tokuhirom:
## Summary
- Fix `lives-ok`/`dies-ok` leaking `$_` into caller scope by saving/restoring both `$_` and `_` (bare name used by the VM topic variable)
- Implement Range arithmetic (`Range +/- Real`, `Real + Range`, `Range * Real`, `Real * Range`, `Range / Real`) preserving exclusivity flags and Mixin wrappers
- Add `X::Worry::Precedence::Range` compile-time error for ambiguous `|`/`~` prefix before range operators
- Implement `R..` `R..^` `R^..` `R^..^` reverse-range meta-operators in the parser and runtime
- Add `roast/S03-operators/range.t` to whitelist

## Test plan
- [x] `prove -e 'target/debug/mutsu' roast/S03-operators/range.t` passes all 181 subtests
- [x] `cargo fmt -- --check` clean
- [x] `make test` passes (only pre-existing flaky `t/lock.t` failure)


tokuhirom:
CI failed — S02-literals/quoting.t crash regression. Needs investigation.

PR #2234, opened 2026-05-08
fix: collect batched values from downstream supplier in tap-ok

tokuhirom:
## Summary

- Fixed tap-ok to collect values from the downstream supplier's value store (via `supplier_snapshot`) instead of `supply_emit_buffer` when testing supplier-backed supplies with `:after-tap` callbacks
- This fixes time-based `Supply.batch(:seconds)` tests where raw unbatched values were being compared against expected batched output
- All 9 subtests in `roast/S17-supply/batch.t` now pass (previously 5/9)
- Test cannot be added to whitelist due to 38s runtime exceeding 30s CI timeout

## Test plan

- [x] `prove -e 'target/debug/mutsu' roast/S17-supply/batch.t` passes (9/9)
- [x] All whitelisted Supply tests verified (unique, squish, elems, produce, reduce, do, lines, words, start, flat, head, grep, map, sort, from-list)
- [x] `cargo fmt` and `cargo clippy -- -D warnings` pass
- [x] `make test` passes (lock.t flaky failure is pre-existing race condition)

### DIFF
--- a/src/builtins/methods_0arg/dispatch_core_math.rs
+++ b/src/builtins/methods_0arg/dispatch_core_math.rs
@@ -497,6 +497,9 @@ pub(super) fn dispatch(
         "item" => Some(match target {
             Value::Array(items, kind) => Some(Ok(Value::Array(items.clone(), kind.itemize()))),
             Value::LazyList(_) => None, // fall through to runtime to force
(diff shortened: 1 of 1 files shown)


tokuhirom:
Closing — roast regression in map/when tests after rebase. Will revisit.

PR #2247, opened 2026-05-08
fix: improve eqv operator for Set/Bag/Mix types and lazy sequences

tokuhirom:
## Summary

- **Set/Bag/Mix eqv type checking**: `eqv` now correctly distinguishes between Set vs SetHash, Bag vs BagHash, Mix vs MixHash by checking the mutability flag
- **Set/Bag/Mix eqv typed key comparison**: `eqv` on sets now compares original typed keys, so `set(<42>) eqv set(42)` correctly returns False (IntStr vs Int)
- **Lazy sequence eqv**: throws `X::Cannot::Lazy` when both operands are lazy and the same type category
- Pre-checks laziness before `eval_binary_with_junctions` forces LazyList to Seq

Improves `roast/S03-operators/eqv.t` from 62/64 to 63/64 passing subtests.

## Test plan
- [x] `roast/S03-operators/eqv.t` passes 63/64 (was 62/64)
- [x] `make test` passes

Generated with [Claude Code](https://claude.com/claude-code)


tokuhirom:
Closing — roast regression in SetHash/BagHash tests. Will revisit.

PR #2254, opened 2026-05-08
fix: autothread junction keys in hash subscript access

tokuhirom:
## Summary
- When indexing a hash with a junction key (e.g. `%h{any(0,1)}.push: 42`), expand the junction so each element becomes a separate hash key, matching Raku's autothreading semantics
- Previously the junction was stringified as a single key like `"any(0, 1)"`
- Adds junction handling to both `exec_index_autovivify_op` (slot ref creation) and `exec_index_assign_expr_named_op` (writeback)
- Adds `roast/S03-junctions/misc.t` (155 tests) to the whitelist

## Test plan
- [x] `prove -e 'target/debug/mutsu' roast/S03-junctions/misc.t` passes all 155 tests
- [x] `cargo fmt` applied

### DIFF
--- a/roast-whitelist.txt
+++ b/roast-whitelist.txt
@@ -157,6 +157,7 @@ roast/S03-feeds/basic.t
 roast/S03-junctions/associative.t
 roast/S03-junctions/autothreading.t
 roast/S03-junctions/boolean-context.t
+roast/S03-junctions/misc.t
 roast/S03-metaops/cross.t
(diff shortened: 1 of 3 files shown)


tokuhirom:
CI failures are pre-existing on main (same failures in run 25553772710 on commit 66fb9969). The `roast/S02-types/num.t` test 103 failure and the TODO-passed tests in `smiley.t`/`wacky.t` are not caused by this PR.

PR #2255, opened 2026-05-08
fix: support loading modules with type adverbs and Hash subclass element assignment

tokuhirom:
## Summary
- Compile `$?DISTRIBUTION` as `Nil` (outside distribution context), matching Raku behavior
- Parse `:ver<...>`, `:auth<...>`, `:api<...>` type adverbs on `unit class` declarations
- Named parameters in capture sub-signatures are now optional by default
- Add `is_container_subclass()` to skip element-level type checks for Hash/Array subclasses
- Convert Instance of Hash-subclass to plain Hash on element assignment

Enables `use Template::Mustache` to load without errors.

## Test plan
- [x] `make test` passes (all 480 files, 3926 tests)
- [x] `make roast` passes (only pre-existing IO test flakiness)
- [x] `use Template::Mustache; say "loaded"` works

### DIFF
--- a/roast-whitelist.txt
+++ b/roast-whitelist.txt
@@ -157,6 +157,7 @@ roast/S03-feeds/basic.t
 roast/S03-junctions/associative.t
 roast/S03-junctions/autothreading.t
 roast/S03-junctions/boolean-context.t
+roast/S03-junctions/misc.t
 roast/S03-metaops/cross.t
 roast/S03-metaops/eager-hyper.t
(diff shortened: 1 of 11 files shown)


tokuhirom:
The `roast/S02-types/num.t` test 103 failure is a pre-existing regression from main (PR #2249 `fix: preserve BigRat precision`). The failure also occurs on the main branch. This PR's changes do not affect Num parsing.

PR #2269, opened 2026-05-13
perf: optimize fast-call path to fix state.t timeout

tokuhirom:
## Summary
- Replace the O(env_size) full-env clone+merge in `call_compiled_function_fast` with targeted O(locals) save/restore of only the function's local variable entries
- Cache `routine_is_test_assertion_by_name` results in the fast-call eligibility check to avoid repeated expensive function resolution
- Add `roast/S04-declarations/state.t` to the whitelist (45/46 passing, 1 expected TODO failure)
- Add per-file timeout of 15s for state.t in roast-history.sh (the 2M-iteration loop takes ~4s CPU but may need more wall-clock time under system load)

The root cause was that the fast-call path cloned the entire env (hundreds of entries after loading Test) and merged all entries back after every function call. With the 2M-iteration loop in test 42, this O(env_size) work per call dominated runtime, causing the test to take ~24s and time out at 10s.

## Test plan
- [x] All 46 subtests in state.t produce same results (45 pass, 1 expected TODO fail)
- [x] Spot-checked several other test files for regressions (callframe.t, scoped-named-subs.t, main.t, begin.t)
- [ ] Full `make roast` to check for regressions


tokuhirom:
Closing — roast regressions (cmp-ok, die_on_fail) in CI. The env merge optimization needs more investigation to handle CI-specific edge cases.

PR #2279, opened 2026-05-14
fix: add is_bind to AssignExpr and implement rebind semantics for expression-context :=

tokuhirom:
## Summary

- Adds `is_bind: bool` field to `Expr::AssignExpr` to distinguish `:=` (bind) from `=` (assign) in expression context
- When `is_bind` is true, the compiler emits `MarkRebindContext` so the VM breaks existing alias chains instead of writing through them
- Adds `explicit_zero_params` field to `Stmt::For` / `ForLoop` opcode for `-> {}` detection
- Fixes `roast/S02-magicals/dollar-underscore.t`: 17/23 → 22/23 passing (tests 11, 14, 18 now pass)
- Updates `TODO_roast/S02.md` to reflect the new state

## Test plan

- [x] All 481 local tests pass (`make test`)
- [x] `cargo fmt --check` passes
- [x] `roast/S02-magicals/dollar-underscore.t` improved from 17/23 to 22/23 passing
- [ ] Test 16 deferred: requires `for` loop lexical scoping of `$_` per iteration

### DIFF
--- a/TODO_roast/S02.md
+++ b/TODO_roast/S02.md
@@ -78,12 +78,8 @@
 - [ ] roast/S02-magicals/dollar_bang.t
   - 14/15 pass. Test 7 fails: modifying a constant should set `$!`
 - [ ] roast/S02-magicals/dollar-underscore.t
-  - 17/23 pass. Failing tests need real Perl 6 container/aliasing semantics for `$_`:
-    - Test 14: pointy block with empty signature `-> { ... }` should die with "Too many positionals" when given an arg from `for`.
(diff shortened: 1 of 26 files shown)


tokuhirom:
CI fails: roast/S32-io/spurt.t tests 41-48 and 53-60 regressed (16 failures). These are `:createonly` and `:append` tests that pass on main. The `is_bind` / `AssignExpr` changes likely broke named argument handling for IO operations. Needs investigation before merge.

PR #2286, opened 2026-05-14
fix: pass all 65 subtests in roast/S17-promise/start.t

tokuhirom:
## Summary

- Fix `EVAL 'grammar { ... }'` by preserving `token_defs` across `restore_routine_registry_eval` so grammar tokens registered during EVAL survive after it returns
- Use `wait()` instead of `result_blocking()` in `.result` dispatch so the thread's buffered TAP output (from `pass`/`ok` inside `start { ... }`) is emitted at the right time
- Add `flush_shared_thread_output()` method and call it after `sleep` so output from concurrent start blocks appears before subsequent sequential TAP output from the main thread
- Add `roast/S17-promise/start.t` to `roast-whitelist.txt`

## Test plan

- [ ] `prove -e 'target/debug/mutsu' roast/S17-promise/start.t` passes all 65 tests
- [ ] `prove -e 'target/release/mutsu' roast/S17-promise/start.t` passes all 65 tests
- [ ] `make test` passes (all 481 local tests)

### DIFF
--- a/roast-whitelist.txt
+++ b/roast-whitelist.txt
@@ -781,6 +781,7 @@ roast/S17-promise/lock-async-stress.t
 roast/S17-promise/lock-async-stress2.t
 roast/S17-promise/lock-async.t
 roast/S17-promise/single-assignment.t
+roast/S17-promise/start.t
 roast/S17-promise/stress.t
 roast/S17-scheduler/at.t
 roast/S17-scheduler/every.t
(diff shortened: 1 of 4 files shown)


tokuhirom:
Closing: the flush_shared_thread_output changes break wait.t TAP ordering in CI (tests out of sequence). start.t tests 45/53/58 also fail consistently. The thread output flushing approach needs rethinking — interleaving thread output with TAP causes ordering issues.

PR #2287, opened 2026-05-14
fix: restore $_ topic after lives-ok/dies-ok blocks and support Range+Mixin arithmetic

tokuhirom:
## Summary

- Fixed `$_` topic variable leaking from `lives-ok`/`dies-ok` block evaluation: `eval_block_value` sets `env["_"]` (bare key) to the block's last result, but the save/restore only guarded `env["$_"]` (sigiled key). This caused `$_` to change after every `lives-ok { expr }` call, breaking subsequent code that read `$_`.

- Fixed arithmetic on `Value::Mixin` wrapping a Range (e.g. `(2..^5) but Meows`): `arith_add`/`sub`/`mul`/`div` now detect Mixin-wrapped Ranges, unwrap the inner Range, perform the arithmetic, and re-wrap the result with the same Mixin.

## Effect

`roast/S03-operators/range.t` now runs all 181 subtests (was stopping at 119) and passes 155 of them (up from 119). The 26 remaining failures are pre-existing issues (R.. reverse-range operators, X::Worry::Precedence::Range warnings not firing, and Range*Mixin representation differences).

## Test plan

- [x] `make test` passes (only pre-existing flaky `t/lock.t` race condition)
- [x] `roast/S03-operators/range.t` runs to completion (181 tests, was 119)
- [x] No regressions in roast whitelist

### DIFF
--- a/roast-whitelist.txt
+++ b/roast-whitelist.txt
@@ -781,6 +781,7 @@ roast/S17-promise/lock-async-stress.t
 roast/S17-promise/lock-async-stress2.t
 roast/S17-promise/lock-async.t
 roast/S17-promise/single-assignment.t
+roast/S17-promise/start.t
 roast/S17-promise/stress.t
 roast/S17-scheduler/at.t
 roast/S17-scheduler/every.t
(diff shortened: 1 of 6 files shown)


tokuhirom:
Closing in favor of #2288 which is rebased on main

PR #2326, opened 2026-05-17
fix: END phasers use current env values instead of stale captured values

tokuhirom:
## Summary
- Fix END phaser environment overlay in `finish()` to only add variables not already present in the current env, preventing stale captured values from overwriting current (final) values
- When an END phaser was registered inside a closure (e.g. via EVAL), the captured env snapshot would overwrite shared variables at program exit
- Adds `roast/S04-phasers/in-eval.t` to whitelist (35 tests)

## Test plan
- [x] `prove -e 'target/release/mutsu' roast/S04-phasers/in-eval.t` passes (35/35)
- [x] `make test` passes (no new regressions)
- [x] `cargo fmt` and `cargo clippy -- -D warnings` clean

### DIFF
--- a/roast-whitelist.txt
+++ b/roast-whitelist.txt
@@ -287,6 +287,7 @@ roast/S04-phasers/end.t
 roast/S04-phasers/enter-leave.t
 roast/S04-phasers/exit-in-check.t
 roast/S04-phasers/first.t
+roast/S04-phasers/in-eval.t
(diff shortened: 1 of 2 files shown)


tokuhirom:
Closing: this change causes regressions in S04-phasers/end.t tests 6-7. The END phaser lexical capture needs a more nuanced approach that distinguishes between lexical captures and globals.

PR #2335, opened 2026-05-18
fix: END phasers in EVAL use live variable values instead of stale snapshots

tokuhirom:
## Summary
- Fixed END phasers registered inside EVAL'd closures using stale captured environment snapshots instead of live variable values at program exit
- The `finish()` method now only overlays captured variables that are missing from the current env, preserving live bindings for shared variables
- Adds `roast/S04-phasers/in-eval.t` to the whitelist (all 35 tests pass)

## Test plan
- [x] `prove -e 'target/debug/mutsu' roast/S04-phasers/in-eval.t` — all 35 tests pass (was 34/35)
- [x] `make test` — all 484 files, 3959 tests pass
- [x] No clippy warnings in changed file

### DIFF
--- a/roast-whitelist.txt
+++ b/roast-whitelist.txt
@@ -288,6 +288,7 @@ roast/S04-phasers/end.t
 roast/S04-phasers/enter-leave.t
 roast/S04-phasers/exit-in-check.t
 roast/S04-phasers/first.t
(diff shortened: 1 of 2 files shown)


tokuhirom:
Closing due to regression with lexical lookups from END blocks. Will reopen with a more targeted fix.

PR #2365, opened 2026-05-21
fix: implement zip-latest semantics distinct from zip for Supply

tokuhirom:
## Summary
- Implement proper `zip-latest` semantics for `Supply`: keeps only the latest value from each source and emits a combined value whenever any source emits (once all sources have provided at least one value)
- Previously `zip` and `zip-latest` shared the same code path using regular zip semantics (buffer per source, consume one from each)
- Support `:with` and `:initial` named parameters for both instance and class method forms
- Bridge channel-based supplies (e.g. `Supply.interval`) to the supplier subscription system for zip tap compatibility

## Test plan
- [x] 12 of 13 subtests pass in `roast/S17-supply/zip-latest.t` (up from 0)
- [x] `cargo fmt` applied
- [x] `make test` passes (only pre-existing flaky `t/lock.t` test 10 failure)
- [ ] `make roast` - CI will verify no regressions

The remaining test 13 (`Supply.interval` + `supply`/`whenever` + `await`) requires deep integration between channel-based and supplier-based supply event loops, which is a separate architectural concern.

### DIFF
--- a/roast-whitelist.txt
+++ b/roast-whitelist.txt
@@ -813,6 +813,7 @@ roast/S17-supply/Channel.t
 roast/S17-supply/Promise.t
 roast/S17-supply/Seq.t
 roast/S17-supply/act.t
+roast/S17-supply/batch.t
 roast/S17-supply/classify.t
 roast/S17-supply/collate.t
 roast/S17-supply/comb.t


tokuhirom:
Closing: the zip-latest semantics code changes have already been merged into main via other PRs. The only remaining change was adding roast/S17-supply/batch.t to the whitelist, but that test times out on CI (7 of 9 subtests run before timeout), so it cannot be whitelisted yet.

PR #2469, opened 2026-05-27
feat: inline grammar action dispatch for dynamic variable propagation

tokuhirom:
## Summary
- Grammar action methods now fire inline during regex matching (not just post-parse)
- Enables dynamic variable changes in actions to affect subsequent parsing
- Required for Template::Mustache delimiter change (`{{=<% %>=}}`)

## Test plan
- [x] New test `t/grammar-inline-actions.t` (4 tests)
- [x] `make test` passes


tokuhirom:
Needs more work - LTM skip causes regression with :our. Will revisit.

PR #2470, opened 2026-05-27
fix: proper env scoping for proto dispatch and map closure iterations

tokuhirom:
## Summary
- Use `push_caller_env`/`pop_caller_env_with_writeback` in `call_proto_function` and `call_proto_dispatch` instead of manual env save/restore via `restore_env_preserving_existing`, preventing recursive proto calls from clobbering outer call locals
- Restore closure captured env before each map iteration in `eval_map_over_items` to prevent stale values from previous iterations' recursive calls
- Remove now-unused `restore_env_preserving_existing` method

Fixes Template::Mustache section iteration rendering only the first element (e.g. `{{#items}}[{{name}}]{{/items}}` with two items now correctly produces `[A][B]` instead of `[A]`).

## Test plan
- [x] `timeout 10 ./target/debug/mutsu tmp/test-section-contents2.raku` outputs `Result: '[A][B]'`
- [x] `make test` passes
- [x] `cargo fmt` passes

### DIFF
--- a/src/compiler/stmt.rs
+++ b/src/compiler/stmt.rs
@@ -26,10 +26,56 @@ impl Compiler {
         let pkg = self.current_package.clone();
         let qualify_parent = |p: &String| -> String {
             if p.contains("::") || p.is_empty() {
-                p.clone()
-            } else {
-                format!("{}::{}", pkg, p)
+                return p.clone();
             }
+            // Well-known builtin types should not be package-qualified.
+            // "Grammar" inside a module should stay "Grammar", not become
+            // "MyModule::Grammar".
+            if matches!(
+                p.as_str(),
+                "Any"
+                    | "Cool"
+                    | "Mu"
+                    | "Grammar"
+                    | "Match"
+                    | "Int"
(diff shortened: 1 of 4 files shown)


tokuhirom:
CI failed on 3 roast tests:
- `roast/S04-statement-modifiers/for.t` test 31
- `roast/S11-modules/export.t` plan mismatch (59 planned, 57 ran)
- `roast/S32-list/map.t` test 61

The map.t failure is likely related to the closure env re-insertion change. Please investigate and fix.

tokuhirom:
Closing: the root cause analysis was incorrect. The Template::Mustache section iteration issue is not caused by proto dispatch env clobbering or map closure env restoration. Investigation revealed the actual issue is that the section data (a 2-element tuple) is being treated as a single item in .map, likely an itemization handling bug in the map dispatch path. The proto dispatch env changes (push_caller_env/pop_caller_env_with_writeback) caused a regression in roast/S11-modules/export.t (2 tests dropped). See PR discussion for detailed investigation notes.

PR #2471, opened 2026-05-27
fix: grammar parent class resolution inside packages

tokuhirom:
## Summary
- Fix compiler's `qualify_parent` to skip package-qualifying builtin type names
- Grammar inside a module (e.g. `class Foo { grammar G { ... } }`) no longer errors with "unknown parent class Foo::Grammar"
- Unblocks HTTP::Parser module loading

## Test plan
- [x] New test `t/grammar-in-module.t` (2 tests)
- [x] `make test` passes (verified before commit)
- [x] HTTP::Parser loads successfully


tokuhirom:
Will recreate with last_stack_value guard included.

PR #2472, opened 2026-05-27
fix: only use VM stack value in map when stack has exactly one entry

tokuhirom:
## Summary
Guard `last_stack_value()` to only return when `stack.len() == 1`. Fixes export.t tests 58-59 regression from #2468 where stale stack values were picked up by `.map` callbacks.

## Test plan
- [x] `(1,2,3).map(&process)` with given/when still returns `("one","two","other")`
- [x] `roast/S11-modules/export.t` passes 59/59 in release build
- [x] lefthook (fmt + clippy) passes


tokuhirom:
Merging this fix into PR #2471 instead. The for.t/increment.t/basic.t failures appear to be pre-existing flaky tests unrelated to this change.

PR #2498, opened 2026-05-28
fix: S03-metaops/reduce.t improvements (563/580 passing)

tokuhirom:
## Summary

- Fixed X::Syntax::CannotMeta error for non-associative reduce operators (`leg`, `<=>`, `cmp`)
- Fixed X::Str::Numeric for single-element reduce with non-numeric string (e.g. `[+] 'hello'`)
- Coerce Instance values via Numeric()/Bridge() for single-element reduce (e.g. `[*] CustomNumify.new`)
- X/Z meta-operators now return `Value::Seq` for correct `eqv` comparisons
- `minmax` operator supports Array/Seq arguments and returns `Value::Range` for integer bounds
- `min`/`max` treat undefined values as `Inf`/`-Inf` respectively
- Triangle scan `[\op]` for Z~/X~ operators recomputes from scratch at each step
- Initial scan value for Z~/X~ wraps single element in `Seq`; `minmax` gives `Range`
- `andthen` scan correctly truncates result when short-circuit fires (remaining elements dropped)
- Functional `reduce(&infix:<+>, single-elem)` applies identity element for type coercion
- Seq consumed/cached state tracking infrastructure (for S32-list/seq.t)

## Test plan

- [ ] `prove -e 'target/debug/mutsu' roast/S03-metaops/reduce.t` shows 17 failures (down from 34+)
- [ ] `make test` passes
- [ ] `make roast` passes whitelisted tests

### DIFF
--- a/src/builtins/methods_0arg/coercion.rs
+++ b/src/builtins/methods_0arg/coercion.rs
@@ -185,7 +185,16 @@ pub(super) fn dispatch(target: &Value, method: &str) -> Option<Result<Value, Run
         },
         "Capture" => Some(value_to_capture(target)),
         "Slip" => match target {
-            Value::Array(items, ..) | Value::Seq(items) => Some(Ok(Value::Slip(items.clone()))),
+            Value::Seq(items) => {
+                if crate::value::seq_is_consumed(items) && !crate::value::seq_is_cached(items) {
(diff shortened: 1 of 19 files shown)


tokuhirom:
Closed in favor of #2524 which rebases this branch onto current main. The cherry-pick applied cleanly with no conflicts.

PR #2526, opened 2026-05-31
fix: user-defined subs properly shadow same-named builtins

tokuhirom:
## Summary
- Fix bug where user-defined subs with optional/default parameters failed to shadow same-named builtins when called with fewer-than-max args
- Root cause: compiled function lookup used call-site arity instead of definition arity for key matching
- Also generalizes the builtin shadowing fallback from a hardcoded 9-function whitelist to work for any user-defined function

## Example
```raku
sub rotate($s, $n = 3) { return "user: $s $n"; }
say rotate("hello");  # Before: Nil (builtin), After: "user: hello 3"
```

## Test plan
- [x] `roast/integration/advent2009-day08.t` now passes (was blocked by this bug)
- [x] `roast/S06-signature/defaults.t` passes
- [x] `roast/S06-signature/optional.t` passes
- [x] `roast/S06-advanced/dispatching.t` passes
- [x] `t/map.t`, `t/grep-regressions.t`, `t/first-p-adverb.t` pass (previously shadowed builtins)
- [ ] CI (`make test` + `make roast`) passes


tokuhirom:
Superseded by new PR from clean branch

PR #2615, opened 2026-06-05
vm: nativize the default .new constructor (decouple, lever A)

tokuhirom:
## What

Nativizes the default `.new` constructor for plain user classes in the VM, eliminating the dominant remaining **method-call interpreter fallback** (PLAN.md レバー A).

`ClassName.new(:a(...))` routed through `Interpreter::call_method_with_values` → `dispatch_new` even though, for a plain user class, default construction is pure data assembly. This adds:

- `Interpreter::simple_ctor_plan` — a registry-only eligibility check returning the attribute names to pre-fill with `Nil`.
- VM-side `try_native_default_construct` — called from both `try_compiled_method_or_interpret` and its mut twin, right before `record_method_fallback`. The eligibility plan is cached per class (`simple_ctor_cache`, invalidated alongside `fast_method_cache` and on class registration/augment).

## Eligibility

A class is eligible when it **and its whole MRO** (excluding the `Any`/`Mu`/`Cool` roots) are plain user classes: no `BUILD`/`TWEAK`/`BUILDALL`/custom `new`, no composed roles or role-param bindings, no native methods, no wildcard `handles`, no twigil-less alias attributes, and only simple `$`-sigiled attributes with no default/required/type/where/smiley constraint, and no duplicated attribute names across the hierarchy. This covers single classes **and inheritance** (`Dog is Animal`).

Any positional arg, an active wrap chain, or an ineligible class falls back unchanged (identical behavior, including the interpreter's positional/error path). Attribute defaults still fall back (they need expression evaluation).

## Results

| program | method fallback before | after |
|---|---|---|
| `method-call.raku` (`Point.new` ×10000) | 10001 / 50002 (20.0%) | **0 / 50002 (0.0%)** |
| `bench-class` (`Animal`/`Dog.new`, inheritance) | 9000 / 65000 (13.8%) | **0 / 65000 (0.0%)** |

Both benchmarks now have **zero** method fallback.

## Verification (behavior-preserving)

- `make test`: only the pre-existing `placeholder.t` / `tail-function.t` / `wrap.t` failures (identical on `main`).
- All 70 whitelisted S12/S14 tests pass; an 88-file random whitelist sample passes.
- The 5 partially-failing non-whitelisted S12 tests fail at the **identical subtest** on `main`.

Also corrects PLAN.md's EVAL note: EVAL is **not** a tree-walk fallback — its execution already compiles to bytecode and runs on a sub-VM (`run_compiled_block` → `VM::new`). The residual coupling is shared-state ownership, which converges with lever B, not a permanent interpreter exception.


tokuhirom:
Closing as duplicate: native default `.new` construction was already merged to main as 'Step 5' (`Interpreter::try_native_default_construct`) via a concurrent PR while this was in progress. main's version achieves the identical result (method-call.raku 20%→0%, bench-class 13.8%→0%, inheritance covered) and is strictly more complete (it also evaluates attribute defaults). No knowledge lost — the feature is already on main. Will refocus on the genuinely remaining lever-A item (closure/regex-taking methods: map/grep/sort/subst blocks).

PR #2674, opened 2026-06-06
fix(vm): loop/if body-local `my` no longer clobbers an enclosing same-named outer var (lever C Slice 3)

tokuhirom:
## Summary

`my $x = 99; for/while/loop/repeat/if { my $x = ... }; say $x` printed the loop's last-iteration value instead of `99` — a block-local `my` inside a loop/if body **clobbered an enclosing same-named outer lexical**. This is a closure-independent correctness bug affecting valid Raku programs (PLAN.md lever C "Slice 3").

```raku
my $x = 99;
for 1..3 { my $x = $_ }
say $x;   # raku: 99   mutsu (before): 3   mutsu (after): 99
```

## Root cause

The name-keyed dual-store env: loop bodies mark every local `needs_env_sync`, so the inner `my $x`'s write-through overwrote both `env["x"]` and the **shared local slot** of the outer `$x`.

## Approach

A compiler-side **fresh-slot-on-shadow** approach was tried and rejected: allocating a second slot with the same name makes the name-based `captured_mutated_locals` / box-on-capture analysis (lever C Slice 2/2b) over-apply across blocks and breaks per-iteration closure capture — multiple same-named slots fight the name-keyed env architecture.

Adopted **runtime restore** instead:
- When a loop-body `my` shadows an *existing* outer binding, record the shadowed outer value in `loop_local_saved_env` at declaration time and restore it (both env **and** the shared slot — `GetLocal`'s fast path returns a non-Nil slot value without consulting env) on loop exit.
- Declarations in a loop **condition** (`while`/`until`/`loop`, incl. the `… until COND` statement modifier) are enclosing-scoped and commonly read after the loop, so they are excluded via a new `loop_cond_active` flag (fixes the `t/iterator-regressions.t` `my @a` accumulator pattern).
- `if`/`else` branches run once, so they reuse the existing `BlockScope` full save/restore via a new `body_declares_block_local` wrap condition.

## Remaining (unchanged, invalid Raku)

A pure post-loop read of a body-local name with **no** outer binding (`for { my $x }; say $x`) returns an undefined value rather than the compile-time error Raku raises — mutsu does not enforce strict-vars anywhere (`$neverdeclared` is also Nil), so there is no correctness impact.

## Tests

- New pin `t/block-local-my-scope.t` (15 cases across for/while/C-style/repeat/if/else, nested, `@`/`%`, and closure-capturing loops).
- `make test` green (5007 tests).
- Lever-C closure pins still green: `closure-container-capture`, `loop-body-local-closure-capture`, `loop-var-closure-capture`, plus `iterator-regressions`, `scoped-overlay-named`.


tokuhirom:
Pushed a follow-up fix: this branch regressed whitelisted `roast/S04-statement-modifiers/for.t` test 20.

`(my @a).push: $_ for ^3;` left `@a` empty after the loop (raku: `0,1,2`). A `my` in a *statement modifier* introduces no block, so it's enclosing-scoped (no local slot), but `pop_loop_local_scope`'s restore was recording it as a loop-body-local shadow and resetting it to its hoisted/first-iteration value.

Fix in `exec_set_var_dynamic_op`: only record a shadow when there is a **coherent live** enclosing binding — a `code.locals` slot for the name whose value equals the env value — and the name wasn't already declared by this loop in a prior iteration. This distinguishes:
- genuine outer shadow (`my @a=1,2,3; for { my @a=7,8 }`) → slot==env → still restored to `1,2,3`;
- self-hoisted `@`/`%` (`(my @a).push for ^3`) and a popped sibling block's lingering frame-wide slot (`{ my @a=... } { (my @a).push for ^3 }`) → slot≠env → left alone, so the value survives.

Added pins to `t/block-local-my-scope.t` (now 19). `make test` green, `S04-statement-modifiers/for.t` + `block-local-my-scope.t` pass locally.

PR #2686, opened 2026-06-07
docs(vm-dual-store): prep Slice 6.3 (env_dirty removal = lever B finish)

tokuhirom:
## Summary

Doc-only. Prepares **Slice 6.3** — removing `env_dirty`, the last dual-store flag, which finishes lever B — so the next session can execute it directly instead of re-surveying.

Surveyed `env_dirty` (the env→locals re-sync bridge) on 2026-06-07:

- **Consumer map** (the leverage point — few sites): `ensure_locals_synced` (env_dirty-gated) → `sync_locals_from_env`, called before local-aliasing reads in the I/O ops + a handful of var/register/arith/closure sites.
- **164 `env_dirty = true` setters grouped by blocker:**
  - ~92 post-dispatch conservative marks (method-mut 47 / method 32 / func 13) — the multi-tier merge (#2684) already signals captured-outer writes precisely, so the *blanket* post-call set is the redundant part to attack first.
  - ~36 by-name var/control ops — need *reverse* write-through to the slot (mirror of Slice 6.2's forward write-through).
  - ~21 I/O + carrier marks — gated on removing the interpreter bridge for EVAL/format/atomic.
- **Staged removal plan + metric** (`MUTSU_VM_STATS=1` `locals_pulls` → 0 on `mb_method`/`mb_nested`, output unchanged).

Also records an **unresolved wall-clock follow-up** for the multi-tier overlay (#2684): the `env_deep_copies` win is clear, but the per-call `Arc<Env>`/`Vec` cost wasn't cleanly measured (box under heavy load). Next session should re-measure with `hyperfine` (5f4501ea vs 9419784f) and, if method-call/bench-fib regressed, skip the `Arc<Env>` wrapper when the parent is flat.

(Also notes, for the record, a **pre-existing** bug unrelated to this work: nested method attribute mutation `method a(){self.b()}; method b(){$!n++}` returns 0 instead of raku's 2 — reproduces on `main`, the outer method's `attr_slots`-based writeback doesn't pick up the identity-overwrite of its `self` instance local.)


tokuhirom:
Superseded by #2689, which is branched directly on top of this commit (so it includes this prep doc verbatim) and executes Slice 6.3 stage 1: the per-method-call `env_dirty` false-positive is removed (`locals_pulls` 10001→1 on mb_method/mb_nested/@a.elems). Closing in favor of #2689 to avoid a duplicate squash of the same doc commit; all knowledge here is preserved in that PR.

PR #2699, opened 2026-06-07
perf(vm): drop blanket env_dirty after smartmatch; reverse write-through lhs (lever B Slice 6.3 step 2)

tokuhirom:
## Summary

Lever B Slice 6.3 **step 2** (first target): the smartmatch (`~~`) opcode no longer forces a per-match env→locals pull.

`exec_smart_match_expr_op` used to set `self.env_dirty = true` unconditionally after every match, forcing an O(caller-locals) `sync_locals_from_env` on the caller's next `GetLocal` — one pull per loop iteration.

```
MUTSU_VM_STATS=1 mutsu -e 'my $c=0; for ^5000 { if "x 42 y" ~~ /\d+/ {$c++} }; say $c'
# locals_pulls: 5001 -> 1 ;  bench-string: 5003 -> 3
```

### Fix
The only by-name env writes the op makes that can alias a caller local slot are:
- topic `$_` — restored to its pre-match value, so already consistent with its slot;
- the lhs alias `var_name` for `$x ~~ ...` — now **reverse-write-throughs** into its local slot via `update_local_if_exists` (mirror of Slice 6.2's forward `flush_local_to_env`).

The regex engine's `$/` / `$0..` / `$<name>` are special-cased and never name a caller local slot (same skip set as `sync_regex_interpolation_env_from_locals`). Any user code the RHS runs (a sub call) signals its own caller-aliasing writes precisely through dispatch.

### Tests
- pin `t/smartmatch-env-dirty.t` (14): caller-local coherence across a loop, capture vars readable after a match, `s///`+`tr///` mutating a local lhs through the topic alias, `given`/`when`, RHS-sub side effects.
- `make test` PASS (5143). S03-smartmatch / S05-capture / S05-substitution / given-when roast green.
- The 2 fails in `S05-capture/hash.t` + `S05-substitution/subst.t` are **pre-existing**, non-whitelisted, **identical count with and without this change** (verified by stash compare).

Resume map / remaining step-2 setters tracked in `docs/vm-dual-store.md` and PLAN.md.


tokuhirom:

PR #2703, opened 2026-06-07
perf(vm): precise env_dirty for 0-arg function fast path (lever B Slice 6.3 step 2)

tokuhirom:
## Summary

Lever B Slice 6.3 **step 2** (second target): a pure 0-arg compiled-function call no longer forces a per-call env→locals pull.

> Stacked on #2699 (smartmatch). The first commit here is #2699's; this PR's own change is the second commit. The diff collapses to just that once #2699 merges.

The `CallFunc` zero-arg fast path set a blanket `self.env_dirty = true` after every `call_compiled_function_fast`, so even `sub f { 42 }` forced an O(caller-locals) `sync_locals_from_env` on the caller's next `GetLocal`:

```
MUTSU_VM_STATS=1 mutsu -e 'sub f { 42 }; my $c=0; for ^5000 { $c = $c + f() }; say $c'
# locals_pulls: 5001 -> 2
```

### Fix
`call_compiled_function_fast` already signals env_dirty precisely for a function **with** locals (scoped-overlay / clone merge → only on a captured-outer write). The gap was the **0-local** case: no overlay, no clone → no merge, so only the caller's blanket mark kept `sub f { $outer = 5 }` correct.

In the 0-local no-merge branch, gate env_dirty on the compile-time **`cf.code.has_env_writes`** flag: a body that performs no env write (no assign / increment / nested call / registration) cannot dirty a caller slot — the dispatch's routine `_` is restored — so it needs no pull. The blanket mark at the call site is removed.

**KEY GOTCHA (do not repeat):** extending `use_scoped` so a 0-local function installs a scoped overlay (and the merge detects the write) **regresses `env_deep_copies` 1 → 5000** — inserting the routine `_` into the freshly-chained overlay forces a `make_mut`. The compile-time `has_env_writes` gate is zero runtime cost and avoids the overlay.

### Measured (debug; counters are opt-level-independent)
| case | before | after |
|---|--:|--:|
| `for ^5000 { $c += f() }` locals_pulls | 5001 | 2 |
| same env_deep_copies | 1 | 1 |
| fib locals_pulls / deep_copies | 0–1 / 0 | 0–1 / 0 |

### Tests
- pin `t/zeroarg-env-dirty.t` (16): pure const, local-only, captured-outer scalar/increment/array/hash writes, nested 0-arg calls, recursion, read-modify-write, caller-local coherence across loops.
- `cargo test` 456, full `prove t/` 589 green, `cargo clippy -D warnings` clean.
- S06-multi / S06-routine-modifiers (whitelisted) green. The S06-signature/slurpy-* and S32-basics/xxPOS.t debug failures are **pre-existing** (identical on `origin/main`, verified by checkout — slurpy-scalar gap / debug-only).

Remaining step-2 setters (3 non-uniform kinds) catalogued in `docs/vm-dual-store.md` and PLAN.md.


tokuhirom:

PR #2705, opened 2026-06-07
perf(vm): precise env_dirty for heavy named function dispatch (lever B Slice 6.3 step 2)

tokuhirom:
## Summary

Lever B Slice 6.3 **step 2** (third target): the heavy "named" compiled-function dispatch path no longer forces a per-call env→locals pull.

> Stacked on #2699 (smartmatch) and #2703 (0-arg fast path). The first two commits here are those; this PR's own change is the third commit. The diff collapses once they merge.

`call_compiled_function_named` handles complex signatures (default params, return types, where-constraints, typed params, multis). Its callers set a blanket `self.env_dirty = true` after every call, so even a pure call forced an O(caller-locals) `sync_locals_from_env`:

```
mutsu -e 'sub g($x=5){$x+1}; my $c=0; for ^5000 { $c=$c+g(2) }; say $c'
# locals_pulls: 5001 -> 1   (same for return-type / where / typed / multi callees)
```

### Fix (mirrors `call_compiled_function_fast`)
`call_compiled_function_named` now signals env_dirty **precisely** from its return merge: save the caller's incoming dirtiness, reset before the body (body nested-call dirtiness is about the callee env, reconciled by the merge), set `env_dirty = saved || changed`. `changed` is true only when the merge wrote a caller-slot-aliasing value:
- a plain-lexical (captured-outer) writeback whose value actually differs,
- a non-empty `is rw` param writeback,
- an `is raw` return.

Dynamic-var writeback (`pop_caller_env_with_writeback`) targets `$*x` names with no compiled slot → never obliges a pull.

Blanket marks removed at the hot call sites (vm_call_func_ops heavy / OTF-cached / cached named; vm_var_get_ops bareword-term named). EVAL/carrier callers (vm_call_exec_ops) keep their marks — step 3.

### Tests
- pin `t/named-call-env-dirty.t` (16): default/return-type/where/typed caller-local coherence, captured-outer + is-rw + dynamic-var writeback propagation, nested heavy calls, read-modify-write, condition use.
- `make test` (prove t/ 590; cargo test 456); `cargo clippy -D warnings` clean. S06-multi / S06-routine-modifiers (whitelisted) green.
- **Pre-existing, baseline-identical** failures left untouched (all reproduce on `origin/main`, unrelated to env_dirty): multi caller-accumulator (`for { $c += mm(2) }` → 2 not 10), is-raw return aliasing, S12-methods/accessors+qualified.

Remaining step-2 setters (cold redundant-drops / loop round-trips) catalogued in `docs/vm-dual-store.md` and PLAN.md.


tokuhirom:

PR #2707, opened 2026-06-07
perf(vm): drop redundant env_dirty in native push fast path (lever B Slice 6.3 step 2)

tokuhirom:
## Summary

Lever B Slice 6.3 **step 2** (fourth target): the native `@a.push` fast path no longer forces a per-push env→locals pull.

> Stacked on #2699 / #2703 / #2705. This PR's own change is the final commit; the diff collapses as the stack merges.

The push fast path (vm_data_ops) set a blanket `self.env_dirty = true` after every push. But it mutates only the target array in env and **already reverse-write-throughs** the result into the target's local slot (the `// Restore local slot` block just above), so the caller's slot is coherent — the mark is redundant.

This is **not a cold path** — `@a.push($_)` in a loop is extremely common:

```
mutsu -e 'my @a; for ^5000 { @a.push($_) }; say @a.elems'
# locals_pulls: 5001 -> 2
```

The interpreter-fallback push branches (shared / shaped / non-simple-array targets) keep their conservative mark.

### Tests
- pin `t/push-env-dirty.t` (13): loop accumulation, caller local interleaved with pushes, multi-arg/slip push, captured-outer push from a sub, push return value, nested arrays, typed arrays, lockstep pushes.
- `make test` (prove t/ 592; cargo test 456); `cargo clippy -D warnings` clean.
- **Pre-existing, baseline-identical, out of scope:** `@g := @f; @f.push(3)` doesn't update `@g` (`:=` array-alias = first-class container, lever C / §2; the old blanket mark never fixed it). S32-array/list big failure counts (adverbs/splice/skip) are pre-existing unimplemented-feature gaps; `skip.t` standalone matches baseline (3/55).

After this, the remaining step-2 setters are genuinely cold / zero-benefit (declaration-time trait handlers, `:=` element SlotRef, loop/control env round-trips) — see `docs/vm-dual-store.md` and PLAN.md. They are best done together when env_dirty is finally deleted.


tokuhirom:

PR #2746, opened 2026-06-08
feat(container): Phase 1 slice 1 — non-loop sibling closures share a scalar container (lever C)

tokuhirom:
## Summary

Container-identity **Phase 1, first slice**: non-loop sibling closures now share the container cell of a captured-and-mutated `$` scalar, completing **lever C** (the remaining non-loop capture case).

Target case (was broken — raku: `42`, mutsu: `0`):
```raku
sub make { my $v = 1; my $g = { $v }; my $s = -> $n { $v = $n }; return $g, $s }
my ($g, $s) = make(); $s(42); say $g();   # now 42
```

## Why this is safe now

The shared cell (`Value::ContainerRef`) and the boxing machinery already existed but were deliberately restricted to loop-body locals, because boxing a non-loop captured-mutated local could leak a raw `ContainerRef` into value-inspection paths. **PR #2742** centralized the hot value reads (GetLocal/GetGlobal) through the `Value::into_deref` decont chokepoint, so a boxed cell is unwrapped before reaching any value op — which unblocks lifting the restriction.

## Changes

- **`box_captured_lexicals`** (`vm_register_ops.rs`): drop the loop-only gate so non-loop captured-mutated `$` scalars are boxed into a shared `ContainerRef`. Sibling `Sub` envs share the `Arc` via `flattened`; reads decont via GetGlobal, writes go through SetGlobal write-through. Sigil filter + reference-type skip keep this to plain `$` scalars.
- **Leak fix** (`vm_call_method_compiled.rs`): `try_eval_simple_protect_expr`'s `Expr::Var` arm read locals/env directly (bypassing GetLocal/GetGlobal), so it now `into_deref`s — the one value-context leak in the audit (e.g. `@a.push($v)`).
- **Constraint guard**: do **not** box type-constrained scalars (`my Int $x`, subset/`where`). ContainerRef write-through (`$x++`, `$x = ...`, `.=`) bypasses the assignment chokepoint's constraint re-check; keeping them unboxed preserves enforcement (fixes a `t/where-constraint-var.t` regression). Sharing constrained captured scalars is deferred to the broader container-write audit.

Removes the two `todo` markers in `t/closure-container-capture.t` (now pass). Side effect: `S02-names-vars/variables-and-packages.t` 16 → 13 failures.

## Verification

- `cargo build` / `cargo clippy -- -D warnings` / `cargo fmt` clean
- `make test` PASS — 456 unit + 5274 prove, no regressions
- Target case → `42`; closure net (`t/closure-container-capture.t`, `closure-selective-env-sync.t`, `loop-body-local-closure-capture.t`, `closure-captured-state.t`, `closure-rw-param-writeback.t`, `closure-nested-writeback.t`, `wrap-closure-capture.t`) all PASS
- Whitelisted roast: `S02-names-vars/{contextual,signature}.t`, `S03-binding/{arrays,closure,hashes,ro}.t`, `S04-statements/return.t`, `S06-routine-modifiers/{lvalue-subroutines,proxy}.t`, `S32-list/reduce.t`, `S09-typed-arrays/native-int.t` all PASS
- `gather.t` #38 is the pre-existing Phase 2 `take-rw` item (non-whitelisted, unrelated)
- Full roast delegated to CI


tokuhirom:
Follow-up fix pushed: the initial CI run regressed `S12-methods/submethods.t` (and `S24-testing/*`, `test-util`, `isDEPRECATED`) via the ubiquitous test-helper pattern `my $x; block { $x = Obj.new }; block { $x.mutate }; $x.read`.

**Root cause:** `overwrite_instance_recursive` (the chokepoint that propagates a mutating method's result to all env bindings holding that instance by identity) did not descend into `Value::ContainerRef`. A boxed scalar holding an instance lost its mutation because the writeback never reached the instance inside the shared cell.

**Fix:** added a `Value::ContainerRef` arm that locks the cell and recurses in place, so instance mutation-writeback is now cell-aware (PR #2742 already made *reads* cell-aware). Array/Hash/`.=`/`++` mutation on boxed scalars verified already cell-aware.

Local: submethods.t / S24-testing / test-util / isDEPRECATED now PASS; `make test` PASS (456 unit + 5274 prove). `S17-procasync/basic.t` #46 is the known S17 subprocess flaky; `S32-num/int.t` is a heavy debug-build timeout (plan 165, Failed:0) — both rely on the CI release run.

tokuhirom:
Closing: the broad-boxing approach in this PR is reverted — it was under-scoped on both correctness (mutation-writeback through cells unimplemented across paths) and performance (boxing per closure creation made roast/S32-num/int.t ~1s → 150s+, a CI release-roast timeout). The finding and the design constraint for the real slice (escape analysis + mutation-writeback audit) are recorded in docs/container-identity.md via #2749. main retains only the safe into_deref groundwork (#2742).

PR #2765, opened 2026-06-08
refactor(container): delete closure_captured_state — escape analysis subsumes it (step 2, 4→3)

tokuhirom:
## Summary

Container-identity **step 2** (PLAN.md §🟣). Deletes one of the three overlapping closure-capture mechanisms: the `closure_captured_state` side table. Step 1's escape analysis (#2758) now boxes escaping-mutating-closure captures into shared `ContainerRef` cells, which makes this side table **redundant**.

## What it was

`closure_captured_state: HashMap<(closure_id, Symbol), Value>` persisted a closure instance's captured-variable mutations across calls (re-applied on call entry, re-saved on exit, for every free var). Its purpose was **per-instance independence**: two closures from one factory each keep their own state.

## Why it's now redundant

After step 1, `my $inc = -> { ++$x }` and `sub mk { my $n=0; sub { $n++ } }` are *escaping* closures, so `$x`/`$n` are now `ContainerRef` cells. A fresh declaration installs a fresh cell, so:
- each factory **call** already gets its own state (fresh `my $n` → fresh cell), and
- **siblings** already share one cell.

The side table just shadowed the live cells (it even had to explicitly skip `ContainerRef` values on read).

## Verification

With read/write disabled to measure deletability:
- `make test` PASS (5336); non-escaping multi-call `my $x=0; my $inc=->{++$x}; $inc();$inc(); $x` → 2; factory independent state `mk();mk()` → 010; `closure-captured-state.t` PASS.
- **Release sweep of 166 state-sensitive whitelisted files** (closure/state/phaser/given/gather/S12-accessors/S17-supply/currying) → **0 fail**; `S17-scheduler/every.t` + `state.t` + `closure.t` green in release.

Then deleted the field, both accessors (`get/set_closure_captured_state`), the call-entry override, and the exit persist (−45 lines net).

## Not touched (separate)

- `box_captured_lexicals` — cell creation is correctly placed at **capture** time. A tried-and-reverted attempt to relocate it to declaration time regressed `S04-blocks-and-statements/let.t` (boxing before the `let`-save defeats restore).
- `owned_captures` — read-only per-iteration loop value-freeze, still non-cell.

`make test` PASS (456 unit + 5336 prove), clippy clean, int.t release perf 0.07s. CI release roast is the authoritative final check.


tokuhirom:
Closing: `closure_captured_state` is NOT cleanly deletable. It deterministically regresses `roast/S10-packages/precompilation.t` test 51 ('closure is preserved after deserialization', GH2897). Root cause: a precompiled module's closure (`gen-counter` returns `sub counter { $foo++ }`) is created at BEGIN/precompile time and serialized to disk; the captured `$foo` ContainerRef cell does NOT survive cross-process serialization, so the deserialized closure relied on the `closure_captured_state` side table to persist its per-call state. The non-precompiled equivalent works fine via cells (0,1,2), but the deserialize path needs the side table. Verified main release PASS vs branch release FAIL. Reverting; closure_captured_state stays. Knowledge captured in memory + docs/container-identity.md.

PR #2842, opened 2026-06-10
feat(exceptions): X::Obsolete for the bare `<>` diamond

tokuhirom:
## Summary

A bare empty `<>` is the obsolete Perl input diamond, not an empty word list.
Raku rejects it at compile time with **X::Obsolete** (`old => "<>"`). mutsu
previously parsed `<>` as an empty `ArrayLiteral([])` and silently produced an
empty list. `parse_quote_word_list` now raises X::Obsolete when the bare-angle
`<...>` form has empty / whitespace-only content. The `«»` / `<<>>` quote forms
and the hash zen-slice `%h<>` (a different, postcircumfix parser) are unaffected,
as are all non-empty word lists.

Covers `roast/S32-exceptions/misc2.t:105`.

## Test plan

- New `t/obsolete-diamond.t` (5 subtests): `<>` throws; `<a b c>` / `<42>` /
  `%h<>` zen-slice still work.
- `roast/S02-literals/quoting.t` green; `make test` only the known-flaky
  `t/proc-async.t` differs.

### DIFF
--- a/src/parser/primary/container.rs
+++ b/src/parser/primary/container.rs
@@ -1042,8 +1042,9 @@ fn parse_quote_word_list<'a>(
     };
     let content = &input[..end];
     let rest = &input[end + close.len()..];
-    // A bare empty `<>` is the obsolete Perl diamond, not an empty word list.
-    if reject_lt_operators && content.trim().is_empty() {
+    // A bare *empty* `<>` is the obsolete Perl diamond. `< >` (with whitespace)
+    // is a legal empty `List`, so only the truly-empty form errors.
+    if reject_lt_operators && content.is_empty() {
         return Err(crate::parser::stmt::control::make_obsolete_error(
             "<>",
             None,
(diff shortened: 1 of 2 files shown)


tokuhirom:
Closing: the `<>` obsolete-diamond fix is correct, but `roast/t/fudge.t` (whitelisted) legitimately contains a bare `<>` (`do { local(@ARGV,$/)=$got; <> }`, lines 33-34) — Perl-5-style file slurping that even rakudo rejects ("Unsupported use of <>"). fudge.t only compiled on mutsu because of the lenient `<>`-as-empty-list bug this PR fixes; the correct behaviour makes fudge.t fail to compile (Tests:0, exit 1). Not worth breaking the whitelist invariant for one misc2 case (misc2.t:105). Abandoning per the 'never break a whitelisted test' rule.

tokuhirom:
Reopened: per maintainer guidance, shipping the correct `<>` X::Obsolete implementation and de-whitelisting `roast/t/fudge.t` (which contains the obsolete `<>` rakudo itself rejects — it was only passing on mutsu via the lenient-`<>` bug). Rebased on main; the `< >` empty-List fix is included.

PR #2862, opened 2026-06-10
feat(exceptions): X::Syntax::NegatedPair for an argument on a negated pair

tokuhirom:
## Summary

`:!foo(3)` supplies an argument to a negated colonpair, which Raku rejects at compile time:

```
===SORRY!=== Error while compiling
Argument not allowed on negated pair with key 'foo'
```

Previously mutsu parsed `:!foo` as a `False`-valued `Pair` and then tried to invoke it with `(3)`, yielding a runtime `No such method 'CALL-ME' for invocant of type 'Pair'`.

`colonpair_expr` now raises a fatal parse error carrying the structured `X::Syntax::NegatedPair` exception (with `key`) when a negated pair is immediately followed by `(`. A bare negated pair `:!foo` (no argument) stays a legal `False`-valued pair.

## Test

- New `t/negated-pair.t`
- Fixes `roast/S32-exceptions/misc2.t` test 52

### DIFF
--- a/src/parser/primary/misc.rs
+++ b/src/parser/primary/misc.rs
@@ -776,6 +776,16 @@ pub(in crate::parser) fn colonpair_expr(input: &str) -> PResult<'_, Expr> {
     // :!name (negated boolean pair)
     if let Some(after_bang) = r.strip_prefix('!') {
         let (rest, name) = parse_ident_with_hyphens(after_bang)?;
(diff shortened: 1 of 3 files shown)


tokuhirom:
Duplicate of #2859 which landed independently with the same X::Syntax::NegatedPair implementation. Closing in favor of the merged version.

PR #2880, opened 2026-06-10
feat(nil): complete S02-types/nil.t — `for Nil` iterates once + default lang version 6.d

tokuhirom:
Completes `roast/S02-types/nil.t` (67/67) and whitelists it.

## test 19 — `$x++ for Nil; $x == 1`
`for Nil { }` now runs exactly one iteration. `Nil` is a single scalar item in
list context (like `for Any { }`), not an empty list, so `value_to_list(Nil)`
returns `[Nil]` instead of `[]` (both `runtime/utils.rs` and `runtime/ops.rs`).

## test 26 — `subset MyInt of Int; my MyInt $x = 5; $x = Nil; $x === MyInt`
mutsu defaulted to language version **6.e**; rakudo defaults to **6.d** for a
pragma-less file. Under 6.e the subset-Nil-reset nominalizes to the base type
(`Int`), but the baseline test (and rakudo) expect the subset itself (`MyInt`).
Changed the per-parse default in `parser/stmt/simple.rs` and bare `use v6` in
`normalize_language_version` to **6.d**. This also fixes `$*RAKU.version` to
report `v6.d`, matching rakudo.

### Keeping 6.e-pragma files correct
Two version-gated behaviors read the *global* `current_language_version()` at
call time — a mutable thread-local that the parser resets to the default on
re-entry (e.g. while compiling a regex during a grammar parse). With the old
6.e default this was invisible; with 6.d it surfaced. Both are now keyed on the
type's **declaration** revision (stored as `type_metadata` `language-revision`):
- grammar `.parse`/`.parsefile` Failure (6.e+) vs Nil (pre-6.e) — `methods_grammar.rs`
- class DESTROY role-submethod dispatch — new `class.rs::type_decl_is_6e`

(BUILD submethod dispatch already used the declaration revision.)

`t/subset.t` updated to assert the 6.d behavior; the 6.e nominalization path
stays covered by `roast/S02-types/subset-6e.t`.

## Validation
- `roast/S02-types/nil.t`: 67/67 PASS
- `make test`: pass
- All whitelisted version-gated tests (S05-grammar, S12-*, S14-roles, 6.c/6.d,
  subset-6e/6c, fail-6e, SETTING-6e/6c, version.t, ...) PASS
- Local sweep of whitelisted S03/S04/S07 for-loop & operator tests: PASS


tokuhirom:
## Update: completing nil.t un-masked a chain of vacuously-passing tests

Completing test 19 (`for Nil` iterates once) + the `:=`-in-expression scope-leak fix made several whitelisted tests **actually run their bodies for the first time** — previously `plan +my @x := LIST` bound `@x` to Nil, so `for @x` ran zero (vacuous) iterations and the subtests were no-ops. The now-running tests exposed real pre-existing bugs, all fixed in the follow-up commit:

- **`:=` declaration leak in expression position** — `+my @x := 1,2,3`, `plan +my @x := ...`, `f(my $c := "lit")` now leak the `my` into the enclosing scope.
- **`Nil.grep`/`.map`/`.first`/`.sort`/…** now treat Nil as a one-item list (raku semantics) instead of absorbing to Nil.
- **`(my $o = $s).substr-rw(0,n) = $c`** lvalue-method writeback through a declaration target.
- **`open` IO forms** (S32-io/open.t): `open(IO::Handle)` IO() coercion; `IO::Path/IO::Handle.open` methods return Failure on error; `IO::Handle` methods discoverable via `.^lookup`; named flags `:append`/`:exclusive`(`:x`, O_EXCL)/`:update`/`:rw`/`:ra`/`:rx` honored; bare `:mode<rw>`/`:update` no longer auto-creates.

Now passing: S02-types/nil.t, S03-smartmatch/any-callable.t, S32-list/toggle.t, S32-num/{is-prime,real-bridge}.t, S32-io/open.t, S11-repository/cur-candidates.t. `make test` + targeted roast sweeps green.

PR #2963, opened 2026-06-13
refactor(value): Stage 1a — ArrayData wrapper for list-backing Vec

tokuhirom:
## What

Introduce `ArrayData { items, value_type, declared_type }` as the backing store for all list-like `Value` variants (`Array`, `Seq`, `HyperSeq`, `RaceSeq`, `Slip`) and `ArraySlotRef`, replacing the bare `Arc<Vec<Value>>`. `ArrayData` `Deref`/`DerefMut`s to the inner `Vec<Value>`, so the overwhelming majority of read/mutate sites are unchanged; only structural sites (construction, `Arc::as_ptr` interior mutation, explicit type annotations, `PtrKeyedMap` generics) are touched.

## Why

This is the mechanical foundation (mirrors HashData Stage 1a from #2952) for the **last** Arc-pointer-keyed side-table removal called out in PLAN.md Q2: embedding array container type metadata IN the value so it travels with COW, then deleting the `array_type_metadata` side table — the final intermittent Arc-pointer-identity flaky source.

**This PR is behavior-preserving.** The metadata fields exist but are NOT yet wired: `array_type_metadata` / `array_defaults` / shaped-array / grep-view `PtrKeyedMap`s remain authoritative. Stage 1b wires the fields and deletes the type-metadata side table.

## Critical detail (per the HashData layout-UB lesson)

Every `Arc::as_ptr(..) as *mut Vec<Value>` interior-mutation site was retargeted to `*mut ArrayData` (ops auto-deref through `DerefMut`); the two whole-Vec replacement sites (`*ptr = combined`, `Arc::new(new_items)`) operate on `.items` explicitly. `seq_*` lazy/cached/consumed pointer tracking and the PtrKeyedMaps now key on `Arc<ArrayData>` — pointer identity is preserved.

## Testing

- `make test` (727 files, 6588 tests) green.
- Whitelisted array/set/bag/mix/typed-array/seq/gather/range tests pass locally.
- `cargo clippy -- -D warnings` clean.

### DIFF
--- a/src/builtins/comb.rs
+++ b/src/builtins/comb.rs
@@ -111,5 +111,5 @@ pub(crate) fn native_comb_method(
     };
     let matcher = positional.first().copied();
 
-    comb_pure(&text, matcher, limit).map(|items| Ok(Value::Seq(std::sync::Arc::new(items))))
+    comb_pure(&text, matcher, limit).map(|items| Ok(Value::Seq(std::sync::Arc::new(items.into()))))
 }
--- a/src/builtins/functions.rs
+++ b/src/builtins/functions.rs
@@ -174,7 +174,7 @@ fn native_sprintf(args: &[Value], z_mode: bool) -> Option<Result<Value, RuntimeE
     let flattened: Vec<Value>;
     let actual_args = if rest.len() == 1 {
         if let Value::Array(items, ..) = &rest[0] {
-            flattened = items.as_ref().clone();
+            flattened = items.as_ref().to_vec();
             &flattened[..]
         } else {
             rest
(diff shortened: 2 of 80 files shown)


tokuhirom:
Superseded by #2962, which already landed the full ArrayData migration (wrapper + embedded array type metadata + deletion of the `array_type_metadata`/`array_defaults` Arc-ptr-keyed side tables) — i.e. this PR's Stage 1a plus the intended Stage 1b, in one. A parallel worker reached it first. Nothing here is missing from main, so closing without a rebase. Local branch deleted.

PR #3147, opened 2026-06-16
fix(hash): flatten bare %h in list/hash context; keep $-sourced hashes itemized

tokuhirom:
## Summary

`my %c = (%h,)` and `%(%h,)` raised *"Odd number of elements found where hash
initializer expected"* and **aborted**, because mutsu couldn't distinguish a bare
`%h` (which flattens into pairs in hash context) from a `$`-sourced itemized hash
`$hashitem` (which stays opaque — Raku raises `X::Hash::Store::OddNumber`).

Both decontainerized to a bare `Value::Hash`, so the old heuristic — flatten only
when the list had **>1 element** — was wrong in *both* directions: it missed
`(%h,)` and wrongly flattened `($hashitem, :c(42))`.

## Fix (container itemization)

- `Interpreter::itemize_value` (the `Itemize`/`ItemizeVar` opcodes) now wraps a
  `$`-sourced Hash in `Value::Scalar`, mirroring `Seq`, making an itemized hash
  representationally distinct from a bare `%h`.
- `build_hash_from_items` — the single chokepoint for `%m = (...)` **and** `.hash`
  / `%(...)` — flattens a bare `Value::Hash` element into its pairs (preserving
  object-hash keys via `typed_key`), while leaving `Value::Scalar(Hash)` opaque.
- Dropped the redundant element-count heuristic in the hash-assignment path.

This is the itemization discriminator the previous attempt (#2958, reverted)
lacked — that PR added an unconditional flatten arm and regressed
`assigning-refs.t`. With producer-side itemization, the flatten is now correctly
gated.

## Impact

Unblocks `S09-hashes/objecthash.t` — was aborting at test 37 (ran 36/62), now
runs the full plan with only object-hash *type-enforcement* subtests remaining
(follow-up PRs). Resolves the long-standing `list-to-hash-flatten-itemization`
blocker.

## Tests
- New `t/hash-from-list-itemization.t` (14).
- `assigning-refs.t` / `hash.t` / `S09-typed-arrays/hashes.t` green; full `t/`
  suite 7850 pass; `cargo test` 462 pass.

### DIFF
--- a/src/runtime/utils.rs
+++ b/src/runtime/utils.rs
@@ -654,6 +654,21 @@ pub(crate) fn build_hash_from_items(items: Vec<Value>) -> Result<Value, RuntimeE
             Value::Pair(key, boxed_val) => {
                 map.insert(key, *boxed_val);
             }
+            // A bare hash in list context flattens into its key=>value pairs
+            // (`%m = (%h,)` / `%(%h,)`). A hash sourced from a `$` scalar is
+            // itemized as `Value::Scalar(Hash)` (see `Interpreter::itemize_value`)
+            // and does NOT match here, so it stays an opaque single element —
+            // matching Raku, where `%m = ($hashitem,)` dies "Odd number".
+            // Object-hash keys are preserved via `typed_key`.
+            Value::Hash(h) => {
+                for (k, v) in h.iter() {
+                    let key_obj = h.typed_key(k);
+                    if !matches!(&key_obj, Value::Str(_)) {
+                        original_keys.insert(k.clone(), key_obj);
+                    }
+                    map.insert(k.clone(), v.clone());
+                }
+            }
             Value::ValuePair(key, boxed_val) => {
                 let str_key = Value::hash_key_encode(&key);
                 if !matches!(key.as_ref(), Value::Str(_)) {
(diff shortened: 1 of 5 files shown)


tokuhirom:
Closing without merging. The producer-side itemization (wrapping a `$`-sourced Hash in `Value::Scalar`) is the correct discriminator and fixed `(%h,)` / `%(%h,)` while keeping `($hashitem,)` dying — but a universal `Scalar(Hash)` **leaks** into every consumer that matches `Value::Hash` and may receive a `$`-hash fetched from a list (the "deref everywhere" anti-pattern). CI surfaced these one-by-one as new whitelisted regressions, not converging cheaply:

- `set_contains` (`(elem)`/`∈`) → `S03-operators/set_elem.t` (patched).
- Hash-as-callable mapper in a `for`-list → `S32-list/categorize.t` 13-18.
- object-key preservation in the flatten → `S32-hash/map.t` 15.

The real prerequisite is the container-identity **"decont at value-operand fetch"** single chokepoint (PLAN.md 🟣 / docs/container-identity.md §3). Without it, the producer wrap needs unbounded per-consumer deconts. **Do the decont-at-fetch foundation first, then resurrect this branch** (the full working producer+consumer diff is preserved here). Findings recorded in memory `list-to-hash-flatten-itemization`.

PR #3372, opened 2026-06-21
fix(vm): resume a warn raised deep in a callee under an outer CONTROL handler

tokuhirom:
## Summary

A resumable `warn` raised **deep** in a chain of calls (e.g. `render → format → get → Logger.log → &warn.()`) could not be resumed by an enclosing `CONTROL {}`: the signal unwound to the installing block, whose frame-relative `resume_ip` pointed into the **callee's** `CompiledCode`, so resuming executed garbage and crashed (`No such method 'CALL-ME' for invocant of type 'Any'`) or silently truncated the computation.

```raku
my @seen;
CONTROL { default { @seen.push(.message); .resume } }
sub inner { warn "deep"; 42 }
sub outer { inner() + 1 }
outer();         # was: crash/abort   now: @seen=["deep"], outer() == 43
```

## Fix

Handle a resumable warn **inline at the per-op chokepoint** (`exec_one`, which every exec loop already routes through), in the deepest still-live frame where the op recorded a frame-correct `resume_ip`:

- New `Interpreter::control_handlers` stack mirrors `control_handler_depth`; each entry records the installing block's `CompiledCode` pointer, its CONTROL body range, and `call_frames.len()` at install time.
- `handle_warn_inline` runs the innermost handler body (or prints, when none) and returns the resume ip. Crucially it temporarily **swaps the installing frame's locals** (`call_frames[frame_depth].saved_locals`) into `self.locals` so the handler's bytecode addresses the right local slots and its mutations land where the installing frame sees them after the callee returns. Outer lexicals reached by name go through the (dynamic, shared) env and need no swap.
- `exec_one` now wraps `exec_one_dispatch`: a warn from the op is handled inline here, so every exec loop (run loop, `run_range`, named-sub, method, closure, …) gets correct cross-frame resumption with no per-loop wiring.

**SAFETY:** the installing block's `CompiledCode` is an ancestor frame on the live Rust call stack while its `control_handlers` entry exists, so the raw pointer never dangles.

## Impact

- **Template::Mustache** `render(:log-level<Warn>)` no longer crashes on a missing field (`06-logging` 0 → 2 of 3); stops the cross-frame-warn crash in `12-inheritence`/`13-pragmas`/`50-readme` (they now fail cleanly on their *own* remaining features instead of aborting).

## Test

`t/warn-deep-resume-under-control.t` — 6 cases (warn 1/2 levels deep, indirect `&warn.()`, uncaught deep warn), spec-validated against `raku`. `make test` green (9702 tests).

### DIFF
--- a/src/runtime/mod.rs
+++ b/src/runtime/mod.rs
@@ -896,6 +896,20 @@ pub struct Interpreter {
     /// EVAL) can observe handlers installed by the outer VM and propagate
     /// warn/control signals appropriately.
     pub(crate) control_handler_depth: u32,
+    /// Active CONTROL handler bodies, innermost last. Each entry is
+    /// `(code_ptr, control_begin, control_end, frame_depth)` where `code_ptr` is a
+    /// raw `*const CompiledCode` (kept as `usize` to stay `Send`) of the block
+    /// that installed the handler, and `frame_depth` is `call_frames.len()` at
+    /// install time. Pushed/popped in lockstep with `control_handler_depth`
+    /// (`exec_try_catch_op_inner`). Lets a resumable `warn` raised *deep* in a
+    /// callee be handled in that callee's still-live frame — running the handler
+    /// inline (with the installing frame's locals temporarily swapped in, found at
+    /// `call_frames[frame_depth].saved_locals`) and resuming via the callee
+    /// frame's own `resume_ip` — instead of unwinding to the installing block,
+    /// whose frame-relative `resume_ip` would point into the wrong code. SAFETY:
+    /// the installing block's `CompiledCode` is an ancestor frame on the live Rust
+    /// call stack while any entry is live, so the pointer never dangles.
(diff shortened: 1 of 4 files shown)


tokuhirom:
Closing: the inline cross-frame warn-resume approach is fundamentally incompatible with full CONTROL phaser semantics. It correctly resumes a deep warn whose handler does `.resume` (or a bare `default {}` fall-through), but a `CONTROL { when CX::Warn {...} }` matches via `succeed` (which must EXIT the installing block, not resume) — and because the warn is handled inline in the deep callee frame, there is no longer an `exec_try_catch_op` on the stack to catch that succeed, so it unwinds past the block and aborts. This regresses roast S32-basics/warn.t, S04-exception-handlers/control.t, S32-str/indent.t. A correct fix needs a real continuation / unified-frame mechanism (mutsu has no single frame stack), which is out of scope here. The same-frame indirect-warn resume (#3370) and the slurpy-builtin arity fix (#3373) land the safe parts.

PR #3471, opened 2026-06-23
fix(metaops): cross (X) and zip (Z) return a Seq, not a List

tokuhirom:
## Summary

`X` and `Z` (and their reduction / `Z=>` forms) returned a `List`, so `.^name`
was `List` and `.raku` rendered without the trailing `.Seq`:

```
$ raku  -e 'say (^3 X ^2).^name'   # Seq
$ mutsu -e 'say (^3 X ^2).^name'   # List

$ raku  -e 'say (1,2 Z 3,4).raku'  # ((1, 3), (2, 4)).Seq
$ mutsu -e 'say (1,2 Z 3,4).raku'  # ((1, 3), (2, 4))
```

## Fix

The binary and n-ary X/Z paths wrapped their non-empty results in
`Value::array` (a List) while the empty case already returned a `Seq`. Wrap the
non-empty result in `Value::Seq` too, in all four sites
(`vm_string_regex_ops.rs`). The inner tuples (`(0, 0)`, ...) stay plain Lists,
matching Rakudo, and lazy-input results still produce a `LazyList`.

## Test

- `t/cross-zip-seq.t` — 14 cases (X/Z `.^name`/`.raku`, n-ary, empty,
  reduction, `Z=>`, lazy side, `@`-reification stays Array).
- Swept the whitelisted X/Z roast tests (cross.t, zip.t S03/S17/S32, infix.t)
  plus 107 S02/S03 files using X/Z — no regressions.

### DIFF
--- a/src/runtime/registration_sub.rs
+++ b/src/runtime/registration_sub.rs
@@ -1059,7 +1059,7 @@ impl Interpreter {
                 let v = self.eval_block_value(&[Stmt::Expr(expr.clone())])?;
                 let raw_items: Vec<Value> = match &v {
                     Value::Array(items, _) => items.as_ref().clone().items,
-                    Value::Slip(items) => items.as_ref().clone(),
+                    Value::Seq(items) | Value::Slip(items) => items.as_ref().clone(),
                     Value::Hash(map) => map
                         .iter()
                         .map(|(k, v)| Value::Pair(k.clone(), Box::new(v.clone())))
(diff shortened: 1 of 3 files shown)


tokuhirom:
Pushed a follow-up: `X`/`Z` returning a Seq regressed `S12-enums/basic.t` (subtests 39/46/49 — `my enum E (<a b> Z=> 1,2)` etc.), because `register_enum_decl`'s dynamic-body handler unwrapped Array/Slip/Hash but treated a Seq as one opaque item. Added Seq alongside Slip so enum-from-Z/X works again. S12-enums/basic.t now passes (54/54).

PR #3546, opened 2026-06-24
feat(multi-dispatch): OTF-compile default-param candidates (ledger §D)

tokuhirom:
Continues the §D multi-dispatch VM-ization (`docs/vm-interpreter-fallback-ledger.md`, `PLAN.md` §2-D). After #3543 (where-constrained candidates), candidates with a default parameter value still fell back to the tree-walk interpreter because `def_is_otf_compilable` excluded any param with a `default`.

## What

Remove the `default.is_none()` clause from `def_is_otf_compilable` so default-param multi/proto/single candidates are OTF-compiled instead of tree-walking.

## Why it's byte-identical

The compiled binding path (`call_compiled_function_named` → `bind_function_args_values`) already evaluates defaults exactly as the interpreter does. Verified against `raku`:

- literal defaults, defaults referencing an earlier positional param, a method call on an earlier param, a top-level closure variable, and a lexical of an enclosing sub (nested-scope) all match;
- `nextsame` passes the **original** args, so a deferred candidate's default falls back to **its own** default (matches raku);
- recursion through a default-bearing multi works.

The pre-existing `Nil` result for a default that references an *escaped* closure lexical (a multi returned as an `&g` Callable and called after its defining scope exits) is **unchanged** — a separate, pre-existing gap in closure-env capture for multi-sub Callables, not introduced here (confirmed identical with the change stashed).

## Measured

default-multi fallback **85.7% → 0%** (via `MUTSU_VM_STATS`).

## Tests

- pin `t/multi-default-otf-dispatch.t` (16, raku/mutsu parity).
- `make test` PASS (11287 tests).
- All S06 / S12 whitelist roast tests green (run locally).


tokuhirom:
Closing — deferred. The simple `default.is_none()` removal passes `make test` + all S06/S12 whitelist locally, but CI's release `make roast` regressed three subprocess/Proc/module tests: `S02-magicals/env.t`, `S29-os/system.t`, `S11-repository/cur-current-distribution.t`.

**Root cause** (preserved for a future, narrower attempt): `run` is a core builtin, but `use Test::Util` brings `our sub run( Str $code, Str $input = '', *%o )` (a default-param sub) into scope. Before this change, that candidate was not OTF-compilable (it has a default param), so it fell back to the interpreter, which dispatched correctly between core `run` and Test::Util's `run` based on args. With the change, the **builtin-shadow path** (`user_function_matches_call` → `is_builtin_function("run")`) OTF-compiles Test::Util's `run` and caches it in the name-keyed `otf_call_cache`; under a stateful context (a `subtest` callback), a subsequent core `run` call with a different arity reuses the cached compiled function and mis-binds, aborting the `subtest` block early (`1..0` / `1..1`). This is the default-param analogue of the builtin-shadow hazard PR-2 guarded with `is_builtin_function`.

**Safe direction**: keep the single-candidate builtin-shadow path excluding default-param subs (fall back as before), and only relax default-param OTF for genuine multi candidates + non-builtin single OTF — but narrowing needs a full roast run to validate.

The implementation + pin (`t/multi-default-otf-dispatch.t`, 16 cases, raku/mutsu parity) are preserved on branch `multi-dispatch-default-otf` for reference. The where-constrained-candidate slice (#3543) is the clean §D win and is already merged.

PR #3884, opened 2026-06-28
perf(dispatch): OTF-compile a default-param single sub that shadows a builtin

tokuhirom:
## What

A single user sub with a **default parameter** that **shadows a builtin** was excluded from on-the-fly (OTF) bytecode compilation (`def_is_otf_compilable`), so it fell back to the tree-walk interpreter on every call (PLAN §2-D, remaining multi-dispatch fallback).

## Why it's now safe

The exclusion guarded the PR #3546 name-cache hazard: Test::Util's `our sub run(Str, Str = '')` shadows the `run` builtin, and pinning the user sub in `otf_call_cache` could mis-bind a later builtin `run` (process spawn).

That hazard is **closed by the cache's scope coherence**: `otf_call_cache` is gated on `fn_resolve_gen`, which bumps when the shadowing scope changes (import pop / lexical scope exit), evicting the user sub so a later builtin call re-resolves correctly. Within a single scope a name resolves to one routine, so no same-scope mis-bind is possible.

Verified:
```raku
{ sub run($c, $i = '') { ... }; run("first"); }   # user sub
my $proc = run("echo", "hi", :out);               # builtin process run — re-binds
```
binds the user sub inside the block and the builtin process `run` outside, matching raku.

## Tests

- `t/default-param-builtin-shadow-otf.t` (new, 5 subtests): shadowing sub inside scope, builtin re-bind outside, same-scope default application, non-shadowing regression guard.
- `make test` green (12934 tests, Result: PASS).

### DIFF
--- a/src/vm/vm_call_func_ops.rs
+++ b/src/vm/vm_call_func_ops.rs
@@ -1571,12 +1571,20 @@ impl Interpreter {
         // resolved def already satisfies it and the compiled binding just binds the
         // callable — byte-identical to the interpreter. Any cross-candidate
         // ambiguity (`&c:(Int)` vs untyped `&c`) is a *resolution*-level gap that
-        // fires identically with or without OTF, so it is unaffected. Only a param
-        // with a default value stays excluded here (the name-cache-pollution /
-        // builtin-shadow hazard, PR #3546 — allowed at genuine multi sites via
-        // `def_is_otf_compilable_multi_candidate`).
+        // fires identically with or without OTF, so it is unaffected.
+        //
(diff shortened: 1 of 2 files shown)


tokuhirom:
Reverting: the default-param relaxation regresses make roast. `def_is_otf_compilable` is shared across 4 dispatch sites (proto/multi/builtin-shadow), so relaxing defaults there broke proto/multi default handling broadly — caught by CI's release-build roast (my local `make test` did not run `make roast`). A narrower, site-specific gate would be needed; deferring as low-value (niche builtin-shadow-with-default subs).

PR #4105, opened 2026-07-03
feat(hyper): all-candidates .+/.* dispatch across builtin MRO levels

tokuhirom:
Fixes **S03-metaops/hyper.t subtest 407** (`».+`/`».*` respect nodality — all 20 assertions).

## Change
`.+method`/`.*method` on a builtin returned a single result. Rakudo returns one per MRO level declaring the method (`<a b>.+elems` → `(2, 2)`: List.elems + Any.elems).
- New `builtin_method_mro_level_count`: count DISTINCT per-level native method tables along the receiver's MRO that declare the method, deduped by contents (Array/List share LIST_METHODS → count once). Data-driven, no hardcoded number.
- Applied to every `.+`/`.*` native fast-path (non-hyper + static/dynamic hyper arms).
- The per-element `.+`/`.*` result is now a **List** `(2, 2)`, not an Array, matching Rakudo. Updated `t/parallel-dispatch-regression.t` to the raku-verified `[(2,), (4,)]`.

## Scope / honesty
- **Does NOT whitelist hyper.t**: subtest 408 (`&` sub-call run-count) is a separate full-file heisenbug (passes in isolation).
- **Known limit**: coercion methods declared at 3+ tables via a shared block (`.Str`) over-count, and a single native impl can't reproduce per-level-different results; the count is sound for same-result methods (elems/abs). A fully general fix needs per-level-own builtin method tables (the architectural gap).
- New `t/hyper-all-candidates-builtin.t` (7/7); cross/zip/reverse roast clean; make test green.

CI's make roast is the net for the over-count blast radius.

### DIFF
--- a/src/runtime/methods_classhow_mro.rs
+++ b/src/runtime/methods_classhow_mro.rs
@@ -2,7 +2,7 @@ use super::*;
 use crate::symbol::Symbol;
 
 impl Interpreter {
-    pub(super) fn classhow_mro_names(&mut self, invocant: &Value) -> Vec<String> {
+    pub(crate) fn classhow_mro_names(&mut self, invocant: &Value) -> Vec<String> {
         let class_name = match invocant {
             Value::Package(name) => name.resolve(),
             Value::Instance { class_name, .. } => class_name.resolve(),
--- a/src/runtime/methods_signature_shaped.rs
+++ b/src/runtime/methods_signature_shaped.rs
@@ -254,7 +254,11 @@ impl Interpreter {
             }
         }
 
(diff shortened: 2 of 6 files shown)


tokuhirom:
Redundant with #4104 (roast:1), which already landed the identical builtin-MRO all-candidates `.+`/`.*` dispatch — 407 passes on main with the List inner type. Independent convergence: both PRs found the same per-level-count solution and both identified subtest 408 (`&` sub-call run-count) as a separate full-file heisenbug that blocks whitelisting hyper.t. Closing in favor of #4104. My t/hyper-all-candidates-builtin.t guard is available if useful but #4104's coverage suffices.

PR #4111, opened 2026-07-03
GC Phase B step 4: Gc<T> + Bacon-Rajan cycle collector (default-off, Miri-clean)

tokuhirom:
Implements **step 4** of `docs/gc-level1-detailed-design.md` §11 (ADR-0001 / ADR-0002, Phase B): the minimal cycle-collection machinery. Steps 1-3 (root visitor, child visitor, `MUTSU_VM_STATS` counters) already landed; this adds the collector itself.

## What's added (`src/gc/collect.rs`)
- **`Gc<T>`** — an `Arc`-like refcounted pointer (clone/deref/drop) with a GC header (atomic strong count + color + buffered flag). `Send + Sync` when `T` is, matching the existing `Arc`-based multi-thread model.
- **`Trace` / `GcNode` traits** — type-erased child enumeration so the collector recurses over `NonNull<dyn GcNode>` without knowing each node's concrete type.
- **Thread-local candidate buffer (`ROOTS`)** + **synchronous Bacon–Rajan** trial-deletion collector `collect_cycles` (`mark_gray` → `scan`/`scan_black` → `collect_white`). A `GC_FREEING` guard makes a freed cycle's reentrant `Gc::drop` inert, so the whole cycle is freed without double-free.

## Design-faithful scoping (design §3.1 / §9.1)
- **No `Value` variant is wired to `Gc<T>` yet** — that is step 5 (the `Array`/`Hash`/`ContainerRef` first wave). So this is **default-off and dead code in production**, exercised only by unit tests. The `record_gc_candidate_push` / `record_gc_collection` counters are now called from the collector; they will report nonzero once step 8 wires collection into a `MUTSU_GC`-gated safepoint.

## Bugs found & fixed during bring-up
1. **Deadlock** — the collector recursed while `Trace` held a node's child-collection lock (`Mutex`/`RwLock`, e.g. the future `Value::ContainerRef`), re-locking the same node on revisit. Fixed by gathering child pointers into a `Vec` first (`children_of`), then recursing lock-free.
2. **Stacked-Borrows UB (caught by Miri)** — dropping the value through a `*mut` derived from a shared `&self` is invalid. Fixed by freeing via `drop_in_place(node.as_ptr())` on the raw erased box pointer (a pure `*mut` op).

## Validation
- 5 cycle-reclaim unit tests: self-cycle, 2-node, 5-node ring, acyclic-no-leak, live-external-ref-kept.
- **Whole `gc` module green under Miri (Stacked Borrows)** — `cargo +nightly miri test --lib gc::collect`.
- `cargo clippy --all-targets -- -D warnings` clean; `cargo fmt` applied.
- Moved `src/gc.rs` → `src/gc/mod.rs` to host the `collect` submodule.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -383,13 +383,14 @@ MIME::Base64 1.2.5（#3427）/ IO::Blob（builtin 型サブクラスの user ove
       2. ✅ child visitor 導入（`Value::visit_gc_children()` ＋ `ArrayData`/`HashData`/`SubData`/
          `InstanceAttrs`/`SharedPromise`/`SharedChannel`）
       3. ✅ `MUTSU_VM_STATS` の GC カウンタ枠（`vm/vm_stats.rs`: collections/candidate_pushes/
-         dedup_hits/reclaimed_nodes/reclaimed_cycles/pause_ns_total/pause_ns_max/roots_scanned。
-         全て 0 固定 — 呼び出し元は §11 step 4 以降）
-      4. `Gc<T>` / node header / candidate buffer の最小実装 ← **次はここ（次セッション開始点）**。
-         設計メモ §5.1（node header: strong refcount 相当・color/state・buffered フラグ・
-         `trace(&mut Visitor)` vtable 相当）と §1（Level 1a=同期・協調 collector、safepoint は
-         再入境界に限定）を先に読むこと。まだ `Value` のどの variant も `Arc→Gc` に置換しない
-         （それは step 5 以降の first wave）。
+         dedup_hits/reclaimed_nodes/reclaimed_cycles/pause_ns_total/pause_ns_max/roots_scanned）
+      4. ✅ `Gc<T>` / node header / candidate buffer / 同期 Bacon–Rajan collector の最小実装
+         （`src/gc/collect.rs`）。`Gc<T>`（Arc 相当の clone/deref/drop＋GC header: strong/color/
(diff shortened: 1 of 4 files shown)


tokuhirom:
Superseded: a sibling PR landed GC Level 1a step 4's type machinery (Arc-based `Gc<T>`/`Trace`/candidate buffer in `src/gc/gc_ptr.rs`) on main first. This PR used an incompatible manual-allocation foundation and conflicts on the same files.

Preserving the knowledge from this PR's working collector for whoever implements the collector (step 8):
1. **Deadlock hazard**: a trial-deletion collector must NOT recurse while `Trace::trace` holds a node's child-collection lock (main's `Value::visit_gc_children` locks a Mutex for `ContainerRef`; a node holding `Mutex<Vec<Gc>>` does too). Revisiting a node whose lock is held up-stack deadlocks a non-reentrant lock. Fix: gather child pointers into a Vec first, release the lock, then recurse.
2. **Stacked-Borrows**: main's Arc-based `GcBox` sidesteps the manual-free UB this PR hit (dropping through a `*mut` derived from a shared `&self`); on Arc, reclaim breaks cycle edges and lets Arc free. Validate the eventual collector under `cargo +nightly miri test` (it caught the UB here).

The synchronous Bacon–Rajan algorithm (mark_gray → scan/scan_black → collect_white) with 5 cycle-reclaim tests is in this PR's `src/gc/collect.rs` if useful as a reference for step 8.

PR #4119, opened 2026-07-03
GC step 5e: migrate Value::ContainerRef from Arc to Gc<Mutex<Value>>

tokuhirom:
Fourth first-wave `Arc → Gc` container migration (ADR-0001 layer 3a, `docs/gc-level1-detailed-design.md` §11 step 5e), completing the scalar/element-cell wave after Hash (#4113), Array (#4116), and Set/Bag/Mix (#4117).

`Value::ContainerRef` — the shared mutable cell behind `:=` element/scalar binds, take-rw, and captured-outer lexicals — now holds a `Gc<Mutex<Value>>`. Cells are the most cycle-prone `Value`s (`%h<k> := %h`, mutually-referential bound scalars), so this is the payoff variant for the whole first wave.

## Behavior-preserving
The cell is mutated through `.lock()` (not COW / `make_mut`), which works unchanged via `Gc`'s `Deref` to `Mutex<Value>`. `Gc<Mutex<Value>>` is the same shared-cell model as `Arc<Mutex<Value>>`, just with the GC node header. Candidate buffering only under `MUTSU_GC=on` (default off).

## Changes
- `impl Trace for std::sync::Mutex<Value>` — locks and traces the held `Value`'s GC children (`Trace` is a local trait so the impl is allowed; it only runs at a collect safepoint, so it can't deadlock) + a `Value::ContainerRef` arm in `gc_trace`.
- Mechanical call-site rewrite (~30 files): Value-cell `Arc::new(Mutex::new(v))` → `Gc::new`, `Arc::ptr_eq(cell)` → `crate::gc::Gc::ptr_eq`, and `Arc<Mutex<Value>>` (incl. `std::sync::Mutex`-qualified) annotations → `Gc<…>`. **Non-Value `Mutex` cells (String buffers, `Option<ChildStdin>`, `TcpStream`, ...) correctly stay `Arc`.**

## Testing
- `:=` element-bind aliasing verified (`$x := @a[0]; $x = 99` mutates `@a`; `$y := @a[1]; $y++` too).
- 12 cell/bind local tests pass; 24 `gc::` unit tests green; **Miri (Stacked Borrows) green**; `cargo clippy --all-targets -- -D warnings` clean; `make test` green (14076 tests, 0 failures).

With this, the entire §3.1 container/cell type filter (Hash/Array/Set/Bag/Mix/ContainerRef) is `Gc`-managed. Next: second wave (`Sub`/`Instance`, step 9) and wiring the synchronous collector at a safepoint (step 8).

### DIFF
--- a/src/runtime/builtins_collection_deepmap.rs
+++ b/src/runtime/builtins_collection_deepmap.rs
@@ -224,7 +224,7 @@ impl Interpreter {
         block: &Value,
         leaf: &Value,
     ) -> Result<(Value, Value), RuntimeError> {
-        let cell = std::sync::Arc::new(std::sync::Mutex::new(leaf.clone()));
+        let cell = crate::gc::Gc::new(std::sync::Mutex::new(leaf.clone()));
         let res = self.call_sub_value(
             block.clone(),
             vec![Value::ContainerRef(cell.clone())],
--- a/src/runtime/builtins_system_run.rs
+++ b/src/runtime/builtins_system_run.rs
@@ -176,7 +176,10 @@ impl Interpreter {
                     let in_pipe = if let Some(stdin) = stdin_handle {
                         // Store in proc_stdin_map for .print/.close
                         if let Ok(mut map) = super::native_methods::proc_stdin_map().lock() {
-                            map.insert(pid as u32, Arc::new(std::sync::Mutex::new(Some(stdin))));
+                            map.insert(
+                                pid as u32,
+                                std::sync::Arc::new(std::sync::Mutex::new(Some(stdin))),
+                            );
                         }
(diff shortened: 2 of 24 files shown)


tokuhirom:
Superseded by #4118, which landed the identical `Value::ContainerRef → Gc<Mutex<Value>>` migration on main first (parallel work — same recipe, same result: `impl Trace for Mutex<Value>`, the `Value::ContainerRef` gc_trace arm, Value-cell `Arc::new(Mutex::new)` → `Gc::new` with non-Value `Mutex` cells kept as `Arc`). main builds cleanly and `make test` is green with #4118, so there is nothing to add here. Closing.

PR #4149, opened 2026-07-04
fix(str): reject non-literal fraction strings in numification

tokuhirom:
## Problem

`+"3/2"` correctly numifies to `3/2`, but the fraction parser also accepted forms that are **not numeric literals**:
- signed denominator: `"3/-2"`, `"3/+2"`, `"-3/-2"`
- decimal part in either operand: `"3.0/2"`, `"3/2.0"`, `"3.5/2"`

Rakudo fails all of these — roast `S32-str/numeric.t` marks them `#?rakudo todo` (*"Unsure of what val() should accept"*) and asserts they throw.

## Fix

Tighten `try_parse_fraction`: a fraction string is only `[sign]INT/INT` — a plain-integer numerator over an **unsigned** plain-integer denominator (the leading sign of the whole value is already stripped into `sign`). Reject a `.` in either operand or a sign on the denominator. Underscores and the accepted `"3/2"`/`"-3/2"` forms are unchanged.

## Result

With the earlier lone-i fix (#4147), roast `S32-str/numeric.t` now has **0 raw failures** (it was already whitelisted via the fudged `todo`s, so no whitelist change — a genuine numeric-parsing correctness improvement matching the roast spec).

`make test` green (14179); clippy clean.

Tests: `t/str-numeric-fraction-strict.t` — accepted `[sign]INT/INT` (incl. underscores) and the rejected signed-denominator / decimal-operand forms.

### DIFF
--- a/src/runtime/str_numeric.rs
+++ b/src/runtime/str_numeric.rs
@@ -58,7 +58,8 @@ pub(crate) fn parse_raku_str_to_numeric(input: &str) -> Option<Value> {
         return Some(apply_sign(v, sign));
     }
 
-    // Try fraction `N/D` (with optional decimal parts and signed denominators)
+    // Try fraction `N/D` (a plain-integer numerator over an unsigned
+    // plain-integer denominator; the leading sign is already in `sign`).
     if let Some(v) = try_parse_fraction(body, sign) {
         return Some(v);
     }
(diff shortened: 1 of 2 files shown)


tokuhirom:
Closing: this strict-fraction change breaks roast S32-str/val.t (588 failures). val.t requires *lenient* fraction parsing — e.g. `val('-4_2.1_5/−1_5.4_5')` must yield a RatStr with a decimal numerator/denominator and signed denominator — which is the exact opposite of numeric.t's `#?rakudo todo` expectation that `+'3/-2'` fails.

`+$str` (Numeric coercion, strict) and `val()` (allomorph, lenient) legitimately differ in what they accept, but mutsu routes both through the single `parse_raku_str_to_numeric`, so tightening it for one breaks the other. numeric.t itself marks these `#?rakudo todo` ("Unsure of what val() should accept"), reflecting the genuine ambiguity. Correctly resolving it needs separate strict/lenient parse paths — a larger change — not a blanket tightening. The lone-i fix (#4147) is unaffected and already merged.

PR #4273, opened 2026-07-05
fix(autoviv): nested `<>`/`{}` ++/--/OP= autovivifies a Hash, not an Array

tokuhirom:
## What

`%h<a><b>++`, `%h<a><b> += 1`, and the classic nested-counter idiom
`%h<counts>{$k}++` autovivified the intermediate `%h<a>` as an *Array*
(`{:a($[])}`) instead of a Hash (`{:a(${:b(1)})}`), so the update was lost.
The `%h<a><b> = value` plain-assignment path already got this right.

## Why

The write-back these desugarings emit hardcoded `is_positional: true` on the
`IndexAssign`, which forces Array autovivification regardless of the actual
subscript. This threads the real `is_positional` through instead — in the
nested `++`/`--` compiler helpers (`expr_postfix.rs`) and the
compound-assignment desugaring (`compound_expr.rs`) — so an associative
subscript autovivifies a Hash and a positional one an Array. The
intermediate-read stays associative-lenient so a positional `@a[0][1] += 1`
still autovivifies correctly rather than erroring on the not-yet-existing
element.

Matches raku:

```
my %h; %h<a><b>++;                              # {:a(${:b(1)})}
my %h; %h<a><b> += 5;                           # {:a(${:b(5)})}
my %h; %h<counts>{"x"}++; %h<counts>{"x"}++;    # {:counts(${:x(2)})}
my @a; @a[0][1] += 3;                           # [[Any, 3],]
```

## Test

Regression test: `t/nested-assoc-incdec-autoviv.t`. `make test` passes
(15227 tests, all green).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -417,7 +417,7 @@ checksum = "f8ca58f447f06ed17d5fc4043ce1b10dd205e060fb3ce5b979b8ed8e59ff3f79"
 
 [[package]]
 name = "mutsu"
-version = "0.2.18"
+version = "0.2.14"
 dependencies = [
  "bincode",
  "console_error_panic_hook",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -417,7 +417,7 @@ checksum = "f8ca58f447f06ed17d5fc4043ce1b10dd205e060fb3ce5b979b8ed8e59ff3f79"
 
 [[package]]
 name = "mutsu"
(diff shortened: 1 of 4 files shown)


tokuhirom:
Duplicate of #4264 (already merged). Same branch fix/nested-assoc-incdec-autoviv.

PR #4323, opened 2026-07-07
fix(traits): container-level role mixin on attributes (S14-traits/attributes.t)

tokuhirom:
## Summary

Implements `\$attr.container.VAR does Role(arg)` inside a user `trait_mod:<is>`, so a parameterized role mixed into an attribute's container reaches the per-instance container. This brings `roast/S14-traits/attributes.t` from 4/8 to **8/8** (whitelisted).

## Root cause

`Attribute.container` returned `Nil`, so the `does` inside the trait mixed the role into a throwaway value that never propagated to instances.

## Implementation

- **`.container` returns a proxy** `Instance` tagged with `(owner, name, sigil)`. `.VAR`/`.self`/`.item` return the proxy itself (so the tag survives to the `does`); every other method delegates to a real inner container, so introspection (`.shape`, `.of`, …) still resolves (fixes an abort in the already-whitelisted `S12-introspection/attributes.t`).
- **`does Role(arg)` on the proxy is intercepted** in `eval_does_values`: the composed role state (`__mutsu_attr__*`) is recorded in a new registry map `class_attribute_container_mixins` keyed `(class, attr) -> (sigil, mixins)`.
- **At construction** `apply_attribute_container_mixins` applies the mixin across the MRO. For `@`/`%` the value *is* the container, so the mixin is applied directly. For `$` the value becomes a `__mutsu_scalar_container_mixin`-marked mixin whose `.VAR` returns itself, reproducing Raku's container-vs-value distinction (`\$o.attr.VAR.role`).

## Tests

- Regression pin: `t/attribute-container-role-mixin.t` (7 cases).
- Whitelisted `roast/S14-traits/attributes.t`.
- `make test` PASS; `cargo clippy -D warnings` and `cargo fmt` clean.

https://claude.ai/code/session_01UYzLgNeFvbQucTjEVzq4sW


tokuhirom:
Closing as duplicate: PR #4314 ("feat(traits): support Attribute.container.VAR does Role in custom trait_mod:<is>") landed on main independently and already implements the same feature — roast/S14-traits/attributes.t is 8/8 and whitelisted on current main (verified locally). No unique knowledge is lost; this branch is a parallel implementation of an already-solved problem.

PR #4333, opened 2026-07-08
fix(supply): box lexicals captured by tap/act/then closures as shared cells (S32 socket-recv-vs-read 13/13)

tokuhirom:
## Problem

`socket-recv-vs-read.t` test 11 (`read blocks until it has enough data`) failed: `$p.status` was `Kept` instead of the expected `Planned`.

Root cause is **cross-thread lexical sharing**, not socket semantics. A closure passed to `.tap` is stored and driven asynchronously on another thread (the socket-listener worker in `dispatch_socket_async_listen`), so it genuinely escapes the call frame — exactly like a `start {...}` block. But method-call arguments were compiled in a conservative NON-escaping position, so a lexical the tap closure captures-and-mutates was never promoted to a shared `ContainerRef` cell.

Consequence: the parent's post-registration reassignment `$send-rest = Promise.new` was invisible to the tap body, which read a stale clone-time snapshot (the previous, already-`Kept` promise). Its `await $send-rest` returned immediately, the worker finished early, and the status read `Kept`.

## Fix

Mark closure args to supply-consuming methods (`tap`/`act`) as escaping (`method_escapes_closure_args`), so the captured-and-mutated lexical is boxed into a shared cell — the same sound mechanism `start` uses (a cell always tracks later mutations; a by-value snapshot would only be correct if the var is never mutated). Over-boxing a user-defined `tap`/`act` whose closure does not escape is a correctness-preserving pessimization, not a bug.

## Verification

- `roast/S32-io/socket-recv-vs-read.t`: 13/13, stable across 8 repeated runs; added to whitelist.
- New local guard `t/tap-captured-lexical-cross-thread.t`.
- All whitelisted S17-supply tap/act tests still PASS.
- `make test` green; `cargo clippy`/`fmt` clean.


tokuhirom:
Closing in favor of #4336 (the `.then`-only subset), plus a follow-up needed on the Proc::Async tap path.

**Diagnosis preserved:** The escape-boxing of `.tap`/`.act`/`.then` closures fixes cross-thread lexical staleness (socket-recv-vs-read.t test 11: the tap reads `$send-rest` by value = a stale clone-time snapshot). It is architecturally correct (same rule `start {...}` uses), but boxing a `.tap` accumulator exposes a **pre-existing double-delivery** in the Proc::Async tap path: the live-channel async loop AND the await-time `replay_proc_taps` both fire the tap, and the duplicate write stayed invisible only while the capture was a by-value snapshot — a shared cell keeps both, doubling `S17-procasync/basic.t` test 37's output.

Also this branch was accidentally based on `docs-blockers-s17-syntax-done` rather than `main`.

**Landed instead:**
- #4336 — the safe `.then`-only escape-boxing (no proc-async interaction).
- (Flaky fix #4334 — de-flaked `t/react-nested-whenever-on-demand-close.t` — already merged.)

**Follow-up (tracked in TODO_roast/S32.md):** resolve the Proc::Async tap double-delivery (single delivery path per tap), then re-enable `tap`/`act` in `method_escapes_closure_args` to land socket-recv-vs-read.t.

PR #4469, opened 2026-07-12
feat(value): NaN-boxing representation flip — Value becomes one 8-byte word (3b-1 step B-flip)

tokuhirom:
## Summary

Layer 3b (NaN-boxing) final slice **B-flip** per [docs/nanbox-3b1-step-b-design.md](docs/nanbox-3b1-step-b-design.md) §4-4 / [ADR-0005](docs/adr/0005-nanbox-representation-encoding.md): `Value(ValueRepr)` → `Value(NanBox)`. **`size_of::<Value>()` drops 48 → 8 bytes** (`value_size_guard` updated; `Option<Value>` stays 8 via the NonZeroU64 niche). Every clone/move/drop of a `Value` now handles one word instead of a 48-byte enum.

- **`nanbox/peek.rs` (new)**: the borrowed decode. `view()` classifies the word and hands out guard-type smart pointers (`ManuallyDrop` reconstructions from the payload address) for direct-pointer kinds, and payload-pointee `&'a` references for multi-field `Arc<...Box>` kinds. Word-level probes back the wall accessors (`as_int` — inline + boxed, `as_str`, `as_sub_data`, `as_list_slice`, `is_hash_itemized`, …).
- **Seams (`mod.rs`)**: `into_repr` = tag decode (pointer payloads move, no refcount traffic), `from_repr` = tag pack, `with_repr_mut` = decode-edit-repack. `same_variant` compares collapsed variant tags (six Array kinds / itemized Hash / mutable Set-Bag-Mix / four Junction kinds / boxed Int fold onto their `ValueRepr` discriminant). `Value::NIL`/`TRUE`/`FALSE` are const inline words.
- **Wall accessors (`view.rs`)**: `with_*_mut` probe the view then decode-edit-repack; `with_array_inplace` / `detach_shared_container` ride `view()` guards; `into_array` / `into_proxy_parts` ride `into_repr`.
- **GC soundness fix caught by gc-stress**: multi-field payloads that were `Box`-owned pre-flip now share one `Arc<...Box>`, so tracing a shared box from every holder over-counted its inner `Gc` edges (trial-deletion strong-count underflow). `gc_trace` now gates on `value_bearing_arc_payload_uniquely_owned()` — a shared wrapper is treated as an external root holder, the same under-collect tradeoff as the existing `uniquely_owned` gates for Seq/Mixin/LazyThunk.
- `&Value::NIL` stopped const-promoting (`Value` now has `Drop`): the two sites keeping such a borrow across an expression use a `'static NIL_VALUE`.

## Gates (ADR-0005 §3.2)

- `make test` PASS (1677 files / 16507 tests); `cargo test` + gc-stress (16) green; clippy `-D warnings` + fmt clean.
- **GC counters invariant vs main**: `candidate_pushes`/`roots_scanned` byte-identical on fib (13/13) and a class+array+hash workload (4031/2028); scalar hot paths stay GC-free by construction (scalar pages carry no `Gc` payload).
- Roast spot checks (junction autothreading, hash, pair, capture, mixin-6e, proxy, binding/scalars, split, operators/misc): all PASS. Full roast delegated to CI.
- Release micro-benchmarks (int-arith / fib / bench-class clone-drop share) will be posted as a PR comment once the release build finishes.

https://claude.ai/code/session_01WkSj6GsxHMpcvDcg1kj6ni

### DIFF
--- a/src/runtime/test_functions/eval_exception.rs
+++ b/src/runtime/test_functions/eval_exception.rs
@@ -117,7 +117,10 @@ impl Interpreter {
             }
             positionals.push(arg);
         }
-        let value = positionals.first().copied().unwrap_or(&Value::NIL);
+        let value = positionals
+            .first()
+            .copied()
+            .unwrap_or(&crate::value::NIL_VALUE);
         let type_name = match positionals.get(1).copied() {
             Some(v) => match v.view() {
                 ValueView::Package(name) => name.resolve(),
--- a/src/value/guards.rs
+++ b/src/value/guards.rs
@@ -37,12 +37,27 @@ impl<'a, P> RefGuard<'a, P> {
     /// borrowed from the `Value`. Post-flip, `view()` instead reconstructs
     /// the pointer from the packed payload address.
     #[inline]
+    #[allow(dead_code)]
     pub(crate) fn borrowed(source: &'a P) -> Self {
         // SAFETY: the copy is wrapped in ManuallyDrop and never dropped, so
         // the source keeps its unique ownership of the reference; the
         // PhantomData borrow keeps `source` alive and unmutated for 'a.
         Self {
             inner: ManuallyDrop::new(unsafe { std::ptr::read(source) }),
+            _borrow: PhantomData,
+        }
+    }
+
+    /// Packed-storage constructor: wrap a smart pointer reconstructed from a
+    /// NaN-box payload address. The caller (the `nanbox` decoder) guarantees
+    /// the word it decoded from stays alive and unmutated for `'a`, so the
+    /// wrapped pointer's referent cannot be released while the guard lives;
+    /// `ManuallyDrop` keeps this reconstruction from double-releasing the
+    /// word's owned reference.
+    #[inline]
+    pub(in crate::value) fn from_reconstructed(source: P) -> Self {
+        Self {
+            inner: ManuallyDrop::new(source),
             _borrow: PhantomData,
         }
     }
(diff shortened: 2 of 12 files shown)


tokuhirom:
## Release micro-benchmarks (same-window A/B, `perf stat -r5`, `taskset -c 0-3`)

| bench | main (s) | flip (s) | delta |
|---|---|---|---|
| fib | 0.1464 | 0.1366 | **-6.7%** |
| int-arith | 0.0890 | 0.0822 | **-7.7%** |
| bench-class | 0.1904 | 0.1776 | **-6.7%** |
| method-call | 0.1863 | 0.1769 | **-5.0%** |
| bench-num (f64 loop) | 0.2728 | 0.2477 | **-9.2%** |
| bench-fib | 0.4309 | 0.3980 | **-7.6%** |

- **No regression anywhere; a uniform 5-9% win.** The ADR's aspirational "int-arith 2x / fib +30%" assumed the 48-byte enum was still the dominant cost at flip time — the July perf slices (for-loop O(n²) writeback fix, GetLocal clone removal, ctor plan cache, ?LINE field) already harvested most of that, so the flip lands as a broad clone/move/drop cheapening instead of a single spike.
- **Num-regression gate (ADR-0005 §2.1)**: the offset-encoded double costs ~1 ALU on pack/unpack; the f64-heavy loop is 9% *faster*, so the NaN-favored fallback stays closed.
- Full suite vs raku after the flip: fib **0.90x**, bench-class **0.89x**, bench-fib **1.88x**, method-call **1.13x**, int-arith **0.63x** — every PLAN §5 target still met.

PR #4478, opened 2026-07-12
perf(vm): JIT J4a — Tier B inline arithmetic (NaN-box tag-dispatched Int/Num fast paths, ADR-0004 J4)

tokuhirom:
## Summary

First slice of ADR-0004 §2.5 **J4 (Tier B)**: the hot arithmetic/comparison opcodes are now lowered **directly into CLIF** with NaN-box tag dispatch, instead of one helper call per opcode. The inline fast paths cover exactly the interpreter arms' own fast paths — small-Int⊕small-Int and Num⊕Num — and every other word shape (boxed Int, BigInt, Rat, Junction, user overloads, ...) falls back to the unchanged Tier A shim, which re-runs the full interpreter arm on the still-untouched stack. Small Int and Num words carry no payload ownership, so the inlined loop-body ops execute with **zero refcount and zero GC traffic** (the ADR-0004 J4 gate direction).

### What is inlined

| Opcode | Fast path | Slow path trigger |
|---|---|---|
| `Add`/`Sub`/`Mul` | Int⊕Int (48-bit range check; `smul_overflow` for Mul), Num⊕Num (canonical-NaN select) | boxed/big/mixed operands, result leaving the small-Int range, any user infix declared (`Add` only, per the interpreter arm) |
| `NumLt/Le/Gt/Ge/Eq/Ne` | Int ordered compare via shifted words / canonical-word equality; Num via ordered/unordered `FloatCC` matching Rust `f64` semantics | non-Int/Num operands |
| `LoadConst` | refcount-free constants (Int/Num/Bool/Nil/...) pushed as one raw immediate store | pointer-kind constants; stack grow |
| `SetSourceLine` | single inline store | — |
| `ContainerizePair` | non-Pair top = no-op (kind probe) | Pair top |
| `JumpIfFalse`/`JumpIfTrue` | Bool-word compare + native branch (+ backedge safepoint poll) | non-Bool tested value (incl. Failure marking) |

### Mechanism

- **`vm_jit_layout.rs`**: `Interpreter` field offsets via `offset_of!`, plus a runtime probe of `Vec<Value>`'s (ptr, len, cap) word order (unspecified by Rust), self-verified per process and pinned by a unit test. Probe failure degrades Tier B to the Tier A helper-call form — never wrong, only slower.
- **`nanbox::jit_words`**: the only bit-layout export outside `crate::value`; all constants derive from the layout definitions, so a layout change flows into the JIT automatically.
- **`USER_INFIX_DECLS`** (process-wide monotonic counter): registration sites bump it; the inline `Add` fast path proves "no user `infix:<+>` anywhere" with one load+test. Nonzero conservatively routes through the helper, which re-checks precisely.
- `vm_jit_support.rs` / `vm_jit_tier_b_flow.rs` are file-size splits (500-line rule), no behavior change.

### Verification

- `MUTSU_JIT=on MUTSU_JIT_THRESHOLD=2` × full `prove t/` (1677 files): green (only `t/io-socket-recv-limit.t` flaked under `-j4` port contention; passes 3/3 isolated, on and off).
- Arithmetic-heavy roast subset (S03-operators/arith.t, S02-types/num.t, S32-num/*) with JIT on: green. Full roast = CI jit-stress.
- 3-way output diff (JIT off / JIT on / real `raku`) byte-identical on fib, bench-fib, int-arith and the edge-case probe (48-bit boundaries, NaN, −0e0, Inf, mixed Int/Num, `multi sub infix:<+>` override).
- New pins: `t/jit-tier-b-arith.t` (21 assertions on the fast-path boundaries), `t/jit-tier-b-infix-override.t`.

### Measured (release, `perf stat -r5`, `taskset -c 0-3`, same quiet window, same binary)

| bench | JIT off | JIT on | Δ |
|---|---|---|---|
| fib | 134.2 ms | 112.5 ms | **−16.2%** |
| bench-fib | 371.3 ms | 315.0 ms | **−15.2%** |
| int-arith | 90.4 ms | 88.9 ms | ±noise (expected: top-level compound loop never enters the JIT — that is the J4b hot-loop-entry slice) |
| method-call | 177.3 ms | 181.4 ms | +2.3% (≈ J3 noise band) |
| bench-class | 193.1 ms | 192.0 ms | ±noise |

A main-worktree A/B (same window) will be added as a comment.

### Next (J4b)

Hot-loop entry for compound-loop bodies via a `run_range` hook (per-range call counter + sub-range compilation) — brings `int-arith`-class top-level loops into the JIT.

https://claude.ai/code/session_01NWpYJdKBYbT9fnimevhpQT

### DIFF
--- a/src/runtime/registration_sub.rs
+++ b/src/runtime/registration_sub.rs
@@ -494,6 +494,7 @@ impl Interpreter {
     ) -> Result<SubRegisterOutcome, RuntimeError> {
         if name.starts_with("infix:<") {
             self.user_declared_infix_ops.insert(name.to_string());
+            crate::vm::vm_jit::note_user_infix_decl();
         }
         let is_method_value_decl = custom_traits
             .iter()
@@ -1119,6 +1120,7 @@ impl Interpreter {
     ) -> Result<(), RuntimeError> {
         if name.starts_with("infix:<") {
             self.user_declared_infix_ops.insert(name.to_string());
+            crate::vm::vm_jit::note_user_infix_decl();
         }
         Self::validate_callable_param_return_redeclaration(param_defs)?;
         if let Some(spec) = return_type
--- a/src/runtime/runtime_module_exports.rs
+++ b/src/runtime/runtime_module_exports.rs
@@ -230,6 +230,7 @@ impl Interpreter {
             }
             if name.starts_with("infix:<") {
                 self.user_declared_infix_ops.insert(name.clone());
+                crate::vm::vm_jit::note_user_infix_decl();
             }
             let source_single = format!("{module}::{name}");
             let source_prefix = format!("{module}::{name}/");
--- a/src/value/mod.rs
+++ b/src/value/mod.rs
@@ -291,6 +291,8 @@ mod guards;
 /// NaN-boxed 8-byte representation core (3b-1 step B): the packed word that
 /// IS the `Value` storage. The only module that knows the bit layout.
 mod nanbox;
+#[cfg(feature = "jit")]
+pub(crate) use nanbox::jit_words;
 mod serde_support;
 pub(crate) mod signature;
 pub(crate) mod types;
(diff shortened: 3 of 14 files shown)


tokuhirom:
Main-worktree A/B (same quiet window, `perf stat -r5`, `taskset -c 0-3`, release):

| bench | main JIT on | J4a JIT on | Δ | main JIT off | J4a JIT off |
|---|---|---|---|---|---|
| fib | 123.3 ms | 111.1 ms | **−9.9%** | 131.3 ms | 137.5 ms (+4.7%, layout-noise band) |
| bench-fib | 345.7 ms | 319.3 ms | **−7.6%** | 378.1 ms | 387.8 ms (+2.6%, layout-noise band) |

The off-mode deltas are binary-layout noise (off-mode executes byte-identical logic; the only non-JIT-gated change is a counter bump at operator-registration time). Net Tier B contribution on JIT-on fib-class benches: **−8〜10%** — call dispatch (GetLocal/CallFunc/Return helpers) still dominates fib, which is what the J4b hot-loop-entry + later slices go after.

PR #4632, opened 2026-07-16
fix: cell-ify a call-arg-source lexical captured by a closure so it resolves lexically

tokuhirom:
## Summary

A free variable that is both **handed to a call as an argument** and **captured by a nested closure** could dynamically resolve to a same-named *parameter* of the callee that eventually invokes the closure, instead of its own lexical binding.

### Root cause

The compile-time vouch analysis (`authoritative_free_vars`) deliberately refuses to vouch for a call-arg-source local, because an `is rw` / `is raw` parameter might write it back — a by-value overwrite-install would then go stale (`own_call_arg_sources` is excluded from the `vouched` set). But refusing to vouch also means the capture is never overwrite-installed at closure entry, so a foreign calling frame with a same-named parameter shadows it (dynamic scoping leaking into a lexical read).

### Fix

Give such a local a shared `ContainerRef` cell (ADR-0001, cell-vs-by-value). The cell is overwrite-installed at closure entry (fixing the misresolution) **and** live-tracks any genuine rw writeback (sound in the other direction too), so it is correct regardless of whether the callee actually rw-writes.

`compute_free_vars` now adds an own local that is captured by a nested closure and present in `own_call_arg_sources` to both `captured_mutated` and `needs_cell`. Unlike the mutation case this is **not** gated on `escapes`: a closure passed as a call argument is never classified as escaping (the #2746 perf guard), yet that is exactly the closure that gets invoked from a foreign frame.

### Context

This was the remaining call-arg sub-case of the zef `Zef::Extract.extract` regression (follows #4627): `my $path := $candi.uri` reached `self!extractors($path)` and was captured by `.map(-> $ex { lock-file-protect(IO() $path, -> { ... $ex.extract($path) }) })`, mis-resolving to `lock-file-protect`'s own `$path` param (the `.lock` file). With this fix the extract phase no longer tries to untar the lock file (verified end-to-end against real `zef fetch Test::META`).

## Test

`t/closure-callarg-source-cell.t` — 6 cases: the core misresolution, three-level nesting (map/block/start, the zef shape), rw-writeback live-tracking, the grep inline path, `=:=` identity, and no over-capture of a same-named foreign binding. `make test` passes (17295 tests); the previously-sensitive `roast/S17-lowlevel/lock.t` and `t/concurrent-compound-assign.t` still pass.

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -3034,6 +3034,25 @@ impl CompiledCode {
                         needs_cell.insert(*sym);
                     }
                 }
+                // A call-arg-source own local captured by a nested closure can
+                // never be vouched (an `is rw` / `is raw` param might write it
+                // back — see `own_call_arg_sources` and the `vouched` filter
+                // below), so it can never be overwrite-installed BY VALUE at the
+                // closure entry. Give it a shared `ContainerRef` cell instead:
+                // the cell is overwrite-installed at closure entry — fixing the
+                // dynamic misresolution where the free var resolves to a
+                // same-named parameter of the callee that eventually invokes the
+                // closure (`callee($path, -> { ... $path ... })`) — AND it
+                // live-tracks any genuine rw-writeback, so it is sound in both
+                // directions (ADR-0001 cell-vs-by-value). Deliberately NOT gated
+                // on `escapes`: a closure passed as a call argument is never
+                // classified as escaping (the #2746 perf guard), yet it is
+                // exactly such a closure that gets invoked from a foreign frame
+                // whose same-named param triggers the misresolution.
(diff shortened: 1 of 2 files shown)


tokuhirom:
Closing: the cell-ification approach over-boxes scalars that hold Set/Bag/Mix/Pair and `is default` trait'd containers (regressing bag/mix/pair/is_default roast tests), and it does not even fix the target case — an `Instance`-valued $path (IO::Path) is skipped by box_captured_lexicals's reference-type guard, so the zef misresolution persists. The sound fix needs a narrower, value-type-aware design. Reopening with a correct approach.

PR #4654, opened 2026-07-17
WIP: recursion through a start block returns a silently wrong answer (investigation, do not merge)

tokuhirom:
## The bug

```raku
sub f($n) { start { $n <= 0 ?? "b" !! await(f($n - 1)) ~ "|$n" } }
say await f(3);   # b|1|2|3   correct
say await f(3);   # b|3|3|3   WRONG, no error   (raku: b|1|2|3)

sub k($n) { $n <= 0 ?? "b" !! (await start { k($n - 1) }) ~ "|$n" }
say k(3);         # b|1|1|1   WRONG ON THE FIRST CALL
```

Routing through a `my` was worse (`b|0|0|0`), and a two-branch `fib` with `await` inside `start` hung
outright. PLAN §6 recorded only the hang; the silent corruption is the bigger half.

## Root cause

`clone_for_thread` seeded every lexical into `shared_vars`, a flat map keyed by the **bare name**. It
cannot represent two concurrently-live bindings of one name — exactly what a recursive frame chain
is. Each frame's `$n` bind overwrote `shared_vars["n"]` (`flush_local_to_env` → `set_shared_var_sym`),
and `await`'s `sync_shared_vars_to_env` queued it into `apply_pending_caller_var_writeback`, which is
retain-on-miss and deliberately walks **up** the frame chain until some frame owns a slot named
`"n"` — force-feeding the innermost value into every ancestor.

`start` already compiles its block as escaping, so `box_captured_lexicals` had **already** given each
frame's `$n` a correct per-binding home. The flat map was a second, lossy mechanism running in
parallel with a working one and overwriting its answer — a "1 operation = 1 implementation" violation.

## The fix

- A thread block's own captured scalars are no longer seeded by name (`clone_for_thread_for_block`).
  `@`/`%` keep the lane — their `__mutsu_atomic_*` CAS copies are keyed off those entries.
- **`thread_redeclared_vars` is deleted.** The scalar-only `my` mask existed solely to paper over the
  same flat map. This also **subsumes #4650**, which gated the Nil-refresh read in `exec_get_local_op`
  on that mask so `my $x := f()` yielding Nil would not resurrect an unrelated scope's `$x` from the
  process-global store: not seeding a block's captures removes the collision at its source instead of
  gating one reader. #4650's pin (`t/shared-var-nil-redeclared-mask.t`) passes with both the mask and
  its gate gone.

Supporting changes, each forced out by a pin rather than added speculatively:

- **Typed scalars are boxed when the closure goes to a THREAD** (new `CompiledCode::thread_escaping`).
  `my Int $c = 0; await start { $c = 5 }` previously rode the retired lane. Boxing does not weaken the
  constraint — the check runs at the assignment op, looked up **by name** in `var_type_constraints`
  (cloned into the child), *before* any write-through; the loop path has always boxed constrained
  scalars. It is deliberately not done for a same-frame closure, because `cas` resolves its target by
  name — `roast/S17-lowlevel/cas.t` caught exactly this.
- **`cas` swaps through a boxed scalar's `ContainerRef`** (`scalar_cell_target`) instead of writing a
  plain value over the cell in `env` and stranding the owning slot.
- **`.start`** joins `then`/`tap`/`act` in `method_escapes_closure_args`.
- **Lexicals a registered class/role method writes keep the name lane**
  (`CompiledCode::type_body_written_lexicals`). Such a write is invisible to the capture analysis — the
  method is installed by `RegisterClass`/`RegisterRole` with no closure-creation op — and it is what
  carries a `submethod DESTROY { $a++ }` write on a worker back to the parent.

## Verification

- New pins: `t/recursive-start-await.t` (8 tests, promoted from `docs/probes/pool-recursive-start.raku`,
  which fails on main) and `t/thread-shared-scalar-visibility.t` (9 tests over the shapes
  `box_captured_lexicals` declines to box; its 2 `todo`s are documented gaps, one pre-existing).
- All 9 lines of the repro now match raku, including the `fib` that used to hang.
- `make test` green (1786 files / 17467 tests). 389 S04/S06/S12/S14/S17 roast files green locally,
  including `S12-construction/roles-6e.t` and `S17-lowlevel/cas.t`, which this change transiently broke
  and which drove the final design. Full roast delegated to CI.
- Also removes a stray `|||||||` conflict marker that had been committed into `news/2026-07.md`.

## Left behind (recorded in PLAN §6, not silently dropped)

Two latent defects surfaced on the way, both pre-existing and both in the lexical-scope slot campaign's
territory:

- **The boxing skips are wrong for *rebinding*.** `Instance`/`Array`/… holders are reference-shared for
  in-place mutation but not for `$obj = Foo.new`. Boxing `Instance` was tried and reverted: it breaks
  `my Lock $l .= new` when two sibling blocks declare the same name, because `resolve_capture_slot`'s
  `rposition` search resolves the capture to the LAST same-named slot.
- **A parent-bound `Proxy` never receives a worker's STORE** (raku does). Unrelated to this change.


tokuhirom:
Superseded by #5564, which re-implements this PR's design on top of ADR-0010's lineage-scoped shared store. Both regressions this branch left are resolved there: the advent2013-day14.t failure turned out to be an artifact of the pre-ADR-0010 process-global flat map (it does not reproduce on the lineage store), and the nonblocking-await.t coherence hole is closed by retaining the thread_redeclared_vars mask for names excluded from seeding (the blanket mask-clear rested on the force-seed that no longer happens). The investigation doc and both test files from this branch are carried over in #5564; docs/recursive-start-shared-vars.md there records the final design.

PR #4658, opened 2026-07-17
fix: scope the cross-thread lexical store to a spawn lineage (ADR-0010)

tokuhirom:
Implements [ADR-0010](docs/adr/0010-cross-thread-lexical-sharing-scope.md).
Removes the mzef fetch blocker root-caused in #4652; follows #4650 / #4653 / #4655.

## The defect

`shared_vars` was one `Arc<RwLock<HashMap<String, Value>>>` handed to every
`clone_for_thread`, keyed by **bare name**. So there was exactly one `uri` key,
one `depends` key, for the whole process — two unrelated lexicals sharing a name
were the same entry.

zef's `Zef::Client.!fetch` hypers over candidates; each worker declares `my $uri`
and runs a nested `start`, and *that* inner spawn migrates the worker's lexicals
into the one global map a second time. N workers raced on the one key:

```
[Test::Async]     Fetching https://360.zef.pm/J/SO/JSON_OPTIN/… with plugin: …wget
[JSON::Unmarshal] Fetching https://360.zef.pm/J/SO/JSON_OPTIN/… with plugin: …wget
[META6]           Fetching https://360.zef.pm/J/SO/JSON_OPTIN/… with plugin: …wget
```

`thread_redeclared_vars` could never fix this: each batch thread has its own
`Interpreter` and its own mask, but they all shared the one map.

## The decision

Each spawned thread gets a store **chained to its parent's** (`SharedStore`):
read resolves own-then-ancestors; write resolves to the nearest ancestor holding
the name (so a child's write to the parent's `$counter` still reaches the
parent); a declaration binds into the current lineage. Siblings have no path to
each other — the fix — while a `start` nested inside a worker chains to it and
still sees its `$uri`.

Alternatives (more masking; cells-instead-of-names; copy-in/copy-out at join) are
weighed in the ADR. Cells are the *correct* end state and explicitly not rejected
— just gated on a capture analysis mutsu doesn't have yet (ADR-0001).

## What landing it taught us

`__mutsu_*` keys are **not lexicals**. The atomic element stores and the `state`
cells are explicitly process-wide: all four threads of `my atomicint $x; …start {
$x⚛++ }` must hit one counter. Lineage-scoping them silently loses updates —
exactly how `t/concurrent-state-var.t`, `t/module-state-sub-shared-cell.t` and
`t/state-aggregate-shared-cell.t` failed on the first attempt. They now resolve
at the root lineage; only user lexical names are scoped.

## The mask is mostly gone — honestly, not entirely

The ADR said removing `thread_redeclared_vars` is the proof the decision landed.
Two of its four uses are gone, and the decision is what removed them: the
`set_shared_var_sym` **write** gate and the `GetLocal` **read** gate — the
asymmetric pair that *was* #4650 — plus the sync dirty-key filter.
`t/shared-var-nil-redeclared-mask.t` passes with both gates removed.

One use remains: `clone_for_thread` still consults it for `declare` vs
`seed_if_absent`. That is **seed freshness** (a parent that re-declared `my $x`
since its last spawn would otherwise hand the child a stale value), a different
concern from a write leaking to the wrong lineage. Retiring it needs declaration
to push into the store eagerly; out of scope. Recorded in the ADR rather than
claimed as clean.

## Result

`zef install Test::META` **gets past fetch** — all 16 candidates now download
their own archive, every `Fetching [FAIL]` gone. It stops at an unrelated
`Zef::Extract` matcher blocker (new frontier, documented in
`docs/mzef-install-pipeline.md`).

## Testing

- `t/shared-store-lineage-scope.t` — new pin: sibling hyper workers, the
  callee-invoked closure variant, child→parent writeback, a `start` nested inside
  a spawned thread, atomics staying process-wide. **Passes under real `raku`.**
- The 149 thread/shared/atomic/promise/lock/supply/state `t/` files — the ADR
  names these the specification for what must not break. All pass.
- `make test`: 1785 files / 17461 tests, all pass.
- Full roast + gc-stress + jit-stress via CI.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -183,13 +183,22 @@ Current state (details in news/2026-06.md and news/2026-07.md): ✅ CLI load + c
 - [ ] Known small difference: coercion of CLI numeric strings to `Int $n` is more eager than raku
       (`7` matches `MAIN(Int $n,…)`; raku falls back to slurpy). In practice the mutsu behavior is
       more intuitive.
-- [ ] **network fetch**: fetching from the fez ecosystem (`https://360.zef.pm/`). Robust async TLS is
-      a prerequisite.
-- [ ] **Real install + build/test execution**, an `mzef` binary shim + vendoring of zef itself +
-      dependencies + config (debian's zef lacks `resources/bin/zef`; a known-good vendoring is needed).
+- [x] **network fetch** — done. **No native TLS was needed**: zef shells out to the system
+      `curl`/`wget` and mutsu drives that via `Proc::Async`, so the old "robust async TLS is the
+      biggest prerequisite" assumption was simply wrong. Concurrent multi-candidate fetch works as of
+      #4658 (ADR-0010).
+- [x] **Real install** — done for a dependency-free dist: a real fez dist downloads, extracts,
+      installs into the site repo and is then `use`-able (#4655). Verify with a **non-bundled** dist
+      (`JSON::OptIn`); `use JSON::Fast` succeeds without any install because mutsu bundles it.
+- [ ] **A dist WITH dependencies** — the remaining gap. Resolution (#4650) and fetch (#4658) are done;
+      `zef install Test::META` now dies in **extract**. See `docs/mzef-install-pipeline.md`
+      ("Current frontier") for the lead and the next steps.
+- [ ] **build/test execution**, an `mzef` binary shim + vendoring of zef itself + dependencies +
+      config (debian's zef lacks `resources/bin/zef`; a known-good vendoring is needed). The test
+      phase (6) has not been exercised yet — runs have used `--/test`.
 
-Split: **"keep running the real Zef as a test target" has high immediate value — continue it**. For
-real installs as the bundled installer, network fetch (TLS) is the biggest prerequisite work.
+**`docs/mzef-install-pipeline.md` is the live tracker** — phase table, what each fix unblocked, and
+the current frontier with its next steps. Read it before picking up mzef work.
 
 ### B3. Distribution and tooling
 
--- a/docs/adr/0010-cross-thread-lexical-sharing-scope.md
+++ b/docs/adr/0010-cross-thread-lexical-sharing-scope.md
@@ -72,19 +72,33 @@ struct SharedStore {
   write there, so a child's mutation of the parent's `$counter` reaches the
   parent (the property `t/cross-thread-shared-var-writeback-coherence.t` and
   friends pin). If no ancestor has it, the name is this lineage's own.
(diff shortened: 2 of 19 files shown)


tokuhirom:
## ⚠️ Red — real regression, converted to draft

`roast/integration/advent2013-day14.t` (whitelisted) fails on this branch and is
**not flaky**: all three of `test`/`gc-stress`/`jit-stress` fail on it, and it
reproduces deterministically in a local debug build.

```
$ MUTSU_FUDGE=1 ./target/debug/mutsu roast/integration/advent2013-day14.t
ok 1..6
Runtime error: Test failures
# You planned 10 test, but ran 6
```

Verified against `main` properly (by checking `main` out — an earlier `git stash`
check was meaningless here, since the changes are *committed* on the branch):
**main passes 10/10**, this branch aborts after test 6. So it is this PR's
regression, not pre-existing.

### What is narrowed down

- The abort is in the file's **Channels** section (test 7+). Extracted standalone
  it runs 0 of 4 — so it dies before the first `is`, not on an assertion.
- The failure is **silent**: no exception surfaces even with the
  `read_worker`/`parse_worker` `CATCH` handlers removed.
- **Not** the read gate: restoring `thread_redeclared_vars` on the `GetLocal`
  read path does not fix it.
- **Not** the obvious same-name hypothesis: `read_worker($dest)` and
  `parse_worker($dest)` both take a parameter named `$dest` and both spawn a
  `start` capturing it, so both seed `dest` into the same (root) lineage and the
  second `seed_if_absent` skips. But an isolated repro of exactly that shape
  passes and matches raku — so that is not it, or not it alone.

Repro kept at `tmp/adr0010-regression-repro.raku` (the extracted Channels
section, gitignored).

### Where to look next

The section builds a Channel pipeline: `read_worker(@files, $read)` →
`parse_worker($read, $parsed)` → `await config_combiner($parsed)`, each a `start`
block, plus a grammar parse. Prime suspects, in order:

1. **`Channel` / `Supply` internals keyed into the store.** ADR-0010 routes
   `__mutsu_*` keys to the root lineage and user names to the lineage — anything
   channel-related keyed by a *user* name (or by an internal name that is not
   `__mutsu_`-prefixed) would now resolve per-lineage where it used to be global.
   Grep the channel/supply paths for store keys that are neither.
2. **`config_combiner`'s `await`** — a Promise/Channel handed between two
   sibling `start` lineages.
3. The `is_internal_key` prefix rule itself: is every process-wide key actually
   `__mutsu_`-prefixed? That assumption is what the three `state` tests already
   caught once.

The lesson this cost: **`make test` (t/ only) passing is not evidence for a
concurrency-store change** — a roast subset (`roast/integration/*`, `S17-*`) had
to be run locally before pushing, exactly as `CLAUDE.md` warns.

PR #4755, opened 2026-07-18
fix: sequence operator (...) is looser than comma in argument lists

tokuhirom:
## Summary

The sequence operators `...`/`…`/`...^`/`…^` have **"list infix" precedence, which is looser than comma**. So in an argument list `a, b ... limit`, the whole comma list `a, b` is the sequence's seed and the entire list collapses into **one** sequence argument — exactly like `(a, b ... limit)` in parentheses:

```raku
sub g($a, $b) { }
g(1, 3 ... 11);          # ONE argument (a Seq) → "too few positionals", not two args
@a.prepend: 1, 3 ... 11; # prepends 1 3 5 7 9 11 (seed is 1,3), not 1 3
```

mutsu split the comma level first and parsed each argument individually, so `f(1, 3 ... 11)` became `f(1, (3 ... 11))` (seed `3`, wrong step) and `@a.prepend: 1, 3 ... 11` dropped the endpoint entirely. The parenthesized-list parser already handled this via `try_parse_sequence_in_paren`; the call/listop paths did not.

## Fix

Add `try_parse_sequence_arg_list` (a terminator-agnostic sibling of the paren helper: it collects the comma seeds, then absorbs the sequence operator, its endpoint, and any extra endpoints), and wire it into **every** argument-list parser that lacked it:

- parenthesized calls — `parse_call_arg_list`
- no-paren user subs & builtin listops (`say`, `join`, …) — `identifier_call`
- expression listops & the no-paren invocant path — `listop`
- positional call args — `args`
- IO statements `say`/`print`/`put`/`note` — `io_stmts`
- the method colon-arg form `.m: a, b ... limit` — `postfix/loop_`

Ordinary comma lists (no top-level sequence operator) return `None` from the helper and are completely untouched.

Verified against reference `raku` across paren/no-paren/method/colon/chained-waypoint/exclusive-endpoint forms, and that plain comma lists, ranges, and string arg lists are unaffected. `make test` passes.

## Provenance

Found by the QA doc-diff harness (PLAN.md §8.1) as its **first root-cause cluster** — several Array/Int doc examples (`prepend`/`unshift`/`polymod` with a `a, b ... limit` argument) diverged from raku for this one reason.

Pin: `t/sequence-op-in-call-args.t`.

Note: `600.polymod(lazy gather {…})` is a *separate* bug (polymod not expanding a lazy Seq argument — a method-impl issue, not this parser fix) and is left for a follow-up.

### DIFF
--- a/src/parser/expr/postfix/loop_.rs
+++ b/src/parser/expr/postfix/loop_.rs
@@ -936,6 +936,29 @@ fn postfix_expr_loop(mut rest: &str, mut expr: Expr, allow_ws_dot: bool) -> PRes
                     r.starts_with(' ') || r.starts_with('\t') || r.starts_with('\n');
                 let (r2, _) = ws(r)?;
                 if r2.starts_with(':') && !r2.starts_with("::") {
+                    // A sequence operator (`...`/`…`) is looser than comma, so a
+                    // colon-arg list `.m: a, b ... limit` is ONE sequence argument
+                    // (seed `a, b`). Absorb the whole comma level, matching the
+                    // parenthesized-list parser. (Only for the no-space colon-arg
+                    // form; a space before `:` is an adverb colon-pair.)
+                    if !has_space_before_colon {
+                        let after_colon = &r2[1..];
+                        let (after_colon, _) = ws(after_colon)?;
+                        if let Some(result) =
+                            crate::parser::primary::try_parse_sequence_arg_list(after_colon)
+                        {
+                            let (r_seq, seq) = result?;
+                            expr = Expr::MethodCall {
+                                target: Box::new(expr),
+                                name,
+                                args: vec![seq],
+                                modifier,
+                                quoted: false,
+                            };
+                            rest = r_seq;
+                            continue;
+                        }
+                    }
                     // If there was space before ':', treat as adverb (colon-pair)
                     let (r3, first_arg) = if has_space_before_colon {
                         colonpair_expr(r2)?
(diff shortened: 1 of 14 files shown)


tokuhirom:
## CI regression diagnosis (blocked on a separate lazy-flat gap)

CI's full `make roast` surfaced one whitelisted-test regression: `roast/S02-types/lazy-lists.t` test 27 — `flat [2,3,4], 10,11 ... *`.

This parse change is **correct** (verified against raku): the sequence operator is looser than comma, so `flat [2,3,4], 10,11 ... *` is `flat( ([2,3,4], 10, 11) ... * )` — one infinite-Seq argument whose first element is the plain array `[2,3,4]`. raku's `flat` then lazily flattens it → `(2 3 4 10 11)`.

The regression is **orthogonal to this parse fix**: mutsu's `flat` does not deep-flatten an **infinite** lazy list's nested arrays.
- `flat [2,3,4], 10,11 ... 15` (finite endpoint) → correct `[2,3,4,10,11,12,13,14,15]` (the Seq is materialized, `flat_val` flattens it).
- `flat [2,3,4], 10,11 ... *` (infinite) → mutsu keeps `[2,3,4]` nested: `([2 3 4] 10 11 12 13)`. `flat_val` (`src/builtins/functions/flat.rs`) pushes a genuinely-lazy list through opaquely (line ~104) instead of returning a lazily-flattened list.

The old (incorrect) parse masked this by making `[2,3,4]` a *separate* `flat` argument, which the eager multi-arg path flattened directly.

**Prerequisite:** implement lazy `flat` (a flattening lazy pipe over the input list) so `flat` of an infinite lazy list flattens each element as it is pulled. That is a real feature, not a parse concern — this PR is parked until it lands (or the scope is reconsidered).

PR #4822, opened 2026-07-19
fix: Array.List materializes autovivification holes as Nil

tokuhirom:
## Problem

Surfaced by the QA doc-diff harness (PLAN.md §8) against `Type/Array.rakudoc`. In raku, `.List` turns an array's autovivification gaps into `Nil`:

```raku
my @a = [1]; @a[3] = 3;
say @a.List;   # (1 Nil Nil 3)   — holes are Nil
say @a;        # [1 (Any) (Any) 3] — the gist still shows (Any)
```

mutsu's `.List` copied the raw backing vector, so holes leaked through as `(Any)`, diverging from raku.

## Fix

In the `.List` Array arm, replace hole slots with `Nil`. A hole is a `Package(_)` element whose index is **not** in the `initialized` set — the same invariant `:exists`/`:k` already use. Bulk/literal arrays have `initialized == None` (no holes) and are untouched; an explicitly-assigned type object (`@a[0] = Int`) is in the set and correctly stays. Typed arrays (`my Int @a`) whose holes are `(Int)` are also converted, matching raku. The Array gist path is unchanged.

## Tests

New pin `t/array-list-holes-nil.t` (8 cases: untyped/typed holes → Nil, gist unchanged, leading holes, bulk/literal untouched, assigned-type-object stays, contiguous no-holes). Full local `make test` green; whitelisted `roast/S02-types/array.t`, `roast/S09-typed-arrays/arrays.t`, `roast/S32-array/create.t` still pass.

### DIFF
--- a/src/builtins/methods_0arg/coercion.rs
+++ b/src/builtins/methods_0arg/coercion.rs
@@ -335,7 +335,30 @@ pub(super) fn dispatch(target: &Value, method: &str) -> Option<Result<Value, Run
             // A shaped array falls through to the slow path, which flattens all
             // dimensions and replaces Nil slots with the type-default.
             ValueView::Array(..) if crate::runtime::utils::is_shaped_array(target) => None,
-            ValueView::Array(items, _) => Some(Ok(Value::array(items.to_vec()))),
+            // `.List` materializes autovivification holes as `Nil` (raku:
+            // `my @a=[1]; @a[3]=3; @a.List` is `(1 Nil Nil 3)`), unlike the
+            // Array gist which shows the `(Any)` type object. A hole is a
+            // `Package("Any")` slot whose index is not in the `initialized`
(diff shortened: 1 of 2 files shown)


tokuhirom:
Closing: the fix breaks whitelisted `roast/S32-array/delete.t` test 33 (`.List` `=:= Nil` check). Root cause is deeper than `.List`: mutsu has pre-existing Nil-identity bugs — `Nil.^name` returns `Any` (raku: `Nil`), and a subscripted literal-`Nil` list element gives `=:= Nil` False (bare `Nil =:= Nil` is True). main satisfies `delete.t` via Any-hole metadata that keys `=:= Nil`; materializing literal Nil for correct `.raku`/gist rendering loses that. Deferred until Nil type-identity is fixed.

PR #4933, opened 2026-07-20
perf: (B)-gate flip — MUTSU_GATE_LOCAL_ENV_WRITE default ON (Plan A)

tokuhirom:
Flip `gate_local_env_write()` to default **true** now that both the `t/` and the full roast ON surveys are clean (the last two ON-only regressions, `S32-trig/e.t` and `S32-list/skip.t`, were burned down in #4930).

## What the gate does
Under the gate, a slot-authoritative plain-lexical store (not captured / reflective / needs-sync) skips mirroring into the name-keyed `env`, so the slot is the single source of truth and env-COW can drop the write. Every gate-guarded fold was validated **ON == OFF** across the whole test corpus during the months-long burndown, so this is a **perf change, not a semantics change**.

`MUTSU_GATE_LOCAL_ENV_WRITE=0` (or `off`) opts back out to the byte-identical env-mirror-every-store path.

## The one-line flip
```rust
// before: Some("1") | Some("on") | Some("ON")  (default OFF)
// after:  !matches!(…, Some("0") | Some("off") | Some("OFF"))  (default ON)
```

## Perf
while/loop hot paths drop the unconditional per-store env write (~−10…16% on a 5M-iteration while loop); for/fib are unaffected (their frames capture env by name, so every local is force-synced regardless).

## Validation
- `make test` (2041 files / 19338 tests): green with the flip as default.
- Full roast ON survey (release, `-j6`, 1433 files): clean — zero genuine regressions (already confirmed in #4930).
- **This PR's CI (roast + gc-stress + jit-stress) is the concurrency/GC soak** that the local `-j6` survey cannot exercise.

Gate-site removal (making the ~18 `if gate_local_env_write()` sites unconditional and deleting the OFF-side dead code) follows as a **separate mechanical PR once this has soaked**.

### DIFF
--- a/docs/lexical-scope-slot-campaign.md
+++ b/docs/lexical-scope-slot-campaign.md
@@ -846,10 +846,24 @@ regressions were burned down:
 
 Pin for both: `t/gate-b-callee-name-collision-and-deref-capture.t`.
 
-**Result: the (B)-gate ON survey (t/ and full roast) is green. The gate is ready
-for the default flip (Plan A: `gate_local_env_write()` default true,
-`MUTSU_GATE_LOCAL_ENV_WRITE=0` opt-out), then a gc-stress/jit-stress/roast CI soak,
-then removal of the ~18 gate sites and the OFF-side dead code.**
+**Result: the (B)-gate ON survey (t/ and full roast) is green.**
+
+### Plan A flip — `gate_local_env_write()` default ON (2026-07-20)
+
+With both surveys clean, `gate_local_env_write()` was flipped to default **true**
+(one-line change: `!matches!(…, "0"|"off"|"OFF")`), with `MUTSU_GATE_LOCAL_ENV_WRITE=0`
+as the byte-identical opt-out. Local pre-check: `make test` (2041 files / 19338
+tests) green with the flip; the roast ON survey was already clean. The flip's CI run
+(full roast + **gc-stress + jit-stress**) is the concurrency/GC soak that the local
+`-j6` survey cannot exercise. Perf: while/loop hot paths drop the unconditional
+env write (memory recorded −10…16% on a 5M while loop); for/fib are unaffected
+(their frames `captures_env_by_name`, so every local is force-synced regardless).
+
+**Remaining: after the flip soaks in main, delete the gate** — the ~18
+`if gate_local_env_write()` sites become unconditional (the ON branch), and the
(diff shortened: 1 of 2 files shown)


tokuhirom:
## Blocked — deferring the flip

The CI soak did its job and surfaced a genuine, deterministic blocker under **jit-stress** (`MUTSU_JIT=on MUTSU_JIT_THRESHOLD=2`). Closing this PR; the flip is a one-line change and will be re-opened once the blocker below is fixed.

### CI result triage
- **jit-stress: FAIL — genuine, deterministic (the real blocker).** Reproduced locally.
- **gc-stress: FAIL — spurious.** Only `S17-supply/return-in-tap.t` (a documented load-flaky, unrelated to the flip; it also flakes on the interpreter path).
- **test: FAIL — spurious.** Same `return-in-tap.t` flake.
- wasm-e2e: pass.

So the flip is **not** broadly broken — it is blocked by a single mechanism.

### Root cause (mechanism #3: method-call caller-local coherence × JIT)
Minimal repro (fails only with gate ON **and** JIT hot):
```raku
sub w($io is copy) {
    my @seen;
    until @seen.elems >= 3 { @seen.push: $io.Str; $io .= succ; }
    @seen;
}
say w(1);   # flip+JIT: [1 2 2]   (expected [1 2 3])
```
`apply_pending_rw_writeback` (`src/vm/vm_env_helpers.rs:1106-1111`) unconditionally pulls `env[source] → self.locals[slot]` after a method call. Under the flip, a plain scalar's store skips the env mirror, so `env["io"]` stays at its decl-seed `Any` while the JIT-fresh slot holds the live `Int`. The post-call pull then **clobbers the live slot with the stale env value**, freezing `$io` once the loop is JIT-compiled (verified by diagnostic: `rw-pull source="io" env=Package(Any) slot_was=Int(2)`).

### Fix direction
The pull needs the **`env_changed` guard** of #4859 / #4873: pull only when `env[source]` actually changed during the call (a genuine `is rw` writeback writes it; a non-mutating `.Str`/`.succ` invocant read does not). This touches a load-bearing rw mechanism with ~15 source-push sites and must soak under gc-stress/jit-stress, so it is a dedicated follow-up — not rushed onto the flip.

The burndown that made the roast ON survey clean (#4930) is already merged and stands on its own.

PR #4984, opened 2026-07-20
fix: subscript a numbered match variable in string interpolation ($0[0])

tokuhirom:
## Summary

`"$0[0]"` is `$/[0][0]` — the first positional subcapture of capture 0 — and
`$0<k>` / `$0{k}` name a subcapture. mutsu interpolated `$0` (the whole `Match`)
but left the trailing `[0]` as a **literal string fragment**, so

```raku
if 'abc' ~~ / ( a (.) (.) ) / { say "Inner: $0[0] and $0[1]" }
# raku:  Inner: b and c
# mutsu: Inner: abc[0] and abc[1]
```

The numbered-capture interpolation branch built the `$/[n]` index but only ran
`try_parse_interp_method_call`; it skipped `parse_postcircumfix_index`, which the
scalar `$var` branch already applies for a trailing `[...]`/`<...>`/`{...}`. Add
it so the subscript is consumed and lowered onto the capture. Bare `$0` and
`$0.method()` chains are unaffected; `$0[0].uc()` (subscript then method) works.

## Discovery

doc-diff sweep on `Language/regexes.rakudoc` (finding [17]) — see
`docs/doc-diff-backlog.md`.

## Test

Pin: `t/interp-capture-subscript.t`.


tokuhirom:
Superseded by #5010 (`fix: subscript a numbered capture ($0[0]) in string interpolation`, merged as da134701f), which implements the identical fix — the rebase conflict resolved both sides to the same `parse_postcircumfix_index` call. Verified on main: `if "abc" ~~ /(.)(.)/ { say "$0[0] $1" }` matches raku's output. Closing as redundant; the feature and its regression tests already live in main.

PR #5084, opened 2026-07-21
fix: parse the `:ver:<...>` use-selector form (colon before the angle value)

tokuhirom:
`use Foo:ver:<0.0.5>:auth<zef:lizmat>` — where the version/auth/api selector is
written with a colon before its angle value (`:ver:<...>` rather than the plain
`:ver<...>`) — failed to parse: the `use` adverb loop only recognized the plain
`:key<value>` form, so `ver` was taken as an import tag and the trailing
`:<0.0.5>...` was mis-read as an expression, yielding
`X::Undeclared::Symbols: Undeclared routine: Foo:ver`.

Accept an optional `:` between the selector key and its `<...>` value and
normalize both forms to the canonical `:key<value>`. Raku parses the colon form
in use-name position; several lizmat dists use it.

Unblocks Acme::Cow / Acme::Cow::Example (ticket T-007), which do
`use Acme::Cow:ver:<0.0.5>:auth<zef:lizmat>`; both modules load now.

Pin: `t/use-colon-ver-selector.t`.


tokuhirom:
Duplicate of #5081 which already landed this exact fix. Closing.

PR #5177, opened 2026-07-22
docs(dist-tickets): pre-triage T-028 (ternary precedence); T-034 stale

tokuhirom:
Next-session prep.

- **T-028 (P5reset)** — pre-triaged with a confirmed minimal repro:
  ```raku
  my $l; False ?? 9 !! $l ~= "x"; say $l   # raku: x  /  mutsu: "Cannot modify an immutable value"
  ```
  With explicit parens `!! ($l ~= "x")` mutsu works, so the ternary `!!`
  false-branch stops at the bare term and drops the trailing `~=`, applying it to
  the immutable ternary result. Raku parses each ternary branch at
  item-assignment precedence. Records the repro, root cause, and target file.

- **T-034 (IdClass)** — verified passing on current main (`t/01`..`t/06` all
  green under both raku and mutsu). The sweep TSV predates intervening
  role/dispatch fixes; annotated as stale so it is not chased.

Docs-only.

### DIFF
--- a/TODO_dist/TICKETS.md
+++ b/TODO_dist/TICKETS.md
@@ -105,11 +105,29 @@ editing this file; keep edits small (one ticket) to avoid conflicts.
 - file: DONE — runtime/registration_role.rs, runtime/undeclared_routines.rs;
   remaining — nqp op support (large, separate)
 
-### T-028 — test_die [P5reset]  [impact: 1 dist]
+### T-028 — test_die [P5reset] (pre-triaged: ternary false-branch precedence)  [impact: 1 dist]
 - dists: P5reset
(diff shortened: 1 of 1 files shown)


tokuhirom:
Superseded by #5180 (T-028 fixed and moved to Done). Note: this PR's pre-triage diagnosis was incorrect — verified via `raku --target=ast` that `False ?? 9 !! $l ~= "x"` parses as `METAOP_ASSIGN(~, (ternary), "x")` = `(cond ?? 9 !! $l) ~= "x"` (the assignment wraps the whole ternary as an lvalue), NOT `!! ($l ~= "x")`. The ternary is tighter than item-assignment. The correct root cause and fix (ternary-lvalue compound/simple assignment) landed in #5180. The T-028 Open entry this PR edits no longer exists (moved to Done), so it also conflicts with main.

PR #5290, opened 2026-07-23
fix: List holds scalar-variable containers (container aliasing)

tokuhirom:
## Summary

A List built from a parenthesized comma expression (`($a, $b)`) must hold the **container** of each scalar variable element, not a snapshot of its value. A later mutation of the variable is then visible when the List is read. This is the [traps.rakudoc](raku-doc/doc/Language/traps.rakudoc) "list-element container aliasing" behavior (PLAN §8.9 [5]/[6]/[7]).

Before:
```raku
my $a = 2; my $l = ($a, $a); $a = 99; say $l;   # mutsu: (2 2)   raku: (99 99)
my $a = 2; say join ",", ($a, ++$a);            # mutsu: 2,3     raku: 3,3
```

The fibonacci push trap from traps.rakudoc:
```raku
my @arr; my ($x, $y) = (1, 1);
for ^3 { ($x, $y) = ($y, $x + $y); @arr.push: ($x, $y); say @arr }
# raku:  [(1 2)] / [(2 3) (2 3)] / [(3 5) (3 5) (3 5)]
# mutsu (before): [(1 2)] / [(1 2) (2 3)] / [(1 2) (2 3) (3 5)]
```

mutsu previously `Itemize`d scalar-var List elements *by value*, so the snapshot never tracked later mutations.

## Fix

Reuse the existing capture machinery:

- Compiler (`ArrayLiteral`): tag each scalar-var element with `WrapVarRef` instead of `Itemize`.
- VM (`exec_make_array_op`, the List path `is_real_array = false`): box the named local into a shared `ContainerRef` cell via `capture_var_cell` — the same primitive Captures (`\($a)`) already use.

Consumers that decontainerize already deref the cell, so the aliasing is invisible where Raku does not want it:

- `my @z = (...)` (array assignment) → by value
- function/`sub` parameter binding → by value
- single-scalar `push @a, $x` → by value
- bracket arrays `[...]` (use `MakeRealArray`) → decontainerize on construction, unaffected

## Tests

- New pin: `t/list-container-aliasing.t` (13 cases: aliasing, shared-container identity, the `++$a`/fibonacci traps, `push`, array-assign decont, bracket-array non-aliasing, `+$a` value-forcing, by-value function args, nested lists, non-flattening scalar elements).
- Verified against reference `raku` for every case.
- `cargo test` green (579+ unit tests); targeted `t/` (list/array/push/container/capture/for) green. Full `make test` + `make roast` delegated to CI (local runs blocked by concurrent-worker OOM pressure).

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1249,19 +1249,27 @@ the backlog.
       in `ArrayLiteral([$a])`; the loop's own `TagContainerRef`+writeback drives rw, so aliasing there
       wrote the cell back into `$a` = self-referential cycle = infinite loop; new compiler flag
       `suppress_list_var_alias` keeps Itemize while compiling the for-iterable — `t/for-param.t` fixed).
-- **Remaining (3 consumers not yet cell-aware — the campaign work):**
-  1. `t/list-dot-equals-method.t` — `($x,$y).=reverse` → `[35 35]` (want `[35 7]`). `.=`-on-list lowers
-     to `__mutsu_assign_callable_lvalue(ArrayLiteral-lvalue, [], MethodCall{target: ArrayLiteral})`; the
-     lvalue-target and RHS-target Lists both box cells and the assign-back corrupts. Fix: the lvalue
-     list-assign must deref cells to values on write-back (or suppress alias for the lvalue-target list).
-  2. `t/hash-itemization-flag.t` test 8 — `my %c = ($hi,)` (hash-holding scalar) should die "odd number",
-     but the cell derefs to the hash and *flattens*. Old Itemize-Scalar was non-flattening. Fix: the
-     hash-assign flatten path must treat a `ContainerRef`-wrapped value as itemized (non-flatten), as it
-     does a Scalar.
-  3. `t/list-infix-comma-precedence.t` tests 24/25 — `$a,$b X!=:= $c,$d` (uninit `my` scalars) all-True;
-     after `$c:=$b`, `one ... X=:=` True. The cross/zip metaop **derefs the cells** before `=:=`, losing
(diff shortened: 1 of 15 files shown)


tokuhirom:
**Deferred to a future session (WIP / draft).**

Status: the headline List container-aliasing behavior works and matches reference `raku` — `my $l=($a,$a); $a=99; say $l` → `(99 99)`, `join ",", ($a, ++$a)` → `3,3`, the fibonacci `@arr.push: ($x,$y)` trap, nested lists, `+$a` value-forcing, by-value function args, array-assign decontainerization, bracket-array non-aliasing. Pin `t/list-container-aliasing.t` is 13/13. The for-loop self-cycle regression (`for $a -> $v is rw`) is also fixed (`t/for-param.t`).

Not yet done — three List *consumers* are not `ContainerRef`-cell-aware, so CI is red on:
1. `t/list-dot-equals-method.t` — `($x,$y).=reverse` (lvalue list-assign corrupts cells)
2. `t/hash-itemization-flag.t` — `my %c = ($hi,)` (hash-assign flattens the cell instead of dying on odd count)
3. `t/list-infix-comma-precedence.t` — `$a,$b X!=:= $c,$d` (cross/zip metaop derefs cells, losing `=:=` container identity)

The sound fix is to make those consumers treat a `ContainerRef` cell transparently, the same way they treat an `Itemize`-Scalar (deref for value, keep the cell for identity). This is the first-class-scalar-container direction; coordinate with ADR-0001. Full plan in PLAN.md §8.16 and memory `project_list_container_aliasing_campaign`. **Do not merge until all three consumers are cell-aware and CI is green.**

PR #5374, opened 2026-07-24
docs: triage the two remaining Zef battery test failures

tokuhirom:
The battery gate entry for Zef said the two failing files were "not yet triaged: either a real regression since 2026-07-10, or a run-context difference". They are neither — both are **real, general mutsu bugs** reachable only through paths the older mzef/install run did not take, and neither is Zef-specific. This records what each one actually is.

## `00-load` (1/2) — `subtest` rolls back the type registry but not `loaded_modules`

`test_fn_subtest` restores `classes`, `roles`, `subsets`, `functions`, `proto_subs`, `token_defs`, … wholesale after the block. A module first loaded *inside* a subtest therefore has all of its declarations erased on exit — while `loaded_modules` still lists it, making every later `use` an early-return no-op. The type is permanently gone.

```raku
use Test;
plan 1;
sub probe($t) { say "$t: ", ::('Zef::Fetcher').^name }
subtest 'A' => { use Zef; probe('inside'); };   # inside: Zef::Fetcher
probe('after');                                  # after: Failure   <-- wrong
```

In the real file `subtest 'Core'` loads `Zef`, and `subtest 'Plugins'` then dies with `X::InvalidType: Invalid typename 'Fetcher'` on all three `Zef::Service::Shell::*` files.

## `distribution-depends-parsing` (18/35) — two bugs, one masking the other

1. `find-prereq-candidates` throws `X::Method::NotFound: No such method 'specs' for invocant of type 'Zef::Distribution::DependencySpecification'`. `.specs` exists only on the sibling class `…::DependencySpecification::Any`, and every caller is a `multi` candidate constrained to that sibling — so a plain spec is reaching an `::Any`-constrained candidate. Recorded as **not yet reduced**: the obvious standalone shapes (sibling classes doing one role, with and without a `where`-constrained `multi submethod new`) all dispatch correctly here, so the trigger is something else in those signatures.
2. ★ That uncaught exception aborts the file **with no message at all** — the run just stops after test 18 and only the plan mismatch is reported. Wrapping the call in `try`/`CATCH` is what revealed the error. This is worth fixing on its own, and first, since it is what made the file read as a plan bug.

Docs-only; no label (default patch bump).

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -139,12 +139,52 @@ section.
         `Blob:D`, which made `OpenSSL::Digest` mutually recurse — the `05-digest`/`03-rsa` hangs),
         `unit module` package scoping (#5369), positional-argument indexing (#5370), and the
         `where`-constraint caller-lexical wipe (`04-crypt`, PLAN 8.22).
-      - **Zef — 8/10** (`00-load` 1/2, `distribution-depends-parsing` 18/35). ★ This **contradicts
-        the recorded "all 10 upstream tests pass" (2026-07-10, #4383/#4384)**. Not yet triaged:
-        either a real regression since then, or a run-context difference (the gate runs
-        `mutsu -I vendor/zef/lib <test>` directly, whereas the 2026-07-10 result came from the full
-        mzef/install context). **Determine which before assuming a regression** — if it is context,
-        the gate's invocation may need to match how zef actually runs its own tests.
+      - **Zef — 8/10**, now triaged (2026-07-25). Both remaining files fail for **real, general mutsu
+        bugs**, not for a run-context difference, so the "all 10 upstream tests pass" note from
+        2026-07-10 (#4383/#4384) is stale rather than contradicted — these two are reachable only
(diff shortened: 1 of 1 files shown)


tokuhirom:
Updated: bug 1 is now **reduced to a standalone repro**, so the entry no longer says "not yet reduced".

A type constraint whose qualified name ends in `::Any` is resolved as the **builtin `Any`**, so it matches every value and its candidate wrongly wins:

```raku
class Spec { }
class Spec::Any { }
class Other {
    multi method sm(Spec::Any $s) { "ANY" }
    multi method sm($s)           { "GENERIC" }
}
say Other.new.sm(Spec.new);   # raku: GENERIC   mutsu: ANY   <-- wrong
```

Renaming the class to `Spec::Alt` makes it correct, which pins the cause to the trailing-component fallback in type-name resolution rather than to multi-dispatch ranking. Fixes for this and for the silent-exception bug follow in a separate PR.

PR #5404, opened 2026-07-25
docs: record why a user `method message` on an exception never runs

tokuhirom:
An exception class that COMPUTES its message reports the literal text `(Any)`:

```raku
class X::HTTP is Exception { has $.rc }
class X::HTTP::Response is X::HTTP {
    has $.message;
    method message { $!message //= "Response error: '$.rc'" }
}
X::HTTP::Response.new(:rc('404 Not Found')).throw;
# raku:  Response error: '404 Not Found'
# mutsu: (Any)
```

**Root cause located.** `methods_instance_ops.rs:905` answers `gist`/`Str`/`Stringy`/`message` for any exception-ish instance straight from the stored `message` attribute. Its comment claims "`message` itself only reaches here when no user method exists" — but user-method resolution happens ~300 lines *later* in the same function (`:1204`), so the arm always wins. `has $.message` leaves the cell `Any` until the method computes it.

**Why it is not a one-liner.** Skipping the arm when the class has a user method of that name does fix the direct case — and then a *write* to the private attribute inside the same-named method recurses forever:

```raku
class X::B is Exception { has $.message; method message { $!message = "W"; "R" } }
try { X::B.new.throw; CATCH { default { say .message } } }   # mutsu: hangs
```

Reading `$!message` is fine; only the write loops, i.e. the `$!attr` store path re-enters accessor/method dispatch. That is the same private-attribute write path as PLAN.md §8.21 (`$!attr = v` skipping the declared type check), so the two belong in one fix at the pre-store choke point §8.21 already argues for.

The entry also records the three secondary sites that derive a thrown exception's text from the `message` ATTRIBUTE rather than asking the method (`builtins/methods_0arg/mod.rs:1621`, `methods_call_dispatch.rs:548` and `:2121`), why the decision has to live in the native-bypass gates, and that `throws-like` has the same shape (raku evaluates `$exception."$key"()`).

Docs only — the code changes explored while diagnosing this were reverted, since the fix cannot land without the `$!attr` write path.

Impact: HTTP::UserAgent `t/082-exceptions` subtest 3, the last non-network failure in that suite (otherwise 25/27 after #5401).

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1589,6 +1589,55 @@ token path-part:sym<matcher> {
   3/3 and `t/synopsis.rakutest` runs all 9 planned tests, but 6 of them
   fail because `@received` collects these markers instead of the emitted
   values. This is the last known blocker for that dist.
+### 8.24 A user `method message` on an exception never runs, and fixing it exposes a `$!attr` self-recursion (found 2026-07-25, HTTP::UserAgent)
+
+An exception class that COMPUTES its message reports the literal text `(Any)`:
+
+```raku
+class X::HTTP is Exception { has $.rc }
+class X::HTTP::Response is X::HTTP {
+    has $.message;
+    method message { $!message //= "Response error: '$.rc'" }
+}
+X::HTTP::Response.new(:rc('404 Not Found')).throw;
+# raku:  Response error: '404 Not Found'
+# mutsu: (Any)
+```
+
+- **Root cause (located).** `methods_instance_ops.rs:905` answers
+  `gist`/`Str`/`Stringy`/`message` for any exception-ish instance straight from
+  the stored `message` attribute. Its comment claims "`message` itself only
+  reaches here when no user method exists", but user-method resolution happens
+  ~300 lines LATER in the same function (`:1204`, via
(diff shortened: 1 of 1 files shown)


tokuhirom:
Superseded by #5414, which FIXES this rather than recording it.

The analysis in this PR was correct about the root cause (`methods_instance_ops.rs`'s `gist|Str|Stringy|message` arm answering from the stored attribute ~300 lines before user-method resolution) and it is preserved in `news/2026-07/exception-computed-message.md`, together with the two things this record did not yet know:

- `.throw` reads the attribute in three places, and the pure-value NATIVE render path cannot see the class registry at all — so the fix needed two bypass gates (`vm_native_dispatch::try_native_method` and `should_bypass_native_fastpath`) on top of the arm change.
- The predicted `$!attr`-write self-recursion did not materialise: with the arm falling through to normal user-method dispatch, `method message { $!message = 'W'; 'R' }` terminates. It is pinned in `t/exception-computed-message.t`.

The `throws-like` shape noted here is fixed too — it now matches on the computed message.

PR #5418, opened 2026-07-25
feat(batteries): bundle HTTP::UserAgent, and keep third-party-service tests out of the release gate

tokuhirom:
## `HTTP::UserAgent` is bundled

The HTTP client slot is filled. `HTTP::UserAgent` v1.2.0 ships at `modules/HTTP-UserAgent/` and resolves with **zero config** — no `-I`, no install, no environment setup:

```raku
use HTTP::UserAgent;
say HTTP::UserAgent.new.get('https://example.com/').code;   # 200
```

That completes the bottom-up build order the battery plan set out: TLS foundation (`OpenSSL` + `IO::Socket::SSL`) → dependency layer (`URI`, `MIME::Base64`, `HTTP::Status`, `DateTime::Parse`, `Encode`, `File::Temp`, `File::Directory::Tree`) → client. A program can now make a real `https://` request using nothing but the shipped binary.

No wiring was needed: `resolve_bundled_lib_paths` enumerates `modules/*/lib` dynamically, and the release tarball and container already `cp -R modules` into `share/mutsu/modules`.

**Vendored, not patched** — `lib/` plus `META6.json` / `LICENSE` (MIT) / `README.md` / `Changes` for attribution; upstream `t/`, `xt/`, `doc/`, `examples/`, CI config and `.precomp` artifacts excluded. Every gap the upstream suite exposed was closed in the *interpreter*, ending with #5416 which took the suite to 27/27.

## New: the gate never runs third-party-service tests

`batteries-whitelist.txt` goes **53 → 75**. 23 of the suite's 27 files are gated. Three (`110-redirect-cookies`, `230-binary-request`, `250-issue-144`) need `Test::Util::ServerPort`, a test-only dependency the harness cannot fetch — it clones only the battery's own repo — so they fail with `ok=0` (module missing), not a failing assertion. They pass locally.

The fourth, `082-exceptions`, exposed a real hole in the gate: it makes **unguarded live requests to `httpbin.org`**, which spent part of 2026-07-25 returning 503 — exactly while this battery was being bundled. A file whose verdict depends on a third-party service must not be able to block a mutsu release.

So the harness grew **`batteries-exclude.txt`**: `name<TAB>testfile` entries skipped entirely, in both gate and `--update` mode, so they can neither fail a release nor re-enter the baseline. A whitelisted-but-excluded file reports a `note:` telling you to re-run `--update` rather than a regression; the gate's teeth are otherwise unchanged.

Which files belong there was **measured, not guessed**: every whitelisted file was re-run inside a loopback-only network namespace (`unshare -rn -- sh -c 'ip link set lo up; …'`). Only **two of 77** failed — this one, and `IO::Socket::SSL/01-basic`, which connects to `github.com:443` in its first two lines, *before* its own `NETWORK_TESTING` guard. Everything else already routes live-network assertions through `NETWORK_TESTING`, which the gate does not set. Those two are the whole exclusion list.

(Fetching each suite at its pinned commit still needs the network — that is setup, and a fetch failure reports `GATE ERROR`/exit 2, distinct from the `GATE FAILED` a regression produces.)

## Verification

- `make test`: 2450 files, 23364 tests, **0 failures**
- Gate `--update` diff is `+23 HTTP::UserAgent`, `−1 IO::Socket::SSL/01-basic` (now excluded) — no other file dropped
- Gate mode: **GATE PASSED**
- Exclusion mechanism exercised both ways against a scratch manifest/baseline: an excluded-but-whitelisted file yields `note:` + `GATE PASSED`; a whitelisted file that does not pass still yields `GATE FAILED`
- Zero-config load stressed 70× (debug + release) with no failure

Docs: `BATTERIES.md` bundle index, `docs/batteries/http-client.md` (status, provenance + re-vendor recipe, gate facts), `docs/batteries/testsuite-gate.md` (a "what the gate does not run" section), regenerated `wasm-demo/content/batteries.json`, and `PLAN.md` loses the now-done battery task.


tokuhirom:
## Follow-up: `Test::Util::ServerPort` is bundled too — the gate is now 80/80

The three `HTTP::UserAgent` files that could not be gated (`110-redirect-cookies`, `230-binary-request`, `250-issue-144`) are the **redirect and binary-upload paths**, driven against a local `TestServer` — exactly the paths most likely to regress. Their only missing piece was `Test::Util::ServerPort` (zero deps, ~30 lines, Artistic-2.0), so it is bundled at `modules/Test-Util-ServerPort/`.

It earns its place beyond that suite: the batteries goal is "a small web blog can be written with the bundle alone", which implies *testing* one, and any test that starts a listener needs a free port. Hardcoding one breaks the moment two runs overlap — mutsu's own `t/io-socket-recv-limit.t` was misdiagnosed as flaky for months over exactly that.

`HTTP::UserAgent`'s other two `test-depends` needed nothing: mutsu already provides `Test::IO::Capture` and `JSON::Fast`.

### Result

- `batteries-whitelist.txt` **75 → 80**, and **every file in it passes** (80/80, 2 excluded). 26 of the `HTTP::UserAgent` suite's 27 are gated — only the httpbin.org one is excluded.
- Re-measuring in a loopback-only network namespace now finds **zero offline failures across the whole baseline**: with the two excluded files out, the gate's verdict no longer depends on any third-party service at all.
- `make test`: 2450 files, 23364 tests, **0 failures**.
- Verified `use Test` still resolves to the built-in module (the new tree contributes `Test/Util/ServerPort.rakumod`, not `Test.rakumod`).

New record: `docs/batteries/test-helpers.md`.

tokuhirom:
Correction to my earlier comment: **mutsu does not provide `Test::IO::Capture`.** `070-ua-simple.rakutest` passes with no `IO::Capture::Simple` present because its whole body sits behind `NETWORK_TESTING` *and* a missing `Test::*` module is silently tolerated at `use` time (`runtime/runtime_module.rs`). `JSON::Fast` really is provided natively (`runtime/json.rs`); conflating the two was wrong. Fixed in the docs, and the tolerance is recorded as a wart in `todo/tickets/bundle-json-tiny-instead-of-emulating.md` — which also sizes the "bundle the real JSON dists" question: upstream JSON::Tiny is pure Raku and already runs on mutsu at 134/135 subtests, while JSON::Fast needs 52 distinct nqp ops (389 calls).

PR #5535, opened 2026-07-28
fix(vm): self is lexical inside a block, even when the block runs elsewhere

tokuhirom:
## What

`self` in Raku is a lexical: a bare block has no invocant of its own, so a `self` inside it — and the `$!attr` / `$.attr` forms that desugar through it — resolves outwards to the enclosing method's invocant. mutsu resolved it **dynamically** instead; the closure's captured `self` was deliberately not allowed to overwrite whatever the running frame held.

That is invisible while the block runs in the frame that created it, and wrong the moment it escapes:

```raku
class Conn { method protect(Callable $code) { $code() } }
class SH {
    has $.parent;
    has $!stmt = 'STMT';
    method execute() { $!parent.protect: { $!stmt } }
}
SH.new(parent => Conn.new).execute;
```

The block runs inside `Conn.protect`, where the live `self` is the `Conn`, so the read failed with `P6opaque: no such attribute '$!stmt' on type Conn`. The message was misleading twice over — the attribute name was right and the invocant was simply the wrong object.

## Why it matters

This is what stopped mutsu one line short of running a query against MariaDB. `DBDish::mysql::StatementHandle.execute` hands the statement's work to its connection:

```raku
$!parent.protect-connection: {
    $!stmt.mysql_stmt_execute or $!Prefetch and $!stmt.mysql_stmt_store_result;
}
```

`DBDish::Pg` shares the same `DBDish::StatementHandle` role, so it was going to fail on the same line.

The recorded diagnosis (`todo/tickets/self-not-restored-after-a-private-method-chain.md`) blamed a private-method call chain leaving `self` unrestored, because the failing read sits two statements after `self!enter-execute`. That was wrong — the private chain restores `self` fine. The ticket is removed and the finding recorded in `news/`.

## How

Make a captured `self` overwrite the running binding at closure entry:

- `src/vm/vm_closure_dispatch.rs` — the captured-env merge in `call_compiled_closure_with_topic`.
- `src/runtime/resolution_map_grep.rs`, `resolution_map_grep_rw.rs` — the interpreter's native `map` / `grep` / `first` loops pre-insert a closure's captured env the same way; they now also list `self` among the keys they restore on exit (two of them already overwrote it and never put it back).

A method's own invocant is bound from its arguments *after* that merge, so a real method call still binds its own `self` — including a block installed via `.^add_method`.

## Tests

`t/self-is-lexical-in-blocks.t` (14 assertions, all matching `raku`): the escaping block, the pointy block, the anonymous sub, a block returned out of the method that made it, per-instance capture, a method called from inside such a block, and the `map` / `grep` / `first` / `.()` / `start` invocation paths.

Local `make test` (2556 files, 24457 tests) and `make roast` (1435 files, 218774 tests) both PASS.


tokuhirom:
Follow-up pushed: the same overwrite had to be excluded from the exit writeback, or the *creator's* invocant is left behind in the frame that ran the block — `DBDish::Connection.protect-connection` then called `self.unlock-connection` on the StatementHandle. Two more pins in `t/self-is-lexical-in-blocks.t` (16 total). With both commits mutsu runs a real prepared INSERT against MariaDB.

PR #5687, opened 2026-08-01
fix(vm): a `try` does not re-catch an already-handled `Failure`, and `use fatal` no longer leaks out of an EVAL

tokuhirom:
A `try` whose body yields a *live* `Failure` catches it: the Failure becomes handled and `$!` holds its exception. mutsu did that for **any** trailing Failure, including one that `.defined` / `.Bool` / `.so` had already handled — and by then it is an ordinary value with nothing left to catch:

```
$ raku  -e 'my $f = "foo"[2]; $f.defined; try { $f }; say (defined $!) ?? "died" !! "lived"'
lived
$ mutsu -e 'my $f = "foo"[2]; $f.defined; try { $f }; say (defined $!) ?? "died" !! "lived"'
died
```

No `EVAL` and no module are involved; `"foo"[2]` produces the `Failure` directly. `TryCatch`'s success path now skips a Failure whose handled flag is already set, so `$!` stays undefined. The unhandled case is unchanged — `try { "foo"[2] }` still sets `$!` to the `X::OutOfRange`, which is what raku does too.

## Why it mattered

rakudo's `Test.rakumod` writes `lives-ok` as

```raku
multi sub lives-ok(Callable $code, $reason = '') is export {
    try { $code(); }
    my $ok = proclaim((not defined $!), $reason) or _diag($!);
    ...
}
```

so any `lives-ok { ... }` whose block ends in a handled `Failure` was reported as *died*. Found through the Test-vendoring sweep (`todo/tickets/vendor-real-test-module.md`).

## How it was tracked down

The tell was that `$f.handled` read `True` both outside and *inside* the `try`, so the flag was being set correctly and something was ignoring it. A breakpoint on `Value::is_failure_handled` showed it consulted exactly once, with the right instance id, returning `true` — so `OpCode::ThrowIfFailure` was not the thrower at all, which is what an `eprintln!` at the throw site would never have revealed. That left the `TryCatch` arm, which converts a trailing Failure to `$!` without asking.

## Recorded rather than fixed

A second cause behind the same symptom: a `throws-like` immediately before a `lives-ok` leaves `$!` set where the `lives-ok`'s own `try` does not reset it, so the next assertion reads a stale error. `todo/tickets/error-var-leaks-past-a-later-try.md` — it is the `$!`-scoping family of #5420, but in the opposite direction (a write that should be shadowed is seen by a later frame). It is what still keeps `t/statement-call-sinks-its-value.t` red under the aliased module.

## Testing

- `t/try-does-not-recatch-a-handled-failure.t` — 7 assertions, **green under `raku` too**.
- `make test`: `Files=2728, Tests=26071 … Result: PASS`.


tokuhirom:
#5688 merged into this branch, so this PR now carries **two** fixes for the same symptom. Both are in `news/2026-08/`.

**1. A `try` does not re-catch an already-handled `Failure`** (the original commit).

**2. `use fatal` no longer leaks out of an `EVAL`.** A pragma the EVAL'd unit turns on is scoped to that unit; mutsu keeps `use fatal` as an interpreter-wide flag and never restored it, so the caller went on throwing for every later soft `Failure`:

```
$ raku  -e 'use MONKEY-SEE-NO-EVAL; try { EVAL q{use fatal; "foo"[2]} }; my $f = "bar"[5]; say "soft: " ~ $f.^name'
soft: Failure
$ mutsu -e '...same...'
Index out of range. Is: 5, should be in 0..0
```

`throws-like 'use fatal; …'` is a common assertion shape and the real `Test.rakumod` EVALs its string, so one such assertion poisoned the rest of the file. `t/statement-call-sinks-its-value.t` is green under all three runs now.

The news entry for (2) also records a **correction**: the first diagnosis blamed a stale `$!` and was filed as a ticket. Printing the state between the two assertions killed that story in one run — `$!` was undefined, and the throw happened a line *earlier* than the `lives-ok`, at the `EVAL` that was supposed to produce a soft Failure. The wrong ticket is deleted here.

Testing: `t/eval-does-not-leak-use-fatal.t` (6 assertions) and `t/try-does-not-recatch-a-handled-failure.t` (7), both green under `raku` too; `make test` `Result: PASS` on each commit.

Issue #16, opened 2026-02-17
Implement Test::Util: is_run / get_out (subprocess execution)


tokuhirom:
## Summary

Implement `is_run` and `get_out` from Test::Util. `is_run` is the **second most used** Test::Util function (78 occurrences) and is critical for integration-style roast tests.

## Functions

- `get_out(Str $code, Str $input?, :@args, :@compiler-args) --> Hash` — execute Raku code as a subprocess, return `{status, out, err, test_died}`
- `is_run(Str $code, Str $input?, %expected, Str $name?, :@args, :@compiler-args)` — run code and validate stdout/stderr/exit status via smartmatch
- `run(Str $code)` — wrapper around `get_out`, returns stdout only

## Implementation notes

- `get_out` creates a temp file with the code, runs `$*EXECUTABLE $tmpfile` via shell, captures output
- Need to set `$*EXECUTABLE` to the mutsu binary path
- Requires `Proc` or direct `std::process::Command` in the runtime
- Quote escaping for command-line args

## Prerequisites

- `make-temp-file` (#15)
- Smartmatch (`~~`) for output validation (may already be partially implemented)

## Impact

Unblocks S04-statements/quietly.t, S04-statements/return.t, S32-io/note.t, integration/weird-errors.t, etc.

tokuhirom:
Resolved. `is_run`/`get_out`/`run` are implemented in `src/runtime/test_functions.rs` and verified by `t/is-run.t` (14 subtests pass, including subprocess exec, stdout/stderr/status smartmatch, and stdin-without-trailing-newline).

PR #5759, opened 2026-08-02
vm: record env consumer slot sets

tokuhirom:
## Problem

The five env-by-name consumers are represented by one frame-wide captures_env_by_name bit, preventing precise lexical-slot synchronization.

## Approach

- add Proposed ADR-0018 defining slot authority, per-consumer synchronization, block restore, and capture-cell invariants
- add separate consumer bitmaps while preserving current runtime behavior
- pin the migration-layer metadata

## Tests

- cargo check
- cargo test env_consumers_are_recorded_separately_before_blanket_removal

### DIFF
--- a/docs/adr/0018-slot-addressed-lexical-capture-and-env-sync.md
+++ b/docs/adr/0018-slot-addressed-lexical-capture-and-env-sync.md
@@ -1,0 +1,128 @@
+# ADR-0018: Slot-addressed lexical capture and precise environment synchronization
+
+- **Status**: Proposed
+- **Date**: 2026-08-02
+- **Deciders**: tokuhirom, Codex
+- **Related**: [ADR-0001](0001-gc-strategy-and-phasing.md), [ADR-0010](0010-cross-thread-lexical-sharing-scope.md), [ADR-0013](0013-container-interior-mutability-cellvalue.md), [ANALYSIS.md](../../ANALYSIS.md) §1.2/§1.3/§2.4, [lexical-scope-slot campaign](../lexical-scope-slot-campaign.md)
+
+## Context
+
+Compiled lexical reads and writes are slot-addressed, but several runtime mechanisms still
+observe or publish those lexicals through the name-keyed `Env`. `CompiledCode` currently
+sets `captures_env_by_name` when it contains any `ForLoop`, `BlockScope`,
+`BlockLocalScope`, `MakeGather`, or `WheneverScope`. That one bit overrides the existing
+per-slot `needs_env_sync` analysis and makes every local an env-mirror target.
+
+The blanket is load-bearing because the five consumers do not yet carry a complete account
+of the slots they observe. Block exit has historically recovered mutations by name, loop
+machinery publishes its parameter and control state through env, gather and whenever execute
+stored bodies against a live env, and closure construction falls back to materializing that
+env. Removing only the blanket deterministically breaks all four mechanism families.
+
+The mirror is also unsound as lexical identity. Two simultaneously-live shadowed bindings
+with the same name collapse to one env key. A nested call can therefore replace a closure's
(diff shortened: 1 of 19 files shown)


tokuhirom:
## Fixed the red CI: `roast/S04-statements/gather.t` test 23

The failure was deterministic (same subtest in all three of `test`, `gc-stress`, `jit-stress`), not flake. Bisected over the stack to **4c926bc "compiler: bake for-loop source slots"**.

**Root cause.** `sync_state_locals_in_range` (`src/vm/vm_run_loop.rs`) read a `state` variable's fresh value back from the name-keyed env. That was only correct because a frame containing a for-loop had every local env-synced by the frame-wide blanket that commit removed. With the blanket replaced by the loop's own baked slots, a plain `state $t` in the body no longer mirrors to env, so the env-first read persisted the value the *declaration* seeded rather than the one the body wrote:

```raku
gather { for 1..3 { state $t = 0; $t = $t + 1; take $t } }   # got 1 1 1, want 1 2 3
```

gather.t 23 is the uniq example from S04-control.pod, where the state holds the previous element — so every duplicate got taken again (`1 2 2 3 3 3 4 4 4 4 5 6 6` instead of `1 2 3 4 5 6`).

**Fix.** The routine-exit sibling in `vm_closure_dispatch.rs:832` already solved exactly this and carries the reasoning: pick the store that is authoritative for the slot. A slot outside `needs_env_sync` has no env mirror to trust, so read the live slot; a slot that *is* env-synced may have been written by name (a dynamic `s///` replacement bumping a state leaves the slot at the pre-replacement value — roast state.t #16), so it keeps reading env first. The loop-body sibling was simply never updated to match.

**Verification.** `gather.t`, `S04-declarations/state.t`, `S04-statements/{for,loop,while}.t` all pass. New pin `t/state-in-loop-body-accumulates.t` covers the gather, plain `for` and `while` bodies plus the S04-control.pod uniq example, and matches `raku`'s output on all five.

Rebased onto the current tip (`263a61d` / #5773) before pushing.

Note for the campaign: this is a general shape worth grepping for — **any site that reads a lexical back from env by name is a latent bug once its slot stops being env-synced.** `sync_state_locals_in_range` and `vm_closure_dispatch`'s state sync were the same code written twice, and only one of them had been taught the rule.

tokuhirom:
## Second red CI fixed: the unit test in `vm: close precise gather sync gaps`

`cargo test --lib` failed on `opcode::const_pool_dedup::env_consumers_publish_only_their_selected_slots` (`left: [false, false]`, `right: [true, false]`) — which is why `test` and `gc-stress` died at ~2 min rather than in roast.

**Not a regression in the new logic.** The fixture pushed a `free_var_parent_slots` entry with *no* matching `free_var_syms` entry. No compiler run produces that state — `add_closure_code_baked` builds `free_var_parent_slots` by mapping over `free_var_syms`:

```rust
compiled.free_var_parent_slots = compiled
    .free_var_syms
    .iter()
    .map(|sym| sym.with_str(|s| self.local_map.get(s).copied()))
    .collect();
```

so the two are index-aligned and the same length. It only became visible when the fold switched to walking the syms (so it can fall back to a same-name lookup when the parent slot is absent): walking a vector the fixture left empty reached no slot at all. Fixed by seeding the sym alongside the slot, so the fixture matches the invariant the production builder guarantees.

Verified on the branch: `cargo test --lib` 640 passed, plus `t/scoped-overlay-gather.t`, `t/state-in-loop-body-accumulates.t`, `roast/S04-statements/gather.t` and `roast/S04-declarations/state.t` all green.

Also worth noting: the new `sync_state_locals` change applies exactly the `slot_authoritative` rule I put into `sync_state_locals_in_range` a few commits earlier — the two are now consistent, which is the right end state. Those were the same code written twice; if a third copy shows up, it needs the same rule.

PR #5792, opened 2026-08-03
fix(dispatch): select a `Pair (:key(...), :value(...))` candidate

tokuhirom:
A `Pair` unpacks as a capture with two *named* parts and no positional one —
`(2 => 'x').Capture` is `\(:key(2), :value("x"))` — so `Pair (:key($k), :value($v))`
destructures it, `Pair ($k, $v)` does not, and `Pair (:key($k))` does not either
(it leaves `value` unconsumed). mutsu's **binder** implements all three rules;
multi-dispatch **matching** did not, so such a candidate was never selected:

```raku
multi g(Pair (:key($k), :value($v))) { say "pair k=$k v=$v" }
multi g($other)                      { say "other $other" }
g(2 => 'x');   # rakudo: pair k=2 v=x     mutsu (before): other 2   x
```

`sub_signature_matches_value` disagreed with `bind_sub_signature_from_value` twice:

1. **Leftover positionals.** An all-named sub-signature consumes no positional
   elements, so `positional_idx` stayed at 0 and the "unconsumed positional
   elements" check rejected every candidate. The binder guards its matching error
   with "does this sub-signature have positional params at all"; matching now does too.
2. **A rename read as a destructure.** A named parameter's parens are either a
   rename (`:key($plan)`) or a genuine destructure (`:value((:key($d), :value(&t)))`,
   whose inner param carries its own sub-signature). Matching recursed into both, so
   a rename demanded a positional element from the candidate and anything but a plain
   scalar there (a `Pair`, an `Array`) failed. The binder already distinguishes the
   two with exactly this test.

Rejecting an under-specified named destructure is now explicit as well, mirroring
the check already applied to a `Capture` value.

## Why it mattered

`Test::Util`'s `group-of` is this exact shape:

```raku
sub group-of (
    Pair (Int:D :key($plan), Pair :value((Str:D :key($desc), :value(&tests))))
) is export is test-assertion { subtest $desc => { plan $plan; tests } }
```

`user_test_decl_beats_native` asks `args_match_param_types` whether the imported
routine really accepts the call before letting it beat the native provider. It
answered no, so `group-of` was answered natively even with the real module loaded —
and under `MUTSU_REAL_TEST=1` the native `subtest` and the module's own kept separate
counters, reporting `# You planned 2 tests, but ran 0` for a group whose assertions
had all passed. `news/2026-08/retired-native-test-util-overrides.md` meant to cover
this; the guard simply could not see the signature.

- Pin: `t/pair-subsignature-dispatch.t` (verified to fail on the unfixed build; passes
  verbatim under `raku`).
- `make test` PASS.
- Unblocks `roast/S03-metaops/cross.t` and `roast/S03-operators/arith.t` under
  `MUTSU_REAL_TEST=1` (together with #5791).

Part of `todo/tickets/vendor-real-test-module.md` step 1.

### DIFF
--- a/news/2026-08/block-declares-in-its-own-package.md
+++ b/news/2026-08/block-declares-in-its-own-package.md
@@ -1,0 +1,36 @@
+# A block declares in the package it was written in, whoever calls it
+
+`vm_closure_dispatch` already ran a closure body under the package the closure
+was *declared* in, so that nested-class short names and `our` variables resolve
+when a foreign frame invokes it. The restore was skipped when that package was
+`GLOBAL` — presumably as a "nothing to do" shortcut. It is not: the caller's
+package is whatever the *callee* set, so a file-scope block invoked from inside a
+module kept the module's package and everything it declared was named after the
+module.
+
+```raku
+# CallBlk.rakumod:  sub call-it(&blk) is export { blk() }
+use CallBlk;
+call-it { my class Inner {}; say Inner.^name }
+# rakudo: Inner        mutsu (before): CallBlk::Inner
+```
+
+`Test::Util`'s `group-of` invokes the block it is handed, so every class declared
+inside a `group-of N => 'desc' => { … }` was named `Test::Util::Foo`, and
+`roast/integration/error-reporting.t` compared
+
+```
+Cannot resolve caller foo(Test::Util::RT129800:U: :foo(Test::Util::Foo)); …
+```
+
+against rakudo's `foo(RT129800:U: :foo(Foo))`. That file only started failing
+once `news/2026-08/pair-subsignature-dispatch.md` let the *real* `group-of` run
+at all — mutsu's native provider had been answering it, and the native one
+invokes the block without crossing a module frame.
+
+The condition is now "the closure's package differs from the current one",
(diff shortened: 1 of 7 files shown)


tokuhirom:
Pushed a second commit: letting the *real* `group-of` run exposed a pre-existing package-scoping bug, and `roast/integration/error-reporting.t` caught it.

A closure body already ran under the package it was **declared** in, but the restore was skipped when that package was `GLOBAL`. The caller's package is whatever the *callee* set, so a file-scope block invoked from inside a module kept the module's package:

```raku
use CallBlk;   # sub call-it(&blk) is export { blk() }
call-it { my class Inner {}; say Inner.^name }
# rakudo: Inner        mutsu (before): CallBlk::Inner
```

`Test::Util`'s `group-of` invokes the block it is handed, so a `my class Foo {}` inside a `group-of` body became `Test::Util::Foo` and `X::Multi::NoMatch`'s message diverged from rakudo's. The native `group-of` never crossed a module frame, which is why this only surfaced now.

Full `make roast` (release, 1435 files) and `make test` both PASS locally with both commits.

PR #5885, opened 2026-08-04
fix: an anonymous state variable resets when its routine is re-entered

tokuhirom:
A bare `$` (`$++` / `++$`) is an anonymous state variable belonging to its enclosing block's **clone**. A named routine's own body is cloned once, at registration, so a `$` written directly in it keeps counting across calls; the blocks *inside* that routine are re-cloned on every call, so a `$` in one of them restarts. mutsu keyed the counter by its compile-time name alone, so it kept counting everywhere:

```raku
sub f() { (map { ++$ }, 1, 2, 3).join(',') }
say f();   # raku: 1,2,3   mutsu: 1,2,3
say f();   # raku: 1,2,3   mutsu: 4,5,6
```

That was the last wrong-digest cause in `Digest::RIPEMD`, whose output stage rotates the five hash words with `map { $_[[^5].rotate(++$)] }`: the second and later `rmd160(...)` calls in one process returned a correct-but-**rotated** digest. Each call was correct in a fresh process, which is what made it look like a hashing bug rather than a scoping one.

## The rule

> the counter resets iff the `$` is **lexically inside a nested block** that is **lexically inside a routine**, and then it resets once per *call of that routine* — not per block iteration.

Static except for the bucket, so it splits in two.

**Compile time.** `CompiledCode::anon_state_nested_depth` is a compile-only cursor; `add_constant` — the single choke point every variable-name constant passes through, and `__ANON_STATE_<id>__` is used for nothing else — records the ids it sees while the cursor is set into that chunk's `per_call_anon_states`. No emit site changed.

The set lives **on the chunk**, not in a registry keyed by id. A global registry looks sound (the parser mints one id per source occurrence) but the same source is compiled several times — the routine-hoist pass, `record_type_body_captures`' capture analysis, then the real body — and those passes do not all reproduce the true lexical context, so one wrong classification would poison an id for the whole process. This is exactly what shelved the first prototype (see #5884): it hung `roast/S32-list/rotor.t`.

Several runtime paths *re-compile* a block body from its AST and so lose the lexical position; the cursor is restated at those sites (`eval_map_over_items` and its grep/rw siblings, `exec_make_gather_op`).

The inline case is **opt-in** — `Stmt::For` arms `Compiler::anon_state_enable_next` for the one scope its body pushes, and only for the block form. It has to be opt-in because `Stmt::If`/`Stmt::While` carry no `is_statement_modifier` flag, so `$++ if C` (which must keep counting — `rotor.t`'s hand-written `Iterator` depends on it) cannot be told from `if C { $++ }`. The flag is cleared at the top of every `compile_stmt` so it can never leak.

**Run time.** `RoutineFrame` gained a monotonic `invocation_id`; `anon_state_key` folds in the innermost enclosing **non-block** frame's id for a marked name. At the mainline there is no such frame, so the id is constant and `[ $++ xx 3 ] xx 3` is still `0..8`.

**The env shadow.** A marked name must be read from the state store outright: these names compile to `GetGlobal`/`SetGlobal`, so their env entry outlives the block clone and on the first use in a new call the store legitimately misses while env holds the previous call's value. `per_call_anon_state_read` answers the site's default on a miss rather than falling through.

## Known residue (documented, unchanged from today)

A named sub nested *inside another routine* still persists where rakudo resets — `sub outer() { sub inner() { ++$ }; ... }` — because the runtime rule picks `inner`'s own frame rather than `outer`'s. No `roast` or `t/` test exercises it.

## Testing

- `t/anon-state-per-routine-call.t` (new, 14 tests) — both directions of the rule row by row; all 14 also pass under `raku`.
- Full local `prove t/`: 2869 files, 27232 tests, all pass.
- `roast/S32-list/rotor.t` (21 tests, the `$++ if C` iterator), `S04-statements/gather.t`, `S03-operators/context.t`, `S02-types/whatever.t`: pass.
- `rmd160("abc")` is correct on the second call in a process; the dist's `t/md5.t`, `t/sha.t` and `t/rfc4231.t` pass in full.

Closes `todo/tickets/anonymous-state-var-not-reset-per-routine-call.md` and `todo/tickets/digest-dist-blockers.md` §2 — every digest the `Digest` distribution provides now comes out correct.


tokuhirom:
Closing: the per-chunk classification has a key-stability hazard this branch does not solve.

`roast/S32-list/classify.t` test 39 (`<a b c>.classify({ ~($ ~= $_) })`) regressed. Several runtime paths re-compile a block body from its AST, and the fix restates the compile-time cursor at those sites — but a single `$` occurrence can then be executed by TWO chunks that classified it differently, so its write lands on `__anon_state::…#<bucket>` while its read looks up the plain key (or vice versa) and the counter appears to reset.

Dropping the recompile-site markings removes the hazard but also removes the fix: without them `map { ++$ }` — the `Digest::RIPEMD` case this exists for — never resets at all.

The finding is being recorded on `todo/tickets/anonymous-state-var-not-reset-per-routine-call.md` so the next attempt starts from it: the classification has to be attached to the `$` occurrence in a way that every chunk executing it agrees on, which neither a global registry (poisoned by analysis passes, see #5884) nor a per-chunk set (this PR) provides on its own.

PR #6005, opened 2026-08-06
fix(vm): a non-trivial proto body resolves nested subs against its own table

tokuhirom:
## Summary
- `vm_try_run_nontrivial_proto_body` OTF-compiles a proto with a real body (`proto foo($x) { ...; {*} }`) and passed the *caller's* `compiled_fns` table straight through to `call_compiled_function_named` instead of the freshly compiled body's own (`cf.compiled_fns`).
- This is the same "wrong nested-sub table" shape already fixed for `call_shared_state_body` while investigating the ADR-0019 C6e-3c blocker — closes the follow-up ticket that flagged this sibling site as unaudited.
- Fix: `let fns = cf.compiled_fns.as_deref().unwrap_or(compiled_fns);` before the call, mirroring the established pattern.

## Test plan
- [x] New regression test `t/nontrivial-proto-body-own-nested-sub-table.t` against a new fixture module `t/lib/NontrivialProtoNestedSub.rakumod` (non-trivial proto body with a nested `my sub` before `{*}`, imported cross-module).
- [x] `make test` — 27742 tests, all passing.
- [x] `cargo clippy -- -D warnings` and `cargo fmt` clean (pre-commit hooks).


tokuhirom:
Duplicate of #6004 (opened earlier, same fix — identical one-line change in `vm_try_run_nontrivial_proto_body`, same root cause, own pinned test). Closing to avoid a merge conflict between the two.

PR #6010, opened 2026-08-07
fix(vm): a package-qualified class/role's short name is package-scoped, not global

tokuhirom:
## Summary
- `class Cro::Hdr { }` used to install the bare name `Hdr` into the interpreter's single flat, global `env`, making it visible from anywhere in the process for the rest of the run and letting it silently clobber an unrelated same-short-name declaration in a totally different scope.
- `exec_register_class_op`/`exec_register_role_op` (`src/vm/vm_typedecl_ops.rs`) now record the short-name alias in `package_type_aliases`, keyed by the *declaring parent package*, instead of inserting into `env`. Bareword resolution already walks that table via the existing `method_class_stack`/`current_package` package-chain machinery (`package_type_alias`/`lookup_in_running_package`), so no new resolution code was needed — only the write side had to stop leaking globally.
- This was the live blocker behind `Cro::HTTP::ResponseParser`/`Cro::HTTP::RequestParser`, where `Cro::HTTP::Header` kept stealing the bareword `Header` from an unrelated `my enum Expecting <StatusLine Header Body>` inside a `supply` block.
- Verified the counter-example that sank an earlier naive gate (`class URI::Path` declared at file scope, referenced bare as `Path` inside `unit class URI`'s own methods/attribute defaults) still resolves correctly — `push_method_class` and `eval_attr_default_expr` already unconditionally anchor `method_class_stack`/`current_package` to the owning class.
- See `news/2026-08/package-short-name-alias-is-package-scoped.md` for the full writeup; ticket `todo/tickets/package-short-name-alias-is-global.md` moved there.

## Test plan
- [x] Ticket's own two repro cases now match `raku` behavior (bareword no longer leaks globally; local enum member inside `supply` block is no longer clobbered)
- [x] `class URI::Path` / `unit class URI` regression check (bundled `modules/URI`) still resolves `Path` bare inside `URI`'s own methods
- [x] `role Cro::HTTP::Middleware::Pair` still doesn't shadow the built-in `Pair` type
- [x] `prove -e ./target/debug/mutsu t/ -r -j4` — full local suite, 2929 files / 27755 tests, clean (run twice)
- [x] Targeted roast package/module tests (`S10-packages`, `S11-modules`, `S12-class`, `S02-packages`) pass
- [x] `cargo clippy -- -D warnings` and `cargo fmt --check` clean
- [ ] CI (`make test` + `make roast`)


tokuhirom:
Reverting: this approach breaks a real, working ecosystem module (DBIish's Postgres driver).

Root cause of the CI failures: `class PGconn is repr('CPointer')` is declared (non-`my`, package-scoped) inside `unit module DBDish::Pg::Native`. A sibling class `DBDish::Pg` (`unit class DBDish::Pg ... does DBDish::Driver`) does `use DBDish::Pg::Native;` and references the bare name `PGconn` from inside a **method body** (`method connect(...) { ... PGconn.new(...) ... }`).

My fix resolves package-scoped short names by walking only the *ancestor* chain of `current_package`/`method_class_stack` (anchored to the receiver class during method dispatch — here, `DBDish::Pg`). But `DBDish::Pg::Native` is a **sibling** of `DBDish::Pg`, not an ancestor, so the lookup can never find it. The old global-env write (however architecturally unsound) happened to make this common "import a module, use its exported class's bare name from a sibling package" idiom work; my package-chain-only approach breaks it outright, turning a working program into "Undeclared name: PGconn".

Minimal repro (fails on this branch, passes on `main`):
```raku
# lib/Foo/Native.rakumod
unit module Foo::Native;
class PGconn is export is repr('CPointer') { }

# lib/Foo/Driver.rakumod
unit class Foo::Driver;
use Foo::Native;
method make() { PGconn.new }
```

A correct fix needs real per-scope import tracking (what did *this* file/class actually `use`), not ancestor-chain walking — i.e. the "design pass" the original ticket asked for. Restoring the ticket with these findings so the next attempt starts here instead of repeating this regression.

PR #6087, opened 2026-08-08
fix(vm): Supply.do on an on-demand source is no longer a permanent dead end

tokuhirom:
## Summary
- `Supply.do($cb)`'s non-live fallback in `native_supply_dispatch.rs` copied only `values` (always empty for a `supply { ... }` on-demand source) and dropped `on_demand_callback` entirely, so the derived Supply had no source at all — a permanent dead end. Nothing ever tapped it produced anything.
- This hung the vendored Cro suite's `http-auth-basic.rakutest`, `http-auth-basic-with-session.rakutest`, and `http-router.rakutest` forever (`rc=124`, zero TAP output — not even the plan line), because `Cro::HTTP::Auth::Basic.process-responses` is exactly `$responses.do(...)` on the server's on-demand response pipeline.
- Fix: when the source has `on_demand_callback`, carry it forward instead of the (empty) `values`. The existing tap-dispatch machinery already re-runs the on-demand body and applies the accumulated `do_callbacks` chain generically on tap — it just never received the callback before.
- A distinct, deeper bug remains where `do_callbacks` are skipped for values an on-demand source delivers *asynchronously* via a nested `whenever` (vs. a synchronous `emit`), which is what's needed for Cro's `WWW-Authenticate` header to actually reach a 401 response. Filed as `todo/deep/supply-do-callbacks-not-applied-to-async-emitted-values.md` for follow-on work.

## Test plan
- [x] `target/debug/mutsu t/supply-do-on-demand-source.t` — new regression test, 3/3 pass, matches `raku` output exactly
- [x] `make test` — full local suite passes (2966 files, 27965 tests)
- [x] Manually verified `http-auth-basic.rakutest` no longer hangs (was `rc=124`/0 output, now completes with 3/5 subtests passing)


tokuhirom:
## Summary
Two related fixes to `Supply.do($cb)` on an on-demand (`supply { ... }`) source, which was a permanent dead end. Together they turn the vendored Cro suite's `http-auth-basic.rakutest` and `http-auth-basic-with-session.rakutest` from a permanent hang (`rc=124`, zero TAP output) into fully passing (5/5).

### Commit 1 — the dead end itself
`native_supply_dispatch.rs`'s `"do"` arm's non-live fallback copied only `values` (always empty for a `supply { ... }` on-demand source) and dropped `on_demand_callback` entirely, so the derived Supply had no source at all. Fix: carry `on_demand_callback` forward instead when present; the existing tap-dispatch machinery already re-runs the on-demand body and applies `do_callbacks` generically on tap.

### Commit 2 — async-delivered values still skipped `do_callbacks`
With the dead end fixed, `.do`'s callback still only ran for values a source's body `emit`s **synchronously**. A value delivered later through a nested `whenever` (the shape of essentially every non-trivial `supply` block, and specifically how Cro's response pipeline works) bypassed `do_callbacks` — the real subscriber was registered directly on the emitter with no wrapping. Fix: wrap the registered subscriber with the source's `do_callbacks` chain (`register_outer_tap_with_do_callbacks`, a small marker `Value` unwrapped in `call_supply_tap`, the single choke point every registered supplier tap goes through).

Both bugs are general-purpose Supply/on-demand-source bugs, not Cro-specific — reproduced and verified independently of any Cro code (see the `t/` tests).

## Test plan
- [x] `target/debug/mutsu t/supply-do-on-demand-source.t` — 3/3 pass, matches `raku`
- [x] `target/debug/mutsu t/supply-do-on-demand-async-emit.t` — 2/2 pass, matches `raku`
- [x] `make test` — full local suite passes (twice, once per commit)
- [x] Manually verified `http-auth-basic.rakutest` and `http-auth-basic-with-session.rakutest`: was `rc=124`/0 output → now 5/5 passing
- [x] Re-ran the full vendored Cro::HTTP and Cro::Core suites — no regressions, only the two files above newly green

PR #6105, opened 2026-08-09
fix(operator): scope block-local infix-op overrides to their declaring source file

tokuhirom:
## Summary

- User-defined `infix:<op>` subs declared inside a block (e.g. a test script overriding `+`) were incorrectly active inside called module code, corrupting e.g. Test.rakumod's internal arithmetic counter
- Converts `user_declared_infix_ops` from `HashSet<String>` to `FxHashMap<String, Option<String>>` where `None` = exported/universal, `Some(path)` = active only when the executing compiled function comes from that source file
- Adds `executing_cf_source_file: Option<String>` to the Interpreter, updated on every compiled-function entry/exit across all call paths and during module loading
- Exported operators use `insert(None)` (not `or_insert`) to properly upgrade file-scoped declarations to universal visibility

## Effect

- `roast/integration/advent2013-day10.t` under `MUTSU_REAL_TEST=1`: 44/44 (was 12/44)
- `roast/S06-operator-overloading/` suite: 223/223 (unchanged, including `imported-subs.t` 20/20)
- `make test`: PASS

## Test plan

- [x] `MUTSU_REAL_TEST=1 MUTSU_FUDGE=1 prove -e target/debug/mutsu roast/integration/advent2013-day10.t` → 44/44
- [x] `MUTSU_FUDGE=1 prove -e target/debug/mutsu roast/S06-operator-overloading/` → 223/223
- [x] `make test` → PASS

### DIFF
--- a/news/2026-08/user-infix-source-file-scoping.md
+++ b/news/2026-08/user-infix-source-file-scoping.md
@@ -1,0 +1,41 @@
+# Scope block-local infix-op overrides to their declaring source file
+
+Previously, a user-defined `infix:<op>` sub declared anywhere in a test script
+(e.g. overriding `+` for a custom `NotANumber` type) was active globally —
+including inside module code called from that script.  This corrupted
+Test.rakumod's internal arithmetic: `proclaim` incremented its run-counter with
+`$num_of_tests_run = $num_of_tests_run + 1`, and the test script's custom `+`
+was intercepted instead of the native one, making the counter always return the
+custom type rather than an integer.
+
+The root cause was `user_declared_infix_ops`, a flat `HashSet<String>` whose
+presence check fired regardless of which compilation unit was executing.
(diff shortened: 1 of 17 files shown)


tokuhirom:
Superseded by #6119, which was merged to main first and fixes the same underlying issue (block-local `sub infix:<op>` leaking into separately-compiled module code) via a different, non-conflicting mechanism (`module_call_depth`).

This PR's own approach (per-operator declaring-source-file tracking via `executing_cf_source_file`) also had a real regression: it doesn't update `executing_cf_source_file` in the EVAL execution path (`src/runtime/resolution_eval.rs`), so `roast/S03-metaops/hyper.t` test 347 (`sub infix:<+++>` declared and used via hyper-op inside the same EVAL string) fails with "Unsupported reduction operator: +++". Closing rather than fixing forward since the underlying problem is already solved on main.

PR #6120, opened 2026-08-09
fix(runtime): propagate a broken Promise from a whenever nested inside a supply body

tokuhirom:
## Summary
- `whenever <Promise> { ... }` registered from *inside* a `supply { }` block's on-demand body (rather than directly under `react { }`) went through `value_to_react_subscription` / `register_nested_on_demand_source`, neither of which had a `ValueView::Promise` arm — only `build_react_subscriptions` (used for a top-level `whenever <Promise>`, fixed in #6112) did.
- So a broken promise with no `QUIT` handler was silently dropped instead of dying the enclosing react — the subscription marker matched none of the fallback branches and was discarded.
- Adds the same channel+thread Promise handling to `value_to_react_subscription` so both call sites (top-level nested-registration replay and the `SupplyDrivePolicy::Promise` await path) pick it up.

Found via `Cro::Core`'s `tcp.rakutest`, through `Cro::Connector.establish`'s `supply { whenever self.connect(...) -> $t { whenever $t.transformer(...) {...} } }` shape — connecting to a dead port must die the react. Verified against the vendored (not-yet-bundled) Cro test suite locally: `tcp.rakutest` goes from 42/44 to **44/44**, `Cro::Core` is now 9/9 fully green. `Cro::HTTP` re-run shows no regression (still 20/34 fully green files, unrelated to this change).

## Test plan
- [x] New pinning test `t/supply-whenever-nested-broken-promise.t`
- [x] `prove -e target/debug/mutsu` on all `t/react*.t` and `t/supply*.t` (79 files, 371 tests) — all pass
- [x] `cargo fmt -- --check` / `RUSTUP_TOOLCHAIN=1.96.1 cargo clippy -- -D warnings` clean
- [x] CI will run `make test` + `make roast`

### DIFF
--- a/src/runtime/react_died.rs
+++ b/src/runtime/react_died.rs
@@ -137,6 +137,43 @@ impl Interpreter {
                     ..ReactSubscription::new(callback)
                 });
             }
+            return None;
+        }
+        // A `whenever <Promise>` registered from *inside* a nested on-demand
+        // supply body (e.g. `Cro::Connector.establish`'s `supply { whenever
+        // self.connect(...) -> $t { whenever $t.transformer(...) {...} } }`)
+        // lands here rather than in the top-level `build_react_subscriptions`
+        // Promise arm. Without this, a broken promise with no QUIT handler was
+        // silently dropped instead of dying the react (this function returned
+        // `None`, and none of `register_nested_on_demand_source`'s other
+        // fallbacks recognize a bare Promise source either).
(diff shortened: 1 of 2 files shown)


tokuhirom:
Superseded by #6116, which landed the same fix (a different, more architecturally consistent implementation reusing `normalize_promise_whenever_markers`/`arm_pending_promise_whenevers`) concurrently while this PR was in flight. Closing without merging.

PR #6882, opened 2026-08-22
fix: separate isa from role composition

tokuhirom:
Summary:
- Make isa use only nominal class relationships.
- Keep role membership in does and smartmatch checks.
- Add regression coverage for built-in roles and nominal ancestors.

Validation:
- cargo fmt --all
- cargo clippy -- -D warnings
- Targeted TAP: isa-nominal-role-separation, does-mutates-the-object, enum-does-enumeration, can-does, can-method, slip-satisfies-list-return, hyper-reflection-env.
- make test started successfully and completed all 868 Rust unit tests, but the local execution wrapper terminated during the TAP phase before the saved log's final summary; it was not re-run per repository policy.


tokuhirom:
Superseded by #6884 (same fix, duplicate PR created by the same agent session — closing to avoid confusion).

PR #6934, opened 2026-08-23
perf: reuse compiled xx repeat closures

tokuhirom:
## Summary

Avoid runtime compilation for compiler-generated `xx` repeat closures, and inline small scalar constant repeat counts in the enclosing bytecode.

## Validation

- `cargo fmt --all`
- `cargo clippy -- -D warnings`
- `prove -e target/debug/mutsu t/xx-constant-repeat.t t/list-repeat-rand-and-pick.t t/xx-thunk-reeval.t t/xx-reevaluates-lhs.t`
- `make test` was started; the local command runner ended it during the Rust-test phase, so CI will provide the complete suite result.


tokuhirom:
Integrated into #6933.

PR #7120, opened 2026-08-29
fix: copy sprintf format literals as UTF-8, not Latin-1

grondilu:
## Summary

`printf "√2 ≈%.9f\n", $x` wrote mojibake (`â2 â1.414…`, `Ï` for `π`) because `format_sprintf_impl` walked the format string a byte at a time and pushed each byte as a Latin-1 codepoint:

```
√  U+221A  UTF-8 E2 88 9A  →  U+00E2 U+0088 U+009A  →  UTF-8 C3 A2 C2 88 C2 9A
```

A UTF-8 terminal then shows `â` and swallows the C1 continuation bytes. `say "√2 ≈ π"` was already correct — only sprintf/printf format *literals* were wrong; `%s` arguments were fine.

The formatter now copies each run of literal text as a UTF-8 slice up to the next ASCII `%`. Directive letters are consumed as a whole Unicode scalar so the scan stays on a character boundary.

Pinned by `t/sprintf-utf8-literal.t`.

## Attribution

This PR is opened from the `grondilu/mutsu` fork by GitHub user **grondilu**.

The change itself was authored by **Grok** (xAI, `Grok <grok@x.ai>`). I am Grok, acting only on grondilu's behalf — not as tokuhirom or any other account.

## Test plan

- [x] `t/sprintf-utf8-literal.t` covers UTF-8 literals around directives, `%s` arguments, and `%%`

### DIFF
--- a/news/2026-08/sprintf-utf8-literal.md
+++ b/news/2026-08/sprintf-utf8-literal.md
@@ -1,0 +1,20 @@
+# sprintf copies UTF-8 format literals as UTF-8, not Latin-1
+
+`printf "√2 ≈%.9f\n", $x` wrote mojibake (`â2 â1.414…`, `Ï` for `π`) because
+`format_sprintf_impl` walked the format string a byte at a time and pushed each
+byte as a Latin-1 codepoint:
+
+```
+√  U+221A  UTF-8 E2 88 9A  →  U+00E2 U+0088 U+009A  →  UTF-8 C3 A2 C2 88 C2 9A
+```
+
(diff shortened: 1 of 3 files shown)


grondilu:
Actual grondilu here:  this was found when running https://rosettacode.org/wiki/Continued_fraction#Raku

Let me know if you prefer I post issues instead of pull requests.  I've just figured out how to get Grok to post PR himself but I assume it should be possible for him to push issues instead.

tokuhirom:
Thanks for the context. Either issues or pull requests are welcome; for a focused fix with a regression test like this, a PR is perfectly fine.

PR #7234, opened 2026-09-02
fix: stringifying a list calls each element's own Str

tokuhirom:
## What

```raku
class C { has $.t; method Str { $!t } }
my @a = (C.new(t => "hi"),);
say ~@a;            # raku: hi     mutsu: C()   (before)
say @a.join("");    # raku: hi     mutsu: hi
```

`Value::to_string_value` is a *pure* renderer: it knows the list-shape rules (space-separated, nested lists flattened) but cannot dispatch a user-defined method, so an `Instance` element came out as the `ClassName()` fallback. `.join` had always resolved its elements through the interpreter; every other string-context entry point had not, so the two disagreed about the same array.

Rather than duplicate the shape rules in an interpreter-aware renderer, the elements are resolved **in place** — each `Instance` replaced by the string its class dispatches, recursing into nested lists — and the resulting list handed to the same pure renderer (`src/runtime/list_element_stringify.rs`). The shape rules stay in one place; only the per-element stringification moves.

Five entry points needed it, which is why the bug survived:

| entry point | site |
| --- | --- |
| `.Str` / `.Stringy` method | native fast path declines a list holding an `Instance`; `dispatch_list_str_method` handles it |
| prefix `~` | `exec_str_coerce_op` |
| string interpolation | `exec_string_concat_op` |
| infix string operands (`eq`, `~`) | `coerce_stringy_operand` |
| native TAP `is` | `stringify_test_value` + `value_for_diag` |

## Why it matters

Found while re-measuring the `XML` battery candidate (`todo/tickets/bundle-xml-battery.md`). `t/namespaces.rakutest` asserts `is @items[3].contents, 'A nested item, oh boy.'`, and `.contents` is a list of `XML::Text` nodes whose `Str` returns the text — mutsu compared `XML::Text()`. With this and #7233 the upstream suite reaches **15/15**, matching `raku`.

## Tests

- New `t/list-str-calls-element-str.t` (12 subtests) — passes under `raku` too. Covers all five entry points, nested lists, mixed elements, `.gist`/`.raku` staying object-shaped, and a plain list keeping the untouched fast path.
- `make test` green.
- A 417-file roast sweep over `S02-types` / `S03-operators` / `S12*` / `S32-list` / `S32-str` / `S04*` (56171 tests) green — the only failures were three `S32-str/*-encode-decode.t` files that need `make roast`'s `t/spec` symlink for their sample data, unrelated to this change.

## Also filed

`todo/tickets/str-method-falls-back-to-stringy.md` — mutsu's `.Str` dispatch falls back to `Stringy`, so a class defining only `Stringy` answers that where raku answers the `Mu.Str` default. `.join` has always had it too, so the new path deliberately reuses `.join`'s exact call and the two agree; the test pins that agreement rather than a literal.

### DIFF
--- a/news/2026-09/list-str-calls-element-str.md
+++ b/news/2026-09/list-str-calls-element-str.md
@@ -1,0 +1,46 @@
+# Stringifying a list calls each element's own `Str`
+
+```raku
+class C { has $.t; method Str { $!t } }
+my @a = (C.new(t => "hi"),);
+say ~@a;            # raku: hi     mutsu: C()   (before)
+say @a.join("");    # raku: hi     mutsu: hi
+```
+
+`Value::to_string_value` is a *pure* renderer: it knows the list-shape rules
+(space-separated, nested lists flattened) but cannot dispatch a user-defined
+method, so an `Instance` element came out as the `ClassName()` fallback.
+`.join` had always resolved its elements through the interpreter; every other
+string-context entry point had not, so the two disagreed about the same array.
+
+Rather than duplicate the shape rules in an interpreter-aware renderer, the
+elements are resolved **in place** — each `Instance` replaced by the string its
+class dispatches, recursing into nested lists — and the resulting list handed to
+the same pure renderer (`src/runtime/list_element_stringify.rs`). The shape
+rules stay in one place; only the per-element stringification moves.
+
+Five entry points needed it, which is why the bug survived: the `.Str`/
+`.Stringy` method (the native fast path now declines a list holding an
+`Instance` and `dispatch_list_str_method` handles it), prefix `~`
+(`exec_str_coerce_op`), string interpolation (`exec_string_concat_op`), infix
+string operands (`coerce_stringy_operand`, which is how `eq` sees it), and the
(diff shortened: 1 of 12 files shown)


tokuhirom:
Rebased onto `main` (so #7233 is in) and fixed a defect CI caught: `roast/integration/advent2009-day20.t`'s `is @b, (@people.sort: {...})` compares an **Array** on one side and a **Seq** on the other, and only the Array side had been converted — so the two disagreed. `stringify_test_value`'s Seq/LazyList arms now re-enter through the list arm instead of mapping `to_string_value` themselves, and the resolver handles an already-reified `Seq`/`HyperSeq`/`RaceSeq` so prefix `~` on one agrees too. Three subtests added for that shape.

Re-verified locally after the fix: **full `make roast` PASS** (1436 files, 218962 tests), `make test` PASS, `scripts/battery-testsuite.sh` GATE PASSED (274/297).

With #7233 in, the `XML` upstream suite now reaches **15/15** files under mutsu — the same as `raku`.

PR #7244, opened 2026-09-02
fix: a sigilless list-destructuring target binds, it does not copy

tokuhirom:
## Problem

```
raku  -e 'my ($x,$y) = 1,2; my (\a,\b) := ($x,$y); a = 10; say $x'   # 10
mutsu -e 'my ($x,$y) = 1,2; my (\a,\b) := ($x,$y); a = 10; say $x'   # Cannot assign to an immutable value
```

The single-variable form (`my \a := $x`) already aliased, and so did hand-unrolling the list. Only the parenthesised list form was broken.

## Root cause

`parse_positional_destructuring` emitted a **value declaration plus two readonly marks** for a sigilless target — the shape `my \a = 5` uses, where the term IS its value — rather than the `SyntheticBlock([MarkBind, VarDecl, MarkSigilless])` shape `my \a := $x` uses, which leaves writability to the runtime.

The staging `@__destructure_tmp__` was the other half: it was declared with a plain assignment, which deitemizes the RHS list's element cells away, so even a correct target shape would have had nothing to alias.

## Fix

* In binding mode the staging temp is declared with `MarkBind`, so it keeps the element containers the RHS list carries. Targets that read a *value* out of it are unaffected — a `$` target of a `:=` is a read-only copy in Raku too (`my ($a,$b) := ($x,$y); $x = 7` correctly leaves `$a` at 1).
* A sigilless target emits the single-variable bind's block shape, so `OpCode::MarkSigillessBind` settles its mutability at run time.
* `OpCode::IndexAutovivifyLazyTerminal` gains a `sigilless` flag, threaded from the declaration through the compiler, that suppresses the scalar-leaf promotion for an immutable `List` — otherwise `my (\a,\b) := (5,6); a = 10` would look writable (rakudo dies). An element that already IS a container is handed back untouched and stays writable.

The flag is deliberately narrow. Making the rule unconditional is correct in isolation but breaks three consumers that lean on the promotion (a chunked `for @flat -> \a, \b` loop parameter — 1201 subtests of `S32-str/val.t`; `.kv` on a mutable QuantHash; `.kv` on a `Pair` in a closure). Those are recorded, with the prototyped fixes for two of them, in `todo/deep/immutable-list-element-bind-is-writable.md`.

## Result

Every row of the originating ticket's measured divergence table now matches rakudo, including the ones that rule out a narrower fix: the RHS need not be a literal list (`my @z = 1,2; my (\a,\b) := @z; a = 10` → `[10 2]`), mixed sigilless/sigilled targets work, and the alias survives into a closure.

One row did not close and got its own ticket: a list literal built out of array/hash *elements* (`(@a[0], @a[1])`) still stores copies, because `MakeArray` captures a source container only for a named variable — `todo/tickets/list-literal-does-not-capture-element-containers.md`.

## Testing

* New pin: `t/list-destructuring-sigilless-bind.t` (16 tests, green under `raku` too).
* `make test` — PASS
* Full local `make roast` — PASS
* `scripts/battery-testsuite.sh` — GATE PASSED (289/312)


tokuhirom:
Rebased onto `main` after 4016f677c landed the same ticket's *first* suggestion (skip the staging temp when every target is sigilless and the RHS is a literal list of variables/literals).

That fast path bails to the old, broken staging path for the rows the ticket had already measured as the reason to reject it — a mixed `my (\a, $b) := ($x, $y)`, and a non-literal RHS such as `my (\a, \b) := @z`. This PR fixes the staging path itself, so it subsumes the fast path; I removed the fast path rather than keeping a second code path for the same shapes, and merged its test cases into the pin (now 18 tests, green under `raku` too).

Re-ran after the rebase: `make test` PASS, full local `make roast` PASS.

PR #7400, opened 2026-09-06
fix(grammar): run a $*-mentioning regex code block inline, in match order

tokuhirom:
Works `todo/deep/grammar-dynvar-code-block-still-defers.md` end to end.

## The bug

An embedded `{ … }` block that mentioned a `$*`/`@*`/`%*` dynamic variable was
deferred to the post-match bottom-up reduce walk instead of running where the
cursor reached it, so within one rule it ran **after** the rule's plain blocks
and after its subrules' blocks — two blocks of one rule executing in reverse
source order:

```raku
grammar G {
  token TOP { :my $*N = 0; <a> { $*N = 1; say "dyn" } { say "inline" } <b> }
  token a { \w }
  token b { \w }
}
say G.parse('xy').defined;
```

raku: `dyn` / `inline`. mutsu (before): `inline` / `dyn`. Re-measured 2026-09-06
against a fresh build — the ticket's diagnosis held.

## What the two riders actually needed

- **`$*` rule parameters** (`token value($*STOPPER = '"')`) needed *nothing*.
  `install_subrule_dynamic_params` writes the binding into `self.env` around the
  subrule's whole match, so an inline block is inside that window and reads it
  directly. `t/grammar-dynvar-failgoal-ws.t` (51 assertions) passed unchanged
  the moment the blocks went inline, and `CodeBlockContext.dyn_params` is gone.
- **Per-match `:my $*x`** needed one change, at the *reading* end — not the
  four match-success snapshot sites the ticket proposed. The per-match value is
  already recorded: the declarator is a `RegexAtom::VarDecl` that evaluates its
  initializer per match into the capture delta's `regex_vars`, and an inline
  block's writes are harvested into the same carrier.
  `install_fresh_rule_dynvars` was re-deriving the value at reduce time by
  re-running the declaration, discarding every write; it now installs the node's
  recorded value and only falls back to re-evaluating where the carrier never
  reached. `record_rule_dynvars` reads the env back only for those fallback
  keys. The outer-declaration accumulation shape (`:my %*PLAYED = (); <card>+`,
  `G4` in `t/grammar-per-match-dynvar-action.t`) is unaffected — the walk still
  leaves the declaring node's binding installed and the node holds the same
  container.

## The blocker underneath (a separate, pre-existing bug)

Moving the blocks inline immediately failed `t/grammar-per-match-dynvar-action.t`
for a reason unrelated to dynamic variables: **a write an inline block made to
any in-regex `:my`/`:let` lexical was dropped as soon as the block sat inside a
group.** The single-candidate matcher propagated `RegexCaptures::regex_vars` out
of a `Group`; the *plural* matcher — the one `walk_tokens` actually uses — did
not, and neither did `merge_regex_captures` (conjunction branches, `~` goals)
nor the two LTM ε-bypass folds. The write leaked into `self.env`, where the next
block's own stale snapshot promptly overwrote it. So

```raku
grammar G { token part { :my $V = 'decl'; \w [ { $V = 'set' } ] { make $V } } }
```

made `decl` where raku makes `set`, with no dynamic variable in sight. Six
sub-pattern shapes were checked against raku and all six were wrong; all six
pass now.

## Cleanup

`code_block_defers_to_reduce` disappears entirely and the reduce walk becomes
pure `:actions` bookkeeping that returns immediately for any grammar declaring
no dynvar. The dead replay machinery goes with it — `CodeBlockContext`, the
`code_blocks` axis and its `Undo::CodeBlocksLen` trail entry, the
`EAGER_CODE_BLOCKS` buffer and `enable_`/`drain_`/`execute_regex_code_blocks`,
`reduce_run_code_blocks`, `setup_regex_code_block_env`,
`install_ctx_regex_vars`/`restore_ctx_regex_vars`, `block_base_vars`,
`subtree_has_code_blocks`, and the `ACTIVE_DYN_PARAMS` stack. Net −490 lines.

## Verified locally

- `make test`: 3733 files / 38583 tests, `Result: PASS` (run twice — once after
  the behaviour change, once after the deletion).
- Every named pin: `t/grammar-per-match-dynvar-action.t`,
  `t/grammar-reduce-time-dynvar.t`, `t/grammar-dynvar-failgoal-ws.t`,
  `t/grammar-inline-code-block-order.t`, `t/grammar-inline-make-subrule-made.t`,
  `t/ordered-alternation-loser-code-block.t`, plus the `yaml`/`xml` battery
  tests.
- `scripts/battery-testsuite.sh` (run alone) — result appended below once it
  finishes.
- New pin `t/grammar-dynvar-code-block-order.t`, 12 assertions, every one run
  against real `raku` first.

Full roast delegated to CI.


tokuhirom:
Local `scripts/battery-testsuite.sh` (run alone, release build) has finished: **GATE PASSED — every whitelisted bundled-library test still passes**, 289/312 test files (2 excluded), unchanged from the baseline. `XML` is 15/15 and the grammar-driven batteries are all green.

PR #7436, opened 2026-09-06
fix(modules): scope a loaded compunit's private top-level sub to that compunit

tokuhirom:
## The bug

```
# HelperMod.rakumod: sub helper() { say "mod helper" }
$ mutsu -e 'require HelperMod; helper();'
mod helper                      # raku: "Undeclared routine: helper"
```

A loaded module's package-less top-level `sub` that is NOT `is export`ed stayed
permanently callable, bare, from the loading scope. Raku scopes such a routine
lexically to its own compilation unit.

Root cause (verified, not assumed): mutsu registers every package-less top-level
routine under the literal `GLOBAL::<name>` key — the same key the loading
scope's own declaration uses — so nothing distinguishes "exported, should reach
the importer" from "merely declared, private to this compunit". Closes
`todo/deep/module-toplevel-private-sub-leak-cleanup.md`.

## Why the ticket's suggested fix (a post-load sweep) cannot work

The ticket asked for an exhaustive list of ambient `GLOBAL::` installers so a
*delete* sweep could positively exempt them. The audit says the exemption list
is not the real obstacle. mutsu resolves a bare routine name through the flat
registry, so the module's **own** bodies — its exported subs, its classes'
methods, its `sub EXPORT` — reach their private helpers by the very key the
sweep deletes. `import_module`'s existing non-exported sweep already says this
out loud: it is gated on "shadows a core builtin" precisely so that
"non-colliding helpers stay in GLOBAL so the module's own bodies can still call
them".

So this ships a **move**, not a delete. `src/runtime/unit_private_routines.rs`
moves the entry into `Interpreter::unit_private_routines`, keyed by the
declaring compilation unit; `unit_private_routine` hands it back only to code
compiled in that unit (or in an `EVAL` nested inside it). That is the same
lexical-by-compunit scoping `user_declared_infix_ops` already uses for
user-declared operators, riding the `current_unit` symbol the VM already
maintains across every compiled-routine call. No new representation was needed
— the compunit identity was already there.

## Audit: every ambient `GLOBAL::` installer, and how each is handled

Candidate selection is *exact* rather than heuristic, because
`hide_toplevel_global_routines` (added by the sibling fix) already empties the
package-less single-routine namespace before the loaded body runs: whatever
occupies it afterwards was registered by that body. Its predicate
(`is_toplevel_global_routine_key`) already excludes three kinds:

| installer | key shape | handling |
| --- | --- | --- |
| `sub EXPORT` (`apply_module_export`, read+deleted *after* the load) | `GLOBAL::EXPORT` | excluded by the shared predicate |
| multi candidates (additive across compunits by design — `multi trait_mod:<is>`) | `GLOBAL::name/<sig>` | excluded by the shared predicate |
| `package Foo { ... }` routines (real shared stash entries) | `GLOBAL::Foo::bar` | excluded by the shared predicate |
| `is export` routines (`register_exported_sub`) | `GLOBAL::name` + `EXPORT::TAG::name` aliases | skipped: name is in the union of `exported_subs` / `unit_module_exported_subs` / `module_owned_exports` (conservative — can only under-fix) |
| NativeCall prelude helpers (`PRELUDE_SUB_TRAIT`: `nativecast`, `nativesizeof`, `cglobal`, …) spliced into every compunit that uses NativeCall | `GLOBAL::name` | skipped: recorded by name at their registration site (`registration_sub`), the only place the trait is visible (`FunctionDef` does not carry `custom_traits`) |
| a module's `sub MAIN` | `GLOBAL::MAIN` | left to `remove_leaked_main_routines`, which must *delete* it (an un-exported MAIN at the dispatchable key would be auto-dispatched at program end) |
| tag-filtered exports hidden by `import_module` | renamed `GLOBAL::name` → `MOD::name` | untouched; runs after this and its own `resolve_hidden_owned_export` rescue is unaffected |
| `promote_exported_main_to_global` | `GLOBAL::MAIN*` | runs before seclusion and only for exported MAIN |
| `import_single_require_symbol` (`require Foo <&sym>`) | `GLOBAL::name` | runs after the load, on the importer's behalf — deliberately public |

Then I instrumented the seclusion and ran the whole `t/` suite to see which
compunits *actually* have private top-level routines, rather than trusting the
table:

```
Cro.rakumod                        :: validate-producer validate-consumer
Cro/HTTP/Middleware.rakumod        :: wrap-response-logging wrap-request-logging
Cro/HTTP/Router/LinkGenerator      :: signature-to-sub
IO/Socket/Async/SSL.rakumod        :: EC_KEY_set_group get_ecdh X509_get_ext_d2i … (30 `my sub … is native(…)`)
Log/Timeline/Raku/Setup.rakumod    :: setup-async-socket-logging setup-thread-logging …
OpenSSL/Digest.rakumod             :: MD5 hexify SHA512 SHA384 SHA256 SHA224 SHA1
Text/CSV.rakumod                   :: progress alias
t/lib/ExportTestMod.rakumod        :: not-exported-triple
t/lib/TuxicScoped.rakumod          :: tuxic-mul
<t/sub-export.t fixture> ESelect   :: greet-fr greet-en
```

All still work. The `ESelect` one is the trap the ticket warned about: its
`sub EXPORT` runs *after* seclusion and still reaches those two private helpers,
because its body is compiled in the module's unit.

## Two resolution gaps this surfaced (both fixed here)

- **Name-keyed call caches.** `pos_light_call_cache` / `light_call_cache` /
  `otf_call_cache` / `func_multi_resolve_cache` / `fn_resolve_cache` are keyed
  by `(name, package)`, which cannot represent a winner that depends on the unit
  asking. A private helper called once from inside its module was then answered
  from the cache for a later call in the loading scope — the leak coming back in
  through the front door. Such names now bypass those caches, exactly as
  `amp_param_shadowed_names` already does for a name a `&`-sigil parameter can
  shadow. Cost is one `is_empty()` branch for every program that never loads a
  module with a private sub.
- **`current_unit` was not entered for method bodies.**
  `enter_compilation_unit` is called by the compiled-*sub* call paths only, so a
  method of a class the module declares ran under the caller's unit;
  `unit_private_routine` falls back to the innermost routine frame's `def_file`
  (what a backtrace renders) for those. Separately,
  `call_compiled_function_positional_light_at`'s return-type-check failure path
  returned without restoring `current_unit` at all — a pre-existing leak that
  also mis-scoped user-declared operators.

## Pin

`t/module-private-sub-does-not-leak.t` (9 assertions, **every one verified
against Rakudo**, which passes the file unchanged) with
`t/lib/PrivateSubMod.rakumod`: the module's exported sub, a method of a class it
declares, and a block inside a module routine all still reach the private
helper; the loading scope's own same-named routine is not displaced in either
direction; a private helper the loading scope never declared is simply not
there; and a later `require` of the same unit publishes neither.

## Gates

- `make test`: **PASS** — 3758 files, 39164 tests, `cargo test` 939 passed.
- `scripts/battery-testsuite.sh` (run alone): exit 0, no regression against the
  whitelist (276/312).
- Whitelisted `roast/S10-packages/` + `roast/S11-modules/`: 22 files, 336 tests,
  PASS.
- `cargo fmt` + `cargo clippy --all-targets -- -D warnings` clean.

## Deferred, filed not fixed

`Pkg::name(...)` falls back to a bare `GLOBAL::name` even when `Pkg` does not
exist (`mutsu -e 'sub zzz($n){$n+1}; say NoSuchPkg::zzz(1)'` prints `2`; raku
raises `Could not find symbol '&zzz' in 'GLOBAL::NoSuchPkg'`). Fully
pre-existing and orthogonal; it is what stopped the pin from asserting "the
private helper is not reachable as `Mod::name`". Recorded in
`todo/tickets/qualified-call-falls-back-to-bare-global-routine.md`.


tokuhirom:
Closing unmerged: the bundled-battery gate regresses 12 whitelisted files (ten Cro::HTTP, IO::Socket::Async::SSL bad-incoming, zef distribution-depends-parsing), reproduced locally and identically.

The design here is sound and the diff is worth reading rather than re-deriving, so the branch stays and the ticket now points at this PR. The blocker is one specific, measured gap: **a block handed to a native callback taker does not enter its declaring compilation unit when it is finally invoked**. `.tap` is such a taker and every `supply`/`whenever` body is one, which is exactly Cro's `supply whenever wrap-response-logging(...)` failing with `Unknown function: wrap-response-logging`.

The closure creation sites are innocent (`MakeAnonSub`/`MakeAnonSubParams` both stamp `source_file: self.executing_source_file()`) and `call_compiled_closure_with_topic` already enters `unit_of_source(data.source_file)`; the tap callback is not invoked through that path at all. Closing it means entering the callee's unit at the generic code-object invocation, which also moves `user_declared_infix_ops` scoping and so needs its own measurement.

#7445 rewrites `todo/deep/module-toplevel-private-sub-leak-cleanup.md` with the probe table, the two narrower shapes to weigh, the two independent pre-existing bugs this branch fixed on the way (worth salvaging separately), and the 12-file acceptance set. It also salvages `todo/tickets/qualified-call-falls-back-to-bare-global-routine.md` onto main, since it only existed here.

PR #7444, opened 2026-09-07
fix(vm): a slice, `.first` and a block call hand out element containers

tokuhirom:
Closes producers **2, 3 and 4** of section B of
`todo/deep/immutable-lvalues-that-mutsu-still-lets-you-assign-to.md`. That
section is not an immutability family — it is the opposite failure: rakudo
performs these writes and mutsu silently lost them.

| repro | raku | mutsu before | mutsu now |
|---|---|---|---|
| `my @a=1,2,3; for @a[0..1] { $_ = 5 }; @a` | `[5 5 3]` | `[1 2 3]` | `[5 5 3]` |
| `my %h=a=>1,b=>2; for %h<a b> { $_ = 5 }; %h` | `{a=>5,b=>5}` | unchanged | `{a=>5,b=>5}` |
| `my @a=1,2,3; @a[0..1].map({$_=5}).eager; @a` | `[5 5 3]` | `[1 2 3]` | `[5 5 3]` |
| `my @a=1,2,3; @a.first({$_=5}); @a` | `[5 2 3]` | `[1 2 3]` | `[5 2 3]` |
| `my $v=1; my $b={$_=9}; $b($v); $v` | `9` | `1` | `9` |
| `my @a=1,2,3; my $b={$_=9}; $b(@a[0]); @a` | `[9 2 3]` | `[1 2 3]` | `[9 2 3]` |
| `my @a=1,2; class S { method take($y is rw) {$y=9} }; S.new.take(@a[0]); @a` | `[9 2]` | dies | `[9 2]` |
| `my @a=1,2; my $r = &g; $r(@a[0]); @a` (`sub g($y is rw){$y=9}`) | `[9 2]` | dies | `[9 2]` |
| `my @a=1,2; (-> $x is rw {$x=9})(@a[0]); @a` | `[9 2]` | dies | `[9 2]` |

## Root cause: one missing producer, not three bugs

`@a.values` has handed out the array's live `Scalar` cells since ADR-0045
(`Interpreter::array_element_producer`), and everything downstream of a cell
already worked before this change — `@a.values.first({ $_ = 5 })`,
`@a.Seq.first({ $_ = 5 })`, `for @a.values { $_ = 5 }` and
`my $b = { $_ = 9 }; $b(@a.values[0])` all wrote through. So the topic
machinery, the `.first` scan and the block-call binder were already correct;
what was missing was a **producer in front of them**.

Two of the ticket's stated diagnoses were measured wrong:

- a slice needed the containers themselves (raku's `@a[0..1]` *is* the two
  `Scalar`s) — not "the source name and the index list carried to
  `vm_loop_writeback.rs`", which would not have covered `@a[0..1].map(...)`,
  `%h<a b>`, or a slice element passed as an argument;
- `.first` was recorded as "never reaches the topic marking". That is the
  *immutability* question (section A). Section B's `.first` row was the same
  producer gap.

## What shipped

`Interpreter::promotable_array_len` / `array_element_cells` /
`array_element_cell_at` / `hash_element_cell_at` factor the gate
`array_element_producer` already applied into one place, and then:

- **the slice read** (`vm/vm_var_index_ops.rs`) promotes each in-bounds element
  of a mutable Array/Hash;
- **`.first`** (`vm/vm_native_first.rs` — the path that actually serves it —
  plus the adverb-carrying `dispatch_first`) scans the element containers and
  decontainerizes its answer;
- **`Interpreter::pending_call_topic_source`**, the exact sibling of
  `pending_call_topic_bare` with the same one-call lifecycle (set by the two
  value-call opcodes, read before `push_call_frame`, cleared by it), binds a
  bare block's implicit `$_` raw to a plain scalar argument through the same
  shared `ContainerRef` cell recipe `binding_signature.rs` uses for `is rw`;
- **`OpCode::IndexArgRef`**, ADR-0067's subscript-ARGUMENT producer — the twin
  of `IndexInvocantRef` one position over, gated at run time on the real callee.
  This also closes the headline of
  `todo/tickets/subscript-argument-container-producer.md`, exactly as that
  ticket predicted it would.

## Three consumers the promotion surfaced, all fixed here

- `@a[0,1]»++` desugars to "read the OLD slice, mutate in place, discard the
  NEW"; the OLD read is now the same containers the mutation writes through, so
  it gets an explicit `DecontListElems` snapshot.
- A defaulted HOLE (`@a[3]:delete` then `@a[2,3,4]`) is not promoted — the
  default must be read *through* the marker.
- The list-destructuring staging temp is excluded from the gate. The obvious
  alternative (decontainerize in the hash initializer) is blocked by a
  representation ambiguity — mutsu spells "a `$`-sourced itemized hash" and "a
  hash in an element cell" identically — recorded in
  `todo/deep/containerref-holding-a-hash-is-indistinguishable-from-itemization.md`.

## Verification

- `t/slice-hands-out-element-containers.t` (16), `t/first-scans-element-containers.t` (12),
  `t/block-call-binds-topic-raw.t` (18) — every expectation run under real
  `raku` v2026.06 as well as mutsu.
- Local `make test`: `Result: PASS` (3762 files, 39272 tests).
- Full local `make roast`: `Result: PASS` (1436 files, 218962 tests).

## Still open in section B

Producer 1 (`@a.list.map({ $_ = 7 })`) is untouched — its route runs through
`SeqSource::MapGrep`, whose blocker is
`todo/deep/deferred-map-callback-runs-in-the-consuming-frames-env.md` /
ADR-0058 §9.

### DIFF
--- a/news/2026-09/first-and-block-topic-element-containers.md
+++ b/news/2026-09/first-and-block-topic-element-containers.md
@@ -1,0 +1,150 @@
+# `.first`, an associative slice and a block call hand out element containers
+
+Section B of
+`todo/deep/immutable-lvalues-that-mutsu-still-lets-you-assign-to.md` is not an
+immutability family at all — it is the opposite failure: rakudo performs these
+writes and mutsu silently lost them.
+
+```raku
+my %h = a=>1,b=>2; for %h<a b> { $_ = 5 }; %h    # raku {a=>5,b=>5}  mutsu was unchanged
+my @a = 1,2,3; @a.first({ $_ = 5 }); @a          # raku [5 2 3]  mutsu was [1 2 3]
+my $v = 1; my $b = { $_ = 9 }; $b($v); $v        # raku 9        mutsu was 1
+my @a = 1,2,3; my $b = { $_ = 9 }; $b(@a[0]); @a # raku [9 2 3]  mutsu was [1 2 3]
+```
+
+Pinned by `t/first-scans-element-containers.t`,
+`t/block-call-binds-topic-raw.t` and `t/slice-hands-out-element-containers.t`
+(whose five POSITIONAL-slice rows are `todo`-marked — see the last section).
+
+## Two narrowings this needed, both found by the bundled-battery gate
+
+The change as first written regressed four whitelisted battery files, and each
+had a different cause.
+
+**`Log::Timeline` (3 files).** The block-call topic alias fired for *every*
+block call, and the alias is not free: it installs a cell in the CALLER's env
+under the argument's source name and registers an exit writeback. Nested
+`Task.log: { ... }` blocks never touch `$_`, and the writeback re-published an
+outer task's state over the inner one's — `logging.rakutest` reported the OUTER
+task's id for the inner task's end entry. `CompiledCode::writes_topic` (set in
+`compute_free_vars` when the body writes `_` by name) now gates it, so a block
+that cannot write the topic pays nothing and observes nothing.
+
+**`Text::CSV` (1 file), which parked the positional half.** Handing out the
+array's own element containers from a POSITIONAL slice corrupts
+`csv(in => $aoa.iterator, out => $fno)`: every row of `$aoa` becomes
+`IterationEnd`, caught by the whitelisted `t/90_csv.t` test 503 "AOA parse out".
+The associative half (`%h<a b>`) is unaffected and ships. `slice_array_entry`
+is therefore not in this change; the rows that need it are `todo`-marked and
+the producer is recorded in the ticket's section B.
+
+A process note worth keeping: the first bisect of that failure named the wrong
+culprit, because `scripts/battery-testsuite.sh` was running in the background
+while the probes ran. Two gates share one workdir and produce false
+REGRESSIONs, exactly as CLAUDE.md warns. Every number above was re-measured
+with nothing else running, and against a `main` baseline confirming the file
+passes there.
+
+## One gap, not three
+
+All three rows turned out to be the same missing thing, and the ticket's stated
+root cause was wrong for two of them.
+
+`@a.values` has handed out the array's live `Scalar` cells since ADR-0045
+(`Interpreter::array_element_producer`), and everything downstream of a cell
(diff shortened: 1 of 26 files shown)


tokuhirom:
Took this over after the authoring agent's session ended, with the battery gate red (4 whitelisted files: `Text::CSV/90_csv.t` and three `Log::Timeline`). Reworked and force-pushed; every number below was re-measured on a clean run, one producer disabled at a time, against a `main` baseline confirming each file passes there.

**Two narrowings:**

1. **`CompiledCode::writes_topic` gates the block-call topic alias.** The alias installs a cell in the CALLER's env under the argument's source name and registers an exit writeback, so firing it for *every* block call is observable, not merely wasteful: `Log::Timeline`'s nested `Task.log: { ... }` blocks never touch `$_`, and the writeback re-published an outer task's state over the inner one's — `logging.rakutest` reported the OUTER task's id for the inner task's end entry. Fixes all three `Log::Timeline` files.

2. **The POSITIONAL half of the slice producer is parked.** Handing out the array's own element containers from `@a[0..1]` corrupts `Text::CSV`: `csv(in => $aoa.iterator, out => $fno)` leaves every row of `$aoa` as `IterationEnd` (its `t/90_csv.t` test 503, "AOA parse out"). The **associative** half (`%h<a b>`) is unaffected and ships. The five positional rows in `t/slice-hands-out-element-containers.t` are `todo`-marked, and section B of the ticket records what a future producer has to account for.

**A process note worth recording:** my first bisect named the wrong culprit (the subscript-argument producer) because `scripts/battery-testsuite.sh` was still running in the background while the probes ran — two gates share one workdir and produce false REGRESSIONs, exactly as CLAUDE.md warns. The subscript-argument producer is innocent and stays.

**Gates:** `make test` PASS (3768 files, 39447 tests) · `scripts/battery-testsuite.sh` run ALONE **GATE PASSED** (289/312, no regression) · full local `make roast` PASS (1436 files, 218962 tests) · `cargo fmt` + `clippy --all-targets -D warnings` clean · all three pin tests pass under mutsu **and** under real `raku`.

PR #7480, opened 2026-09-07
perf(vm): build a named routine's callframe().code object on demand

tokuhirom:
Ninth perf slice for `todo/deep/vendor-real-test-module.md` (callgrind-driven), following #7472, #7476, #7477.

## What

Every named routine call pushed its own `Sub` value on entry so that `callframe().code`, `&?ROUTINE` and the backtrace could hand it out:

```rust
Value::make_sub(pkg, name, cf.params.clone(), cf.param_defs.clone(), vec[], false, self.clone_env())
```

A `Gc` allocation, two `Vec` deep clones, and — because `clone_env` flattens — a whole-lexical-scope map clone whenever the caller's env was a scoped overlay (every call made from inside another routine), then dropped on return. Almost no call ever reads it. `todo/perf/method-dispatch-flattens-the-env-on-every-call.md` had already named "a lazy `callframe().code` env" as one of the three per-call full-view consumers.

`block_stack` and `CallFrameEntry::code` now hold a `CodeFrame` (`src/runtime/code_frame.rs`):

- `Ready(Value)` — a closure/block body or an interpreter-carrier routine, unchanged;
- `Lazy(Arc<LazyRoutineCode>)` — a named compiled routine: package, name, params, param_defs, and the caller env as an `Arc` bump (later caller writes copy-on-write away from it, so it is the same snapshot the eager flatten took).

`Interpreter::code_frame_value` builds the `Sub` on first read exactly as the entry path used to (full lexical view via `Env::flattened`, routine mixins re-applied via `materialize_routine_mixins_shared`) and caches it in the frame, so repeated reads within one frame see one object. Every reader goes through it (`callframe(N).code`, `&?ROUTINE`, `&?BLOCK`, `CallFrame.code` attributes); the backtrace's routine search compares `(package, name)` without materializing. The cycle collector roots a lazy frame's env values and its built object.

## Measured

Callgrind, 300 `ok 1, "x"` in a loop under `MUTSU_REAL_TEST=1`, one-assertion baseline subtracted, release build:

| | before | after |
| --- | --- | --- |
| per assertion | 276,814 Ir | 261,339 Ir |
| `Env::flattened` | 8.2k | 0 |
| `drop_in_place<Gc<SubData>>` | 7.9k | 0 |
| `proclaim`'s frame | 166.4k | 144.1k |

**-5.6% per assertion**; -22.2% cumulative since the session opened at 335,929.

## Verified

- `prove t/` on the debug binary: 3782 files, all pass.
- 40 whitelisted roast files that read call frames / `&?ROUTINE` / backtraces / wrap chains (release, via `scripts/run-roast-test.sh`): all pass.
- 150 callframe/closure/gather `t/` files plus the assertion loop under `MUTSU_GC=on MUTSU_GC_EVERY_CANDIDATE=64 MUTSU_GC_VERIFY=1`: all pass.
- `cargo clippy --all-targets -- -D warnings`, `cargo fmt`.

https://claude.ai/code/session_01J314VsqdhJ15T1rUJLvmyH

---
_Generated by [Claude Code](https://claude.ai/code/session_01J314VsqdhJ15T1rUJLvmyH)_

### DIFF
--- a/news/2026-09/callframe-code-object-is-built-on-demand.md
+++ b/news/2026-09/callframe-code-object-is-built-on-demand.md
@@ -1,0 +1,53 @@
+# A named routine's `callframe().code` object is built on demand
+
+Ninth perf slice for `todo/deep/vendor-real-test-module.md`. Every named
+routine call pushed its own `Sub` value on entry so that `callframe().code`,
+`&?ROUTINE` and the backtrace could hand it out:
+
+```rust
+Value::make_sub(pkg, name, cf.params.clone(), cf.param_defs.clone(), vec![],
+                false, self.clone_env())
+```
+
+That is a `Gc` allocation, two `Vec` deep clones, and -- because `clone_env`
+flattens -- a whole-lexical-scope map clone whenever the caller's env was a
+scoped overlay, which it is for every call made from inside another routine.
+Then the object is dropped on return. Almost no call ever reads it. On the
+vendored `Test.rakumod`'s assertion loop the two frames (`ok`, `proclaim`)
+cost ~16k instructions per assertion in `Env::flattened` and
+`drop_in_place<Gc<SubData>>` alone, and
+`todo/perf/method-dispatch-flattens-the-env-on-every-call.md` had already named
+"a lazy `callframe().code` env" as one of the three per-call full-view
+consumers standing between the interpreter and a cheaper scoped-env story.
+
+`block_stack` and `CallFrameEntry::code` now hold a `CodeFrame`:
+`Ready(Value)` for a closure/block body or an interpreter-carrier routine
+(unchanged), or `Lazy(Arc<LazyRoutineCode>)` for a named compiled routine.
+The lazy frame records the ingredients -- package, name, params, param_defs,
+and the caller's env as an `Arc` bump (later caller writes copy-on-write away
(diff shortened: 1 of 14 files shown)


tokuhirom:
CI `test` job failed on `t/supply-on-demand-closing-tap.t` test 3 (`closing fires for an async (start-block) on-demand supply tapped with .tap` — the 10s poll for `$closed` expired). Standing down on it as not this PR's failure, and re-running the failed job once:

- The same commit (56340c5) passed the full `prove t/` in the `gc-stress` and `jit-stress` jobs, which run the identical file.
- This PR's only behavioural change is the lazy `callframe().code` object; a `rust-gdb` breakpoint on `push_lazy_block` never fires while running this test file (no named compiled routine is called in it), so the diff is not on the failing path.
- Locally on this head: 10/10 debug, 10/10 release, 40/40 with 8 concurrent copies, and `scripts/flake-repro.sh -n 40 -l 4` → `pass=40 fail=0`.
- `docs/flaky-test-policy.md` §6 already names this file as one that cost re-runs in the 2026-07-23 CI survey; it is not in `flaky-tests.txt`. That is a pre-existing test-3 race (the `start`-block `done` → `closing` path) worth its own quarantine entry or root cause, separate from this PR.

---

PR #7523, opened 2026-09-07
fix: six interpreter bugs the vendored Test module exposed (the default flip is withdrawn)

tokuhirom:
Six general interpreter bugs, found by running Raku's own 953-line `Test.rakumod` verbatim under mutsu. **The provider flip this PR originally made is withdrawn** — the fixes are independent of it and land on their own; `use Test` still resolves to the native TAP provider.

## Why the flip was withdrawn

The vendored module runs verbatim, and both suites go green under it: `prove -j4 t/` 3836 files / 40528 tests, `make roast` 1436 files / 218939 tests. The `Bundled-library test suites` gate does not: **282/312, with four upstream distributions regressing** on four unrelated interpreter gaps that only the real module's code shapes reach.

None of the four is a `Test` compatibility problem, and none had a fix small enough to ride along. The `NativeLibs` one was root-caused to a genuine one-line defect, fixed, measured, and **reverted**: making the custom `sub EXPORT` actually run exposes a missing `&trait_mod:<is>.candidates` / `.dispatcher`, which turns a quiet wrong answer into a hard error and takes DBIish's `install-driver` down with it (`t/prelude-helper-not-block-lexical.t` goes red).

Landing the flip anyway would mean re-baselining that gate's whitelist, which hides exactly the compatibility regression the gate exists to catch. So the default goes back to opt-in (`MUTSU_REAL_TEST=1`) and the fixes land without it.

Under the shipped default the gate is green: **284/312 locally, the six remaining rows all `DBIish` MySQL tests that need a live server this container has not got**. The four real regressions are gone.

## The six fixes

Each was a pre-existing `main` bug that the module merely exposed — the pattern the whole campaign followed.

| | |
| --- | --- |
| `fix(compiler)` | a slice subscript argument is not an `is rw` writeback target |
| `fix(grep)` | a deferred grep's capture merge stops clobbering its consuming frame |
| `fix(vm)` | a user-defined tied container's `STORE` survives the capture cell |
| `fix` | `~&foo` on a named regex yields its name; one `list $s` stops disarming a Seq's single-use gate |
| `fix(module)` | a module first loaded inside an `EVAL` keeps its imports |
| `fix(module)` | …and reinstates its `GLOBAL::` aliases only on the EVAL rollback |

### `fix(compiler)`: slice subscript arguments

`compile_call_arg_with_escape` queued an `is rw` snapshot/writeback pair for *every* `Expr::Index` argument, not distinguishing an element from a slice. rakudo rejects the slice bind outright:

```raku
sub f($x is rw) { $x = 9 }
my @a = 1, 2, 3;
f(@a[1..*]);       # X::Parameter::RW: expects a writable container (variable)
say @a;            # (1 2 3) -- untouched
```

mutsu bound it and wrote the callee's scalar back over the slice, leaving `@a` as `(1 9)`; a hash slice set *both* keys. The same queue also fired where it had nothing to write — its guards are computed over compile-time-fixed global temp names, so a recursive descent re-entering the same call site clobbers them. That is `roast/integration/99problems-21-to-30.t`'s P26 `group`. A statically-visible slice index now compiles as the plain value read it is; dropping the `WrapVarRef` is what lets the binder raise `X::Parameter::RW` on its own. Pinned by `t/rw-slice-arg-is-not-a-container.t`.

### `fix(grep)`: the deferred capture merge

ADR-0058 made `.grep` produce a deferred Seq, so its callback runs at the pull. The captured-env merge was still written for the old timing — it overwrote every captured name unconditionally while saving only names the consuming frame did not already hold, so a name held by both was replaced with its capture-time value and never put back:

```raku
my $s = (1, 2, 3, 4).grep({ $_ %% 2 });   # `s` is captured here, as Any
say $s.elems;                             # 2   -- the pull runs here
say $s.^name;                             # mutsu: Any     rakudo: Seq
```

`.first`'s batched matcher carried the identical shape. Pinned by `t/grep-capture-merge-keeps-caller-lexicals.t`.

### `fix`: `~&foo` on a named regex, and the `list` listop

`regex_str_coercion` resumed rakudo's "Regex in string context" warning with the empty string for *every* regex; rakudo resumes it with the regex's `.name`. `&foo.Str` already answered `"foo"`, so the two coercions disagreed with each other as well as with rakudo — and the empty string is never safe, because it makes every named regex compare `eq` to every other one.

Separately, `list $s` on a single `Seq` is rakudo's no-op, but mutsu reified it on the way in, and `SeqBody::reify` is the only thing that marks a body `retained` — which is exactly what exempts it from the consuming path that raises `X::Seq::Consumed`. Located with `rust-gdb -batch`; the backtrace named `try_native_function -> reify_map_grep_seq_args` outright.

**Worth recording:** `roast/S03-operators/context-forcers.t`'s "list listop doesn't cache" had been passing for the *wrong* reason — the grep fix above removed a spurious death that had been satisfying its `dies-ok`.

### `fix(module)`: a module first loaded inside an `EVAL`

Such a module lost the routines it and its dependencies had imported, while `loaded_modules` kept it recorded as loaded, so the later genuine `use` short-circuited and could never put them back. `Test`'s `use-ok` is `EVAL ( "use $code" )`, so every `use-ok 'M'; use M;` pair hit it. Measured against rakudo: imports are lexical to whichever scope ran the `use`, while the module and its own state persist process-wide.

An earlier attempt — rolling `loaded_modules` back alongside the registry — was measured and rejected: it fixed the repro but moved the failure to `No such method 'realstart'`, because it made the module *reload* when rakudo says it should never have been unloaded.

### `fix(module)`: …only on the EVAL rollback

The commit above dropped a `GLOBAL::` exclusion, and that **regressed `roast/S11-modules/require.t` test 10** (`&bar didn't leak to outer scope`) — caught by CI. `reinstate_module_functions` has two callers, the EVAL rollback and the generic block-scope registry restore, and the change widened both.

The reasoning in the first commit was wrong for a file with **no `unit module` declaration**: `NoModule.rakumod` is `sub bar is export`, whose body runs at `current_package() == GLOBAL`, so its export registers `GLOBAL::bar` — indistinguishable by key from an importing-scope alias, which the "delta is taken before `import_module`" argument had assumed it could separate. Now gated on the EVAL path only, via `is_eval_unit` in `eval_block_value_inner`, which is where an EVAL's rollback actually runs.

## What is recorded rather than fixed

The four gate root causes, reduced as far as they got, are in **`todo/deep/vendored-test-battery-gate-regressions.md`** — including, deliberately, the repros that were measured and do **not** reproduce, so the next attempt does not re-walk them:

- **`NativeLibs`** — root-caused. `apply_module_export` resolves only `GLOBAL::EXPORT`, so a `unit module Foo;` file's custom export sub (registered as `Foo::EXPORT`) is **never called at all**. Two mutsu divergences were cancelling out, and the test had been passing under the native provider for the wrong reason. The fix and why it must not land alone are both recorded.
- **`Digest`** — narrowed from "wrong HMAC bytes" to something much smaller. Dumping every hash call's input/output and checking against a reference: all eight sha224/256/384 calls match exactly, only the two `sha512` calls do not, **with byte-for-byte correct inputs**, and those same bytes hash correctly standalone. HMAC itself, the key padding, the `Z[+^]` zip, the `reduce` accumulator and `samewith`'s return path were each instrumented and cleared.
- **`Cro::HTTP` ×2** — not yet reduced.
- The five `DBIish` rows need live database servers.

The flip itself is `todo/deep/vendor-real-test-module-flip.md`, which keeps the measurements that were expensive to obtain: the dual-provider sweep results, the closed timeout class (`roast/S03-buf/write-int.t`, ~93 000 assertions, 14.6 s under `prove -j4` against a 30 s budget — a 51% margin), the suite cost, and the three harness details a future flip must settle. `retire-the-native-test-provider.md` is marked blocked on it, and PLAN.md's "landed 2026-09-07" claim is corrected — perf is no longer what blocks the flip.

## Verification

Under the shipped (native) default, release build:

- `prove -j4 t/` — 3836 files, 40528 tests, **PASS**
- `make roast` — 1436 files, 218939 tests; two failures, both container-environment (`6.c/S32-io/file-tests.t`, `S16-filehandles/filetest.t` — filesystem permissions under root), whitelisted and green on runners
- `Bundled-library test suites` — 284/312, the six remaining rows all `DBIish` MySQL tests needing a live server
- `cargo clippy --all-targets -- -D warnings`, `cargo fmt` — clean; CI's `lint-configs` green

https://claude.ai/code/session_01LJTuQKn62x5rChpomLtB7T


tokuhirom:
## CI status: one blocker, bisected — not a flake, and not fixable inside this PR

`test`, `gc-stress` and `jit-stress` all fail on the **same two roast files**, so this is one blocker, not three. Both reproduce 100% locally; a re-run is not indicated. The `t/` suite is green (3825 files) since the tied-container fix in this branch.

### `roast/S03-operators/context-forcers.t` #100 — a bug this PR *uncovers*, not one it causes

Bisected by reverting my own commits one at a time: the `fix(grep)` commit in this branch is what flips it. But the fix is not wrong — it removes the accident that made the test pass.

The test is `my \seq = $l.grep(* > 0); dies-ok { sink (list seq)».abs, (list seq)».abs }`: consuming a `Seq` twice must die. On `main`, grep's captured-env merge overwrote `seq` and never restored it, so the *consumed* Seq stayed in the env and the second read died — correct outcome, wrong reason. Restoring `seq` correctly (which is exactly the clobber this PR fixes) writes back a pre-pull snapshot, and that **resurrects the consumed Seq**.

The real defect is underneath, and involves neither `grep` nor `Test`:

```raku
my \seq = (1..5).grep(* > 0);
my $snapshot = seq;            # a second reference to the same Seq
sink (list seq)».abs;          # consume through `seq`
sink (list $snapshot)».abs;    # rakudo: dies    mutsu: does not
```

**A `Seq`'s consumption state is not shared between references to it.** That is ADR-0034 territory (`Seq reification is in-place and distinct from consumption`) and a materially larger change than this PR — attempting it here would be widening the PR, so I have not.

I tried one narrow guard (skip a restore that writes back the *same* Seq body). It did not help: the env entry is not the same `Arc`, which is the same identity gap stated another way. Reverted; the branch is at the pushed state.

### `roast/S02-magicals/sub.t` #4-6 — pre-existing on `main`, exposed by the provider switch

Not this branch's code at all. A release build of **plain `origin/main`** fails these three under `MUTSU_REAL_TEST=1` and passes them under the native provider. The diagnostic is `Regex object coerced to string (please use .gist or .raku to do that)` raised inside `Test.rakumod`'s `is`, with `expected:` and `got:` rendering identically — i.e. `&?ROUTINE` inside a regex/token/rule compares unequal while stringifying the same.

This is the same shape as the two defects this campaign already fixed here (the `is rw` slice writeback and the tied-container `STORE`): a real gap that mutsu's native `Test` provider was papering over. It needs its own root-cause, which I have not done yet.

### Where this leaves the PR

Both remaining failures are interpreter bugs that the vendored `Test` reveals rather than creates — which is the whole point of the switch, and also why the switch cannot land before they are fixed. The sensible order is to fix the `Seq` identity gap (and `S02-magicals/sub.t`) as their own changes off `main`, then bring this branch green. I have put that choice to the repository owner rather than starting a deep refactor unilaterally, and I am keeping this PR watched until it is green, merged or closed.

---

tokuhirom:
## Next blocker on `f9b9149`: the bundled-library gate, not roast

Both roast blockers are gone on this head — I verified `S02-magicals/sub.t` and `S03-operators/context-forcers.t` pass under **both** providers on a local release build of `f9b9149`, and CI's own roast step reports `Files=1436, Result: PASS`.

The `test` job now fails in a later step, **Bundled-library test suites**:

```
=== summary: 281/312 test files pass (2 excluded) ===
  REGRESSION: whitelisted 'Cro::HTTP  http-middleware.rakutest' did not pass this run
  REGRESSION: whitelisted 'DBIish     25-mysql-common.rakutest' did not pass this run
  ... (10 rows: Cro::HTTP x2, DBIish x6, Digest, NativeHelpers::Blob, NativeLibs)
```

Reproduces locally with `MUTSU_BIN=target/release/mutsu scripts/battery-testsuite.sh`.

### Root cause: `use-ok` EVALs the `use`, and a later real `use` cannot restore the nested imports

`NativeHelpers::Blob`'s `t/01-basic.t` is the smallest case. It dies at test 4 with

```
Unknown function: nativecast
  in sub BODY_OF at lib/MoarVM/Guts/REPRs.pm6 line 64
  in sub carray-from-blob at lib/NativeHelpers/Blob.pm6 line 69
```

and it is **provider-dependent** in the file (24/24 under `MUTSU_REAL_TEST=0`, 3/24 under the vendored module) for one reason: line 7 is `use-ok 'NativeHelpers::Blob'`, and the vendored `Test`'s `use-ok` EVALs the `use`. Replacing just that one line with `pass` makes the file pass **24/24 under the vendored provider**.

The bug underneath is **not** provider-specific. Minimal repro, no `Test` anywhere:

```raku
# tmp/nclib/Inner.rakumod
unit module Inner;
use NativeCall;
sub inner-probe() is export { defined(&nativecast) ?? 'visible' !! 'MISSING' }

# tmp/nclib/Outer.rakumod
unit module Outer;
use NativeCall;
use Inner;
sub outer-probe() is export { inner-probe() }

# script
use lib 'tmp/nclib';
use MONKEY-SEE-NO-EVAL;
my $r = EVAL 'use Outer; 1';   # first load happens inside the EVAL's import scope
use Outer;                     # no-op: already in loaded_modules
say outer-probe();             # rakudo: visible    mutsu: MISSING  (both providers)
```

A module whose **first** load happens inside a scope that later pops (an `EVAL`, and `use-ok` is exactly that) is recorded in `loaded_modules`, so the subsequent genuine `use` short-circuits — while `pop_import_scope` has already removed the bare/`GLOBAL::` aliases that the nested module's own `use NativeCall` registered. Nothing puts them back, and the nested module can no longer resolve `&nativecast`.

I tried the obvious narrow fix — re-registering the native provider's export stash in the already-loaded branch of `load_module` — and it does **not** fix the symptom: those aliases are re-registered for the *importer*'s scope, whereas what was lost is the nested module's own. Reverted rather than pushed; the branch is untouched by me. The real fix looks like it belongs in `pop_import_scope`, which already carves out "a module's own fully-qualified source definitions" and would need an analogous carve-out for the imports belonging to a *module* loaded inside the popped scope, rather than to the scope itself.

### Where things stand

- Both original roast blockers: **fixed** on this head (verified locally, both providers).
- New blocker: the batteries gate, root-caused above, with a `Test`-free repro that rakudo disagrees with.
- The `1 mutsu process died of a fatal signal` line in the same job is the pre-existing deliberate `NativeCall strdup(0)` crash; the allowlist step reports it as expected and it is not what failed the job.

---

tokuhirom:
## A third general bug the switch exposed: a module first loaded inside an `EVAL` lost its imports

This is what the `Bundled-library test suites` gate was red on, and it is fixed here (`main` merged in, then the fix cherry-picked on top).

A module whose **first** load happened inside an `EVAL` lost the routines it and its transitive dependencies had imported, while `loaded_modules` kept it recorded as loaded — so the subsequent genuine `use` short-circuited and could not put them back, and the module could never resolve them again.

```raku
# Inner.rakumod:  unit module Inner; use NativeCall;
#                 sub inner-probe() is export { defined(&nativecast) ?? 'visible' !! 'MISSING' }
# Outer.rakumod:  unit module Outer; use NativeCall; use Inner;
#                 sub outer-probe() is export { inner-probe() }

use MONKEY-SEE-NO-EVAL;
EVAL 'use Outer; 1';
use Outer;
say outer-probe();     # rakudo: visible    mutsu, before: MISSING
```

What went missing is the **nested** module's own view of `&nativecast`, not the importing script's.

### Why it only shows up under the vendored provider

`Test`'s `use-ok` is `EVAL ( "use $code" )`, so every file shaped like

```raku
use-ok 'Some::Module';
use Some::Module;
```

hit it — which is exactly the shape the bundled-library suites use. mutsu's native provider does not implement `use-ok` by EVALing, so only the real module reached the path. Same pattern as the other two fixes: a pre-existing `main` bug, exposed rather than caused.

### Root cause

`eval_eval_string` snapshots the routine registry and restores it afterwards. `reinstate_module_functions` already exists to exempt a module's own routines from that rollback, keyed off `module_registered_functions` — but that set was collected with `!ks.starts_with("GLOBAL::")`, and a `unit module`'s body runs at `current_package() == GLOBAL`, so `Inner`'s own `use NativeCall` registered `GLOBAL::nativecast` and was excluded. `loaded_modules` is deliberately never rolled back, so the two halves disagreed: loaded, with its imports gone.

### What rakudo actually does, measured

Measured rather than guessed. After `EVAL 'use Outer; 1'`:

| | rakudo |
| --- | --- |
| `EVAL 'outer-probe()'` (a fresh EVAL) | `X::Undeclared::Symbols` |
| `use Outer; outer-probe()` (the outer scope) | `visible` |
| `EVAL 'use Outer; outer-probe()'` (same EVAL) | `visible` |

So rakudo scopes the **imports** to whichever scope ran the `use`, while the **module and its own state persist process-wide** — which is what lets a later `use` elsewhere re-import and work.

### The fix

The `GLOBAL::` exclusion is dropped from the `module_registered_functions` delta. The delta is taken *before* `import_module`, and that timing already separates the two kinds of alias along exactly rakudo's line:

- an alias installed while the module's **own body** ran (its `use NativeCall`, its `use Inner`) is in the delta, is lexical to *that module*, which stays loaded, and now survives whatever scope the load sat inside;
- an alias `import_module` installs for the **importing scope** is added after the delta and stays excluded, so it remains lexical to that scope: `{ use Foo } EVAL('foo()')` still dies. `roast/S11-modules/lexical.t` passes unchanged.

An earlier attempt — rolling `loaded_modules` back alongside the registry — was measured and **rejected**: it fixed the repro but only moved the bundled-library failure on to `No such method 'realstart' for invocant of type 'Any'`, because it made the module *reload* when rakudo says it should never have been unloaded. Recorded in the news entry as a measured dead end.

### Measured effect

- `Bundled-library test suites` gate, locally: **276/312 with 15 regressions → 284/312 with 6**. Every remaining one is a `DBIish` MySQL test that needs a live server and fails identically without this change.
- The smallest victim, `NativeHelpers::Blob t/01-basic.t`, died at test 4 with `Unknown function: nativecast  in sub BODY_OF at lib/MoarVM/Guts/REPRs.pm6 line 64`; it now passes **24/24 under both providers**.
- `roast/S11-modules/lexical.t` unchanged.

Pinned by `t/module-loaded-in-eval-keeps-imports.t` (3 tests, measured against `raku` first).

### Still open, deliberately

`{ use Outer; } use Outer; outer-probe()` still answers `MISSING` where rakudo answers `visible`. That is a second, independent mechanism — `pop_import_scope`'s own `registry.functions` retain applies the identical `!starts_with("GLOBAL::")` rule at a block's exit and needs the same module-owned carve-out. Nothing in roast or the bundled libraries depends on it today, so it is filed as `todo/tickets/block-scoped-use-drops-a-nested-modules-imports.md` rather than widening this PR.

---

tokuhirom:
## Status: roast and TAP are green; the `Bundled-library test suites` gate is the one thing left

### What was fixed this round

The `test` / `gc-stress` / `jit-stress` jobs were red on a single root cause — `roast/S11-modules/require.t` test 10, `&bar didn't leak to outer scope`, **a regression from the EVAL/import fix in the comment above**. `reinstate_module_functions` has two callers, the EVAL rollback and the generic block-scope registry restore, and dropping the `GLOBAL::` exclusion widened both.

The claim I made for that fix — that taking the delta before `import_module` separates a module's own aliases from the importing scope's — is **wrong for a file with no `unit module` declaration**. `NoModule.rakumod` is `sub bar is export`, whose body runs at `current_package() == GLOBAL`, so its export registers `GLOBAL::bar`, indistinguishable by key from an importing-scope alias.

Fixed in `bfce0eef8` by gating the `GLOBAL::` half on the EVAL path only (`is_eval_unit` in `eval_block_value_inner`, which is where an EVAL's rollback actually runs). Verified both directions: `require.t` and `lexical.t` pass, and `t/module-loaded-in-eval-keeps-imports.t` plus `NativeHelpers::Blob t/01-basic.t` 24/24 still hold.

Local full runs on that head: `prove -j4 t/` **3836 files / 40528 tests PASS**; `make roast` **1436 files / 218939 tests** with only three container-environment failures (`6.c/S32-io/file-tests.t`, `S16-filehandles/filetest.t` — filesystem permissions under root — and `S32-io/IO-Socket-Async.t`, exit 124 / `Failed: 0`), all whitelisted and green on runners. CI agrees: roast no longer fails.

### What is still red, and why I am not fixing it in this round

The gate reports **282/312, 9 regressed files**. These are provider-caused and therefore this PR's — established with the correct control, `MUTSU_REAL_TEST=0` vs default. (Reverting a single fix commit is the *wrong* control; it gave me a false "not caused by this change" reading earlier in this PR, which I retract.)

They are not one bug. Full analysis, including which repros were measured and do **not** reproduce, is in `todo/deep/vendored-test-battery-gate-regressions.md` on this branch.

**`NativeLibs t/01-basic.t` is root-caused.** Two mutsu divergences were cancelling out, with neither `Test` nor `EVAL` needed to show either:

| module shape | rakudo | mutsu |
| --- | --- | --- |
| custom `sub EXPORT` returning a type object | `visible` | `MISSING` |
| `use X` above the `unit module` line | `MISSING` | `visible` |

Test 6 passed under the native provider **for the wrong reason** — the package leak supplied the answer rakudo gets from `NativeLibs`'s `sub EXPORT(|) { Map.new('NativeCall' => NativeCall, ...) }`. `use-ok`'s EVAL rolls the leak back and the missing half surfaces.

The defect: `apply_module_export` resolves only `GLOBAL::EXPORT`, but a `unit module Foo;` file registers `Foo::EXPORT`, so **the custom export is never called for that entire family**. Proven with a `note` inside the sub:

| `sub EXPORT` placement | invoked? |
| --- | --- |
| file with no `unit module` | yes — the only working shape |
| above the `unit module` line | **no** |
| below the `unit module` line | **no** |

Also consulting the qualified name makes all three run and match rakudo. **I wrote that fix, built it, and reverted it deliberately**: it does not make the file green, it moves the failure to test 1 (`No such method 'dispatcher'` — `&trait_mod:<is>.candidates.first(...)` finds nothing), and it **regresses `t/prelude-helper-not-block-lexical.t`**, breaking DBIish `install-driver` for both SQLite and mysql, because running that EXPORT converts a silent no-op into a hard error mid-import. It is gated on multi-dispatch introspection (`.candidates` / `.signature` / `.dispatcher`) landing first.

`Digest t/rfc4231.t` (3 failures, **wrong computed HMAC bytes** — same family as the deferred-grep capture-merge bug, only observable once `reflective_name_access_possible()` has latched) and the two `Cro::HTTP` files are not yet reduced. The five `DBIish` rows need live database servers and cannot be judged from a container without them.

### What I have not done, on purpose

I have not re-baselined the gate. Lowering the whitelist would hide exactly the compatibility regression the gate exists to catch, and BATTERIES.md's case for rung 2 is that the real module must run verbatim. If that trade-off is ever worth making it is a maintainer call, not a step to take while chasing a green check.

---

Issue #7635, opened 2026-09-08
`do { let $x = 2; Nil }` never rolls the speculative value back


tokuhirom:
## Repro

```raku
my $x = 1;
do { let $x = 2; Nil };
say $x;
```

| | output |
| --- | --- |
| `raku` | `1` |
| `mutsu` | `2` |

The statement-position spelling is correct:

```raku
my $y = 1;
{ let $y = 2; Nil };
say $y;   # raku: 1   mutsu: 1
```

## Root cause

`Stmt::Block` has a `has_let_deep` branch emitting `OpCode::LetBlock`, which resolves the
`let` saves at block exit. `compile_do_block_expr` (`src/compiler/helpers_do_expr.rs`) has
no such arm, so a value-position block never resolves its own saves and the speculative
value survives.

## Why the obvious fix does not work

Adding the arm to `compile_do_block_expr` was tried while resolving #7569 and had to be
backed out. **`Expr::DoBlock` is not a Raku block.** The parser and about a dozen compiler
desugars use it as a generic "run these statements, yield a value" node:

- item context — `$( let $a = 23; $a )` (`src/parser/stmt/simple_expr_stmt/core.rs`)
- the chained-comparison desugar (`src/chain_compare.rs`)
- `cas`, compound-assignment lowering, method-body wrappers
  (`src/compiler/expr_call.rs`, `src/compiler/helpers_ast_utils.rs`)

None of those introduce a scope at which a `let` may resolve. Emitting a `LetBlock` for
every `Expr::DoBlock` made the save resolve at the innermost wrapper instead of the
enclosing block, breaking the idiom roast uses throughout:

```raku
my $a = 42;
{
  is($(let $a = 23; $a), 23, "let() changed the variable");
  Mu;
}
is $a, 42, "let() should restore the variable, as our block failed";
```

That regressed `roast/S04-blocks-and-statements/let.t` (3 subtests) and `temp.t` (1), both
whitelisted. `t/block-shape-statement-value-parity.t` now pins the `$( ... )` direction so
a future attempt fails loudly rather than silently.

## What a real fix needs

A way to tell a **genuine source block** (`do { ... }`, a routine's tail `{ ... }`) from a
synthesized wrapper, carried on the AST node itself — `Expr::DoBlock` would need a flag set
by the parser only at the sites that parse real source braces, with every desugar leaving it
false. `ast.rs`'s own placeholder-classification comments (around the `NoSignature` /
[...]

PR #7637, opened 2026-09-08
refactor(compiler): share one shape classifier between the two bare-block compile passes

### LINKED ISSUE #7569: Bare-block compilation is still split between `Stmt::Block` and `compile_do_block_expr`
## Background

This is the residue of
`todo/deep/unify-statement-expression-control-construct-compilation.md`, which
has otherwise been resolved (see
`news/2026-09/unify-statement-expression-control-construct-compilation.md`).
`for` and `if`/`elsif` chains now have one lowering each, shared by both source
positions (`src/compiler/control_for.rs`, `src/compiler/control_if.rs`), and
`while` / C-style `loop` / `lazy for` in expression position were already thin
`gather`-desugaring wrappers over the statement path.

The bare `{ ... }` block is the one construct still compiled by two independent
passes.

## Root cause

- **Statement position**: the `Stmt::Block` arm of `Compiler::compile_stmt`
  (`src/compiler/stmt.rs`).
- **Expression position**: `compile_do_block_expr` and
  `compile_do_block_expr_scoped` (`src/compiler/helpers_do_expr.rs`).

Both decide the same things — whether the block declares routines, whether it
imports (`PushImportScope`/`PopImportScope`), whether it needs a per-execution
`ResetStateLocals`, whether a `CATCH`/`CONTROL` in it makes it an implicit `try`,
whether ENTER/LEAVE/KEEP/UNDO phasers make it a real phaser block scope, and how
placeholder variables declared inside it are attributed — and each writes its own
answer. They then emit *different* opcodes for the scope itself (`BlockScope` vs
`DoBlockExpr`), which is why this did not fall out of the `for`/`if`
unification: those two shared a single opcode and a single skeleton, and the
block pair shares neither.

## Why it is large

Unlike `for` and `if`, this is not "the same code with a `collect` flag". A
correct unification has to decide first whether `BlockScope` and `DoBlockExpr`
should stay two opcodes at all, and if they should, what exactly each guarantees
about the value stack, the env, the import scope and the phaser queues — then
express the statement form as the value form with the result discarded (or the
other way round). Both arms are load-bearing for a very large amount of
whitelisted roast code, and the placeholder-attribution rules in particular
(ADR-0048 D3/D6, `is_construct_body_block`, `synthetic_block_body`) are subtle
and were arrived at incrementally.

An ADR is probably warranted before any code lands: the opcode question is
exactly the kind of costly-to-reverse decision `docs/adr/README.md` asks to be
recorded.

## Evidence

```
grep -n "fn compile_do_block_expr" src/compiler/helpers_do_expr.rs
grep -n "Stmt::Block(stmts) =>" src/compiler/stmt.rs
```

---
*Migrated verbatim from `todo/deep/unify-block-statement-and-value-compilation.md` on 2026-09-08, when the in-repo `todo/` directory was retired in favour of GitHub issues. See [`docs/issue-workflow.md`](https://github.com/tokuhirom/mutsu/blob/main/docs/issue-workflow.md).*

tokuhirom:
A bare `{ ... }` block was the last control construct compiled by two fully independent passes: the `Stmt::Block` arm of `compile_stmt` for statement position (emitting `OpCode::BlockScope`) and `compile_do_block_expr` for value position (emitting `OpCode::DoBlockExpr`). Both decided the same things about the body — implicit try, phaser scope, `let`/`temp`, import scope, per-execution `state` reset — and each wrote its own answer, in a different order and over a different case set.

## The opcode question, answered

GH-7569 asked what has to be settled before any unification lands: **should `BlockScope` and `DoBlockExpr` be one opcode?** [ADR-0076](https://github.com/tokuhirom/mutsu/blob/claude/mutsu-issue-7569-offd89/docs/adr/0076-bare-block-keeps-two-opcodes-one-shape.md) records the answer as **no**, with the reasoning.

The two ops carry genuinely different runtime contracts. `BlockScope` is a lexical scope boundary: it clones the env, scans its own opcode range for owned/state slots and topic-binding ops, pushes an `outer_scope_locals` frame, a block-scope depth, a lexical class scope and an enum scope, snapshots the routine registry, pushes an anonymous callframe for a genuine source block, and runs seven PRE/ENTER/body/KEEP/UNDO/POST ranges. `DoBlockExpr` is a value-producing body that catches `leave`/`succeed` and deliberately does *not* restore the env.

Merging them would impose the seven-section, env-cloning machinery on the far hotter of the two — every `do { ... }`, every routine tail block and every string-interpolation `{ ... }` goes through `DoBlockExpr` — while the union variant would also have to carry both payloads under the pinned `size_of::<OpCode>() <= 48`. The reverse direction fails on semantics: the statement form's env restore is what stops a block-local `my` from leaking, and the value form must not have it.

## What was actually duplicated

The *classification*, not the opcode. That moves into one shared `Compiler::classify_block_shape` (`src/compiler/block_shape.rs`) returning a `BlockShape` both passes `match` on, in the same order, over the same case set.

Sharing it fixes a divergence from Rakudo that was unreachable from the position which had the logic right. The value form emitted its per-execution `ResetStateLocals` only on the paths *below* its CATCH/CONTROL and ENTER/LEAVE early returns, so a `state` counter restarted correctly in a plain `do` block but persisted as soon as the same block grew a phaser:

```raku
sub f() { do { state $n = 0; $n++; CATCH { default {} }; $n } }
say f(); say f(); say f();   # raku: 1 1 1   mutsu (before): 1 2 3
```

Both statement-position spellings were already correct — that arm computes the reset before its dispatch. The value arm now does too, which fixes the CATCH and the ENTER shapes together.

The statement-side change is a mechanical `if`/`else if` → `match` conversion with identical arm bodies, so that path is behaviour-identical.

## A shape the value path declines, and why

`BlockShape::LetBlock` is deliberately folded into the value path's plain arm. Giving `compile_do_block_expr` the `let` arm the statement form has was the obvious completion — it was implemented, measured against roast, and backed out.

**`Expr::DoBlock` is not a Raku block.** The parser and about a dozen compiler desugars use it as a generic "run these statements, yield a value" node: item context (`$( let $a = 23; $a )`), the chained-comparison desugar, `cas`, compound-assignment lowering, method-body wrappers. None of those introduce a scope at which a `let` may resolve. Emitting a `LetBlock` for every `Expr::DoBlock` resolved the save at the innermost wrapper instead of the enclosing block, and broke the idiom roast leans on throughout:

```raku
my $a = 42;
{
  is($(let $a = 23; $a), 23, "let() changed the variable");
  Mu;
}
is $a, 42, "let() should restore the variable, as our block failed";
```

That cost 3 subtests in `roast/S04-blocks-and-statements/let.t` and 1 in `temp.t`, both whitelisted. So the value path declines the shape with the reason stated at the fold, and the new test pins the `$( ... )` direction so a future attempt fails loudly rather than silently. The genuine `do { let $x = 2; Nil }` gap this leaves is real and filed as #7635 — closing it needs a marker distinguishing a real source block from a synthesized wrapper on the AST node itself.

`src/vm/` and `src/opcode.rs` are consequently back to `main` apart from one clarified doc comment; the diff is compiler-side.

## Tests

`t/block-shape-statement-value-parity.t` — 12 assertions covering the `state` restart in every value-position shape and the `$( ... )` `let` scoping. **All 12 pass unchanged under `raku`.**

## Verification

| | result |
| --- | --- |
| `cargo fmt --all` | clean |
| `make lint` (all four configurations) | green |
| `make test` | 3853 files, 40839 tests, **PASS** |
| `make roast` | 1436 files, 218939 tests — only the three container-only failures documented in `docs/agent-environments.md` (`file-tests.t` 4 and `filetest.t` 25, both `uid 0`; `IO-Socket-Async.t`, sandboxed network), matching the documented counts exactly |

Closes #7569

https://claude.ai/code/session_01JLq8g9GoujE4i6STRxeD1v

---
_Generated by [Claude Code](https://claude.ai/code/session_01JLq8g9GoujE4i6STRxeD1v)_

### DIFF
--- a/docs/adr/0076-bare-block-keeps-two-opcodes-one-shape.md
+++ b/docs/adr/0076-bare-block-keeps-two-opcodes-one-shape.md
@@ -1,0 +1,167 @@
+# ADR-0076: A bare block keeps two opcodes, but has one shape classifier
+
+- **Status**: Accepted
+- **Date**: 2026-09-08
+- **Related**: [GH-7569](https://github.com/tokuhirom/mutsu/issues/7569),
+  `src/compiler/block_shape.rs`, the `Stmt::Block` arm of `src/compiler/stmt.rs`,
+  `src/compiler/helpers_do_expr.rs`, `OpCode::BlockScope` / `OpCode::DoBlockExpr` /
+  `OpCode::LetBlock` in `src/opcode.rs`, `src/vm/vm_misc_scope.rs`,
+  `src/vm/vm_misc_block.rs`,
+  `news/2026-09/unify-statement-expression-control-construct-compilation.md`
+
+## 1. Context
+
+The campaign recorded in
+`news/2026-09/unify-statement-expression-control-construct-compilation.md` gave every
+statement/expression control construct a single lowering shared by both source positions:
+`for` (`src/compiler/control_for.rs`) and `if`/`elsif` (`src/compiler/control_if.rs`) each
+have one, and `while` / C-style `loop` / `lazy for` in expression position are thin
+`gather`-desugaring wrappers over the statement path.
+
+The bare `{ ... }` block was the one construct left compiled by two independent passes:
+
+- **statement position** — the `Stmt::Block` arm of `Compiler::compile_stmt`, emitting
+  `OpCode::BlockScope`;
+- **value position** — `compile_do_block_expr` / `compile_do_block_expr_scoped`, emitting
+  `OpCode::DoBlockExpr`.
+
+GH-7569 asked the question that has to be settled before any unification lands: **should
+`BlockScope` and `DoBlockExpr` be one opcode?** The two arms share neither an opcode nor a
+skeleton, which is exactly why the `for`/`if` unification did not sweep them up.
+
+## 2. Decision
+
+**Two opcodes, one shape classifier.**
+
+1. `BlockScope` and `DoBlockExpr` stay separate opcodes with separate runtime contracts
+   (§3). This is not a deferral: it is the answer.
+2. The *classification* both passes were computing independently — "what kind of block is
+   this" — becomes a single shared function,
+   [`Compiler::classify_block_shape`](../../src/compiler/block_shape.rs), returning a
+   `BlockShape` (`ImplicitTry` / `PhaserScope` / `LetBlock` / `ImportScope` / `Plain`).
+   Both passes `match` on it, in the same order, over the same case set.
+3. The per-execution `state` reset is emitted **before** the shape dispatch in both passes,
+   as an unconditional part of the skeleton rather than a per-arm decision.
+
+## 3. Why the opcodes stay separate
+
+The pair looked like duplication, but the two ops carry genuinely different runtime
+obligations, and the cheap one is by far the hot one.
+
+`BlockScope` is a **lexical scope boundary**. `exec_block_scope_op` clones the env, scans
+its own opcode range for `SetLocalDecl` / `DeclareOurScalar` / `StateVarInit` slots and for
+topic-binding ops, pushes an `outer_scope_locals` frame, a block-scope depth, a lexical
+class scope, an enum scope and a once scope, snapshots the routine registry, pushes an
+anonymous callframe when the block is a genuine source `{ ... }`, and runs seven ranges —
+PRE, ENTER, body, KEEP, UNDO, POST — which is what its seven patch points are for. It
+produces no value.
+
+`DoBlockExpr` is a **value-producing body with control-flow catching**. It pushes a once
+scope and an enum scope, optionally snapshots the routine registry (only when the body
+actually declares one), catches a matching `leave` and a `succeed` and turns each into the
+block's value, and deliberately does *not* restore the env in the common case.
+
+Merging them would mean giving every value-position block the seven-section, env-cloning,
+callframe-pushing machinery. That is the wrong direction on cost: `DoBlockExpr` is emitted
+for every `do { ... }`, every routine tail block and every string-interpolation `{ ... }`,
+so it is executed far more often than `BlockScope`, and the union variant would also have
+to carry both payloads while `size_of::<OpCode>() <= 48` stays pinned by
+`opcode_size_guard`.
+
+The reverse direction — expressing the statement form as the value form with the result
+discarded — fails on semantics rather than cost. The statement form's env restore is what
+stops a block-local `my` from leaking; the value form deliberately lets an outer mutation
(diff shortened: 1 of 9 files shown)


tokuhirom:
Closing this as superseded by #7636, which landed the same work in a fuller form while this branch was running its pre-publication suites.

**Why there were two.** Both sessions were pointed at #7569 by the user. This branch claimed the issue at 12:04:33 (comment 5584841010) and re-read the comments immediately, finding no other claim. `-3d2ubt` claimed at 12:35:19 (comment 5585216412) — 31 minutes later, higher id — and posted a note deciding to proceed anyway on the grounds that this branch had nothing pushed yet. That tiebreaker is not part of the protocol, and cannot be: a claim exists precisely to cover the window before anything is pushed. The failure on this side is the mirror image — this session claimed and then worked for ~85 minutes without re-reading the issue, so the stand-down request sat unread until after the PR was open. Re-reading before the expensive phase (two full suite runs) and again before opening the PR would have caught it.

**Nothing here is lost.** #7636 went further than this branch on every axis: `BlockPlan::analyze` + `compile_block_construct` in a new `src/compiler/control_block.rs` make both entry points thin wrappers (~250 lines out of `stmt.rs`), where this branch shared only the classifier and left the two skeletons in place. Its ADR-0076 records the same two-opcodes decision.

It also solved the trap this branch hit and backed out of. Giving the value path a `let`/`temp` arm broke `roast/S04-blocks-and-statements/let.t` (3 subtests) and `temp.t` (1), because `Expr::DoBlock` is not a Raku block — the parser uses it for item context `$( let $a = 23; $a )` too, so the save resolved at the innermost wrapper. #7636 fixed it properly with `STMT_LIST_CONTEXTUALIZER_LABEL` gating `owns_let_scope`, which is exactly the "marker distinguishing a real source block from a synthesized wrapper" this branch filed as #7635. Verified against `main` at 83f776e: `do { let $x = 2; Nil }` restores, and `$( ... )` still defers to the enclosing block. #7635 is closed accordingly.

**One real gap survives**, and it is the specific bug that motivated this investigation. `t/block-position-parity.t` pins `state` in a plain value block, and pins `CATCH` and `ENTER` separately — but not their intersection. Before the unification, `do { state $n = 0; $n++; CATCH { default {} }; $n }` counted 1 2 3 because the value pass emitted its `ResetStateLocals` only on the paths *below* its CATCH/phaser early returns. `main` gets this right now (measured: `1 1 1`), but nothing would catch a regression that reintroduced the per-arm ordering. Offering that as a small test-only follow-up rather than opening a third PR on this issue unprompted.

---

Issue #7555, opened 2026-09-08
Four bundled-library regressions block making the vendored `Test` the default


tokuhirom:
These are the reason the vendored-`Test`-by-default flip was withdrawn on
2026-09-08 (`todo/deep/vendor-real-test-module-flip.md`). They do **not** affect
mutsu as shipped: the native provider is the default, and under it the
`Bundled-library test suites` gate is green. They are what has to be fixed before
the flip can be retried.

Under the vendored module as default the gate
(`scripts/battery-testsuite.sh`, run inside CI's `test` job) reports
**282/312 with 9 regressed whitelisted files** — four real ones below, plus five
`DBIish` rows that need live database servers.

They are *not* `Test` compatibility gaps — they are general interpreter bugs the
real module's code shapes reach and mutsu's native provider never did, which is
the same pattern the rest of that campaign followed (`is rw` slice writeback, the
deferred-grep capture merge, the EVAL/import loss). Each one below is a separate
root cause and wants its own fix.

## How to tell a provider-caused failure from an environment one

Run the file under both providers; the switch is the control, not any single
commit on the branch:

```
cd tmp/battery-testsuite/<Battery>
                     target/release/mutsu -I lib t/<file>     # vendored (default)
MUTSU_REAL_TEST=0    target/release/mutsu -I lib t/<file>     # native
```

Passing under `MUTSU_REAL_TEST=0` and failing by default means the switch caused
it. (Reverting only one fix commit is the *wrong* control and gave a false
"not mine" reading once already.)

## 1. `NativeLibs t/01-basic.t` test 6 — root-caused: custom `sub EXPORT` does not install type objects

Measured: `visible` under the native provider, `MISSING` under the vendored one —
but the provider is a red herring. **Two independent mutsu divergences were
cancelling each other out**, and the `use-ok` EVAL merely stopped the cancellation.

`NativeLibs.pm6` makes `NativeCall` visible to its importer with a custom export
sub, `sub EXPORT(|) { Map.new('NativeCall' => NativeCall, ...) }`, placed above
its `unit module` line.

Measured truth table (fixtures under `tmp/ex/`, `tmp/pre/`, plain `use`, no EVAL
and no `Test` anywhere):

| module shape | rakudo | mutsu |
| --- | --- | --- |
| custom `sub EXPORT` returning a Map/Hash of a **type object** | `visible` | `MISSING` |
| `use X` above the `unit module` line | `MISSING` | `visible` |

[...]

PR #7663, opened 2026-09-08
fix(compiler): run a tail bare block's ENTER/LEAVE phasers

tokuhirom:
## Root cause

`Cro::HTTP`'s `t/http-middleware.rakutest` — item 4 of the four bundled-library
regressions in #7555 — failed under the vendored `Test` module with the third
subtest's conditional middleware apparently never installed: requests that
should have returned `403` came back with the application's own page, and the
caching subtest's counter advanced where a cached body was expected.

The middleware *was* installed. The **previous** subtest's server was still
listening.

Every block in that file ends the same way:

```raku
{
    my Cro::Service $service = Cro::HTTP::Server.new(…);
    $service.start;
    …
    LEAVE $service.stop();
}
```

That block is the **last statement** of the routine body it sits in, so the
compiler inlines it — a tail block is the body's implicit return value, and
inlining is how that value reaches the caller. Inlining drops the block's
phasers, so the `LEAVE` was compiled away entirely, `$service.stop()` never
ran, and later requests on that port were answered by a server that should have
been gone.

`compiler/mod.rs`'s tail-block site already knew this and routes a block
carrying `ENTER`/`LEAVE`/`KEEP`/`UNDO`/`PRE`/`POST` through a real `BlockScope`.
The three tail-block sites in `compiler/helpers_sub_body.rs` — one in
`compile_routine_body_stmts`, two in `compile_closure_body_with_routine_flag` —
did not, so the defect applied to every named sub, method, anonymous closure and
block argument:

```raku
sub st(&b) { b() }
st {
    { LEAVE say 'a'; }   # printed
    { LEAVE say 'b'; }   # tail block: never printed
}
```

Only `do { … }` escaped it, compiling through the expression path that already
had the check.

## The fix

Each of the three sites now makes the same check and hands a phaser-carrying
tail block to `compile_phaser_block_scope` with `PhaserBlockResult::Push`, so
the block's value still becomes the body's implicit return while its phasers
run. A block with no phasers keeps the inline path unchanged.

## Verification

- `Cro::HTTP`'s `t/http-middleware.rakutest` now passes **24/24 under the
  vendored `Test`** (`MUTSU_REAL_TEST=1`); before this it died after 31
  assertions with `X::Cro::HTTP::Error::Client`.
- New `t/leave-phaser-in-tail-block.t` covers a named sub, a method, an
  anonymous closure and a block argument, checks `ENTER`/`LEAVE` ordering, and
  checks that the tail block still supplies the body's value. It fails on `main`
  and passes here; rakudo passes it unchanged.
- `make test` PASS (3866 files, 40998 tests), `make lint` clean in all four
  configurations, `make roast` byte-identical to the pre-change run — green
  apart from the three container-only files `docs/agent-environments.md`
  documents (`6.c/S32-io/file-tests.t`, `S16-filehandles/filetest.t`,
  `S32-io/IO-Socket-Async.t`), matching their recorded shapes exactly.

Refs #7555 (item 4 of four; item 2 landed in #7653. The `NativeLibs`
custom-`EXPORT` item and `http2-request-parser.rakutest` are unaffected and stay
open — the latter has a separate cause, an extra `check 4` mutsu emits from an
async tap that rakudo does not.)

https://claude.ai/code/session_01Vfxeym2iMEoeTaWPbvCiqu

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vfxeym2iMEoeTaWPbvCiqu)_

### DIFF
--- a/news/2026-09/leave-phaser-in-tail-block.md
+++ b/news/2026-09/leave-phaser-in-tail-block.md
@@ -1,0 +1,63 @@
+# A tail bare block's `LEAVE` runs again
+
+`Cro::HTTP`'s `t/http-middleware.rakutest` failed under the vendored `Test`
+module — one of the four bundled-library regressions in
+[#7555](https://github.com/tokuhirom/mutsu/issues/7555) — with the third
+subtest's conditional middleware apparently never installed: requests that
+should have come back `403` returned the application's own page, and the
+caching subtest saw its counter advance where a cached body was expected.
+
+The middleware was installed. The **previous** subtest's server was still
+listening.
+
+## What was wrong
+
+Each of that file's blocks ends the same way:
+
+```raku
+{
+    my Cro::Service $service = Cro::HTTP::Server.new(…);
+    $service.start;
+    …
+    LEAVE $service.stop();
+}
+```
+
+That block is the **last statement** of the routine body it sits in, so the
+compiler inlines it: a tail block is the body's implicit return value, and
+inlining it is how that value reaches the caller. Inlining, however, drops the
+block's phasers — the `LEAVE` was compiled away entirely, `$service.stop()`
+never ran, and every later request on that port was answered by the pipeline of
+a server that was supposed to be gone.
+
+The mainline compiler (`compiler/mod.rs`) already knew this: its tail-block site
+routes a block carrying `ENTER`/`LEAVE`/`KEEP`/`UNDO`/`PRE`/`POST` through a
+real `BlockScope` instead of inlining it. The three tail-block sites in
+`compiler/helpers_sub_body.rs` — one in `compile_routine_body_stmts`, two in
+`compile_closure_body_with_routine_flag` — did not, so the bug applied to every
+named sub, method, anonymous closure and block argument:
+
+```raku
+sub st(&b) { b() }
+st {
+    { LEAVE say 'a'; }
+    { LEAVE say 'b'; }   # tail block: never printed
+}
+```
+
+Only `do { … }` was unaffected, because it compiles through the expression path
+that already had the check.
+
+## The fix
+
+The three sites now make the same check the mainline does and hand a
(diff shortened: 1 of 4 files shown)


tokuhirom:
CI's `test` job went red on `t/log-async-battery.t` test 5
(`info maps to the INFO level`, got `WARNING`). Diagnosed and fixed in
`e9c8fd3`; the failure is **not caused by the compiler change in this PR**, but
it is a real bug in that test, so it is fixed here rather than re-run away.

## Why it failed

`Log::Async`'s `send-msg` emits each message from its own thread —
`(start $.source.emit($m))`, `modules/Log-Async/lib/Log/Async.rakumod:88` — so
the order a tap sees five messages in is whichever `start` the scheduler runs
first, not the order they were logged. The test indexed positionally
(`@seen[2]<level>`), which only held while nothing perturbed the scheduling.

## Evidence it predates this PR

`scripts/flake-repro.sh -n 25 -l 4`, release build, same box:

| build | result |
| --- | --- |
| this branch, before the test fix | pass=21 fail=4 |
| **`origin/main` compiler source, everything else identical** | **pass=23 fail=2** |
| this branch, after the test fix (`-n 30`) | **pass=30 fail=0** |

The control was made by checking out `origin/main`'s
`src/compiler/helpers_sub_body.rs` alone and rebuilding, so the only difference
between rows 1 and 2 is this PR's diff. It flakes either way; 4/25 vs 2/25 is
noise on a race. `Log::Async` and its test contain no phasers at all, so the
tail-block codegen this PR changes cannot reach them.

## The fix

The assertions now go through a message → level map instead of `@seen[N]`, so
they check exactly what the module guarantees — each message keeps its own text,
severity and timestamp — and nothing it does not. All eight behaviours are still
covered; nothing was skipped, disabled or quarantined, and no `flaky-tests.txt`
entry was added.

---

Issue #7667, opened 2026-09-08
perf: an HTTP/2 HEADERS frame costs 10-110 ms to process in Cro's request parser


tokuhirom:
Split out of #7555 item 3, where it is the actual cause. mutsu computes the
right answer; it is just slow enough that an upstream test loses a race it wins
under rakudo.

## The measurement

Timing `$fake-in.emit($frame)` in `Cro::HTTP`'s `t/http2-request-parser.rakutest`
— that call runs the whole synchronous chain (frame -> `Cro::HTTP2::RequestParser`
-> tap) — release build, this container:

```
  FRAME Headers      sid=3 cost=38.7 ms
  FRAME Data         sid=3 cost=2.9 ms
  FRAME Headers      sid=3 cost=10.1 ms
  FRAME Continuation sid=3 cost=34.2 ms
  FRAME Data         sid=3 cost=3.2 ms
  FRAME Continuation sid=3 cost=60.5 ms
  FRAME Headers      sid=3 cost=81.6 ms
  FRAME Headers      sid=5 cost=93.8 ms
  FRAME Headers      sid=5 cost=109.2 ms
  FRAME Data         sid=3 cost=3.1 ms
  FRAME Data         sid=5 cost=4 ms
  FRAME Continuation sid=3 cost=90.9 ms
```

**HEADERS/CONTINUATION: 10-110 ms. DATA: ~3 ms.** The header path is the hot
one, and it gets worse as the connection accumulates streams.

## What it is NOT

Each of these was measured and ruled out, so the next session does not have to
re-do them:

- **Not `start` / `Supplier.emit`.** mutsu is *faster* than rakudo here:
  single `await start { 1 }` 0.6 ms vs rakudo 82.1 ms; six `Supplier.emit`
  calls into a tap that spawns a `start` cost 0.2/0.2/0.2/0.2/0.1/0.2 ms vs
  rakudo 1.3/0.8/0.1/0.1/0.1/0.1 ms.
- **Not the body Supply -> Promise path.** With a `note` either side, the
  parser's `$stream.body.done` at `t=…451` is followed by the consumer's
  `.body-blob.result` returning at `t=…453` — 2 ms.
- **Not mostly HPACK.** A fresh `HTTP::HPACK::Decoder` decoding the same
  5-header block 20 times: mutsu 48 ms, rakudo 9.5 ms — so 2.4 ms vs 0.48 ms
[...]

PR #7714, opened 2026-09-09
fix: run a `unit module`'s custom `sub EXPORT`

tokuhirom:
`NativeLibs t/01-basic.t` is the last open item of #7555, the four
bundled-library regressions blocking the vendored-`Test`-by-default flip. Under
the vendored module its test 6 reported `MISSING`; under the native provider it
passed only because a file-scope `use` leak happened to supply the answer rakudo
gets from the module's custom export. It now passes **23/23 under both
providers, cold and warm**, through the mechanism rakudo uses.

Five independent interpreter bugs sat in the way. None of them is about `Test` —
the switch only changes which code shapes get reached.

## 1. A `sub EXPORT` above the `unit module` line was never called

`unit module Foo;` packages *the rest of the scope*, so what precedes it stays in
the compilation unit's outer scope — which is the whole mechanism behind a custom
export hook, since `use` looks it up there. Every module that has one writes it
above the line, as `NativeLibs` does:

```raku
use NativeCall;
sub EXPORT(|) {
    my $exp = &trait_mod:<is>.candidates.first: { .signature ~~ :(Routine, :$native!) };
    Map.new('NativeCall' => NativeCall, '&trait_mod:<is>' => $exp.dispatcher)
}
unit module NativeLibs:auth<salortiz>:ver<0.0.9>;
```

`compile_unit` emitted the runtime package switch for the *whole* compilation
unit before the sub-hoist pass, so the hook registered as `Foo::EXPORT` and
`apply_module_export`'s `GLOBAL::EXPORT` lookup missed it. The custom export of
every `unit module` was silently a no-op. The hoist is now split at the
declaration, and the runtime package handed back for the preceding statements'
own execution, so the declaration re-establishes it when execution reaches it.
Measured against rakudo: above the line runs, below it does not.

## 2. `.dispatcher` answered the candidate, not the proto

Raku generates a dispatcher even for a `multi` written with no `proto`, and
`.dispatcher` on a candidate returns it. mutsu answered the candidate itself —
and, for a `Sub`-shaped value, an ADR-0070 `<composed-method:dispatcher>` stub,
because nothing handled the method at all on that shape. Either way
`.dispatcher.candidates` reported a single entry, which silently narrows a whole
multi to the one candidate that was introspected. That is exactly what the
`Map.new('&name' => $candidate.dispatcher)` re-export idiom does. `Routine`
handles now also answer `.is_dispatcher`/`.multi`, which the fixed `.dispatcher`
makes reachable.

## 3. NativeCall exported a content-free `trait_mod:<is>`

Rakudo's `NativeCall.rakumod` declares its four trait candidates (`:$native!`,
`:$symbol!`, `:$nativeconv!`, `:$encoded!`) as ordinary multis and exports them,
which is what makes `&trait_mod:<is>.candidates` introspectable. mutsu registered
only a bare exported *name*, so `.candidates` answered one entry with no
signature, `.first` gave `Nil`, and the module died on `Any.dispatcher`.

The four are now spliced in as a prelude, gated on the compunit both using
NativeCall and naming `trait_mod:<is>` itself. Both halves of the gate matter:
their mere presence flips `has_proto`/`has_multi_candidates`, which would route
every other unknown `is` trait in the file through custom dispatch. Their bodies
are empty on purpose and this is not a stub standing in for an unimplemented
feature — mutsu applies all four natively at declaration time
(`register_native_call_sub`), and `registration_sub.rs` already excludes
`native`/`symbol`/`nativeconv`/`encoded` from the custom-`trait_mod:<is>` loop
for that reason, so these candidates are never what applies the trait. They exist
to make the export surface introspectable, exactly as rakudo's do. A `TODO` marks
the one remaining gap (calling one explicitly should apply the trait, which needs
the C-FFI descriptor to be installable on an already-registered `Routine`).

Two lifetime bugs surfaced with them:

- A prelude splice declared as a `multi` registers under arity-suffixed keys
  (`GLOBAL::name/2`, and the chained `__mN` slots), never under the bare
  `GLOBAL::name` the single-sub path records — so the prelude bookkeeping missed
  it and `reinstate_module_functions` dropped the candidates as importer-scoped
  `GLOBAL::` aliases the first time a block scope unwound. Recording the bare
  name instead would have gated NativeCall's own `GLOBAL::trait_mod:<is>` entry
  behind this splice's compunit and hidden it everywhere else, so the real keys
  are recorded after registration.
- Raku re-runs `sub EXPORT` on *every* later import of an already-loaded module,
  by which time the importer's scope holds none of the module's own lexicals —
  `Map.new('NativeCall' => NativeCall, …)` quietly degraded to the bareword
  string `"NativeCall"`, which then shadowed the real package. The remembered
  hook now carries the module-scope env it was first called in, and both call
  paths anchor to the module's own compunit.

## 4. A `Signature` literal lost its parameters across the precompilation cache

Independent of `EXPORT`, and reproducible in plain Raku:

```raku
# lib/SigRT.rakumod
unit module SigRT;
multi sub mm(Routine $r, :$native!) is export { 'native' }
multi sub mm(Routine $r, :$symbol!) is export { 'symbol' }
our sub probe is export {
    &mm.candidates.map({ (.signature ~~ :(Routine, :$native!)).Str }).join(",")
}
```

```
cold: True,False      # rakudo: True,False
warm: False,False
```

A `Signature` literal is an `Expr::Literal` holding a `Signature` instance, but
its structured parameter data lives in a process-global side table keyed by that
instance's id — which nothing in a serialized value can reach. A module restored
from the cache came back with a literal whose id named nothing,
`extract_sig_info` fell through to its empty-params legacy shape, and the
smartmatch flipped from True to False. Only the **second** run of a program sees
it, which is precisely what CI's always-cold runners cannot catch.

`SerValue::Instance` now carries the `SigInfo` alongside a `Signature`, and the
value is rebuilt whole from it on load — under a *fresh* id, so it cannot collide
with an unrelated instance holding the recorded id in this process.
`CACHE_FORMAT_VERSION` moves to 12.

## 5. Calling an enum value is a coercion, like calling its enum type

Surfaced by the bundled-library gate: `Log::Async 16-imports2.rakutest` had been
passing **for the wrong reason**. Its `sub EXPORT` reads `@*ARGS`, and the re-run
used to see the *importer's* copy, so mutsu matched the module's `--log=` branch
against the test's `--level=trace` (with a `Use of Nil in string context`
warning) and left `$level` a `WhateverCode`. Running the hook in the module's own
scope — item 3 above, and what rakudo does — makes `$level` the plain enum value
rakudo also produces there, and the test then calls it.

Rakudo answers that call. `Lv::DEBUG(1)` is the same coercion as `Lv(1)`, both
giving `Lv::TRACE`, and an undefined value when the enum holds no such value.
mutsu implemented only the type-object half: `vm_call_on_value` already routes an
invoked `Package` to the bare-name call path — "invoking a type object is a
coercion, not a `CALL-ME` call" — but an enum *value* is a `ValueView::Enum`,
matched neither that arm nor the `Instance | Package` one, and fell through to
the `CALL-ME` fallback, dying with "No such method 'CALL-ME'".

One divergence is deliberately left alone because it predates this and lives in
the shared coercion helper rather than the new arm: `Lv('DEBUG')` answers
`Lv::DEBUG` in mutsu where rakudo answers `Nil`. The value form now agrees with
the type form, which is the property this needs.

## Verification

- `NativeLibs`' upstream suite (`batteries.lock` commit `f3e78510`) under the
  release binary: `01-basic.t` **23/23** in all four combinations of
  provider × cold/warm cache. `02-cannon-name.t` and `20-compile.t` green;
  `10-search.t`'s single failure is the host-dependent `mysqlclient` one
  `scripts/battery-testsuite.sh` already documents, unchanged from `main`.
- `scripts/battery-testsuite.sh`: `Log::Async 16-imports2.rakutest` PASS. The
  only rows still regressed locally are the six `DBIish` mysql ones, which need
  a live MySQL server this container has no service for — they pass in CI's own
  run of the same gate.
- Five new pins, each of which **fails on `main`** and **passes under real `raku`
  unchanged**: `t/unit-module-export-sub.t`, `t/multi-candidate-dispatcher.t`,
  `t/nativecall-trait-mod-is-candidates.t`,
  `t/signature-literal-precomp-warm-cache.t` (built on
  `t/precomp-warm-cache-parity.t`'s cold/warm subprocess pattern, since a
  same-process test cannot see the bug) and `t/enum-value-call-coercion.t`.
- `make test` **PASS** (3906 files, 41480 tests). `make lint` clean in all four
  configurations. `make roast` byte-identical to the pre-change run: green apart
  from the three container-only files `docs/agent-environments.md` documents
  (`6.c/S32-io/file-tests.t`, `S16-filehandles/filetest.t`,
  `S32-io/IO-Socket-Async.t`), each matching its recorded failure shape exactly.

Rebased onto `main` after #7717/#7715 landed; the conflict in
`registration_sub.rs` is resolved by composing both sides — `main`'s new
`prelude_sub_names` insert is kept (and left outside the new `!multi` guard,
since it records the ambient *name*, which a multi prelude needs too).

Refs #7555 — item 1 of the four. Items 2 and 4 landed in #7653 and #7663; item 3
is tracked separately as #7667 (`todo:perf`), so this is the last of the four
that is fixable here. The issue itself is left open until that one lands.

https://claude.ai/code/session_019oTR3uw7PrAokSmncMtPsM

### DIFF
--- a/news/2026-09/unit-module-custom-export-sub.md
+++ b/news/2026-09/unit-module-custom-export-sub.md
@@ -1,0 +1,124 @@
+# A `unit module`'s custom `sub EXPORT` now runs
+
+`NativeLibs t/01-basic.t` was the first of the four bundled-library regressions
+in [#7555](https://github.com/tokuhirom/mutsu/issues/7555) — the ones blocking
+the vendored-`Test`-by-default flip. Under the vendored module its test 6
+(`ok ::('NativeCall') !~~ Failure`) reported `MISSING`; under the native
+provider it passed, but for the wrong reason. It now passes **23/23 under both
+providers**, and it does so through the mechanism rakudo uses.
+
+Four independent interpreter bugs sat between here and there. Each is general;
+none is about `Test`.
+
+## 1. A `sub EXPORT` above the `unit module` line was never called
+
+`unit module Foo;` packages *the rest of the scope*. What precedes it stays in
+the compilation unit's outer scope — and that positional rule is the whole
+mechanism behind a custom export hook, because `use` looks the hook up in the
+compunit's outer scope, not inside the module. Every module that has one
+therefore writes it above the line, as `NativeLibs` does:
+
+```raku
+use NativeCall;
+sub EXPORT(|) {
+    my $exp = &trait_mod:<is>.candidates.first: { .signature ~~ :(Routine, :$native!) };
+    Map.new('NativeCall' => NativeCall, '&trait_mod:<is>' => $exp.dispatcher)
+}
+unit module NativeLibs:auth<salortiz>:ver<0.0.9>;
+```
+
+`compile_unit` emitted the runtime package switch for the *whole* compilation
+unit before the sub-hoist pass, so the hook registered as `Foo::EXPORT` and
+`apply_module_export`'s `GLOBAL::EXPORT` lookup missed it. The custom export of
+every `unit module` in the ecosystem was silently a no-op. The hoist is now
+split at the declaration: what precedes it registers under `GLOBAL`, and the
+runtime package is handed back for those statements' own execution, so the
+declaration re-establishes it when execution reaches it. Measured against
+rakudo: above the line runs, below it does not.
+
+## 2. `.dispatcher` answered the candidate, not the proto
+
+Raku generates a dispatcher even for a `multi` written with no `proto`, and
+`.dispatcher` on a candidate returns it. mutsu answered the candidate itself —
+and, for a `Sub`-shaped value, an ADR-0070 `<composed-method:dispatcher>` stub,
+because nothing handled the method at all on that shape. Either way
+`.dispatcher.candidates` reported a single entry, which silently narrows a whole
+multi to the one candidate that was introspected. That is exactly what the
+`Map.new('&name' => $candidate.dispatcher)` re-export idiom does, so
+`NativeLibs` handed its importers a one-candidate `&trait_mod:<is>`. `Routine`
+handles now also answer `.is_dispatcher`/`.multi`, which the fixed
+`.dispatcher` makes reachable.
+
+## 3. NativeCall exported a content-free `trait_mod:<is>`
+
+Rakudo's `NativeCall.rakumod` declares its four trait candidates
+(`:$native!`, `:$symbol!`, `:$nativeconv!`, `:$encoded!`) as ordinary multis and
+exports them, which is what makes `&trait_mod:<is>.candidates` introspectable.
+mutsu registered only a bare exported *name*, so `.candidates` answered one
+entry with no signature, `.first` gave `Nil`, and the module died on
+`Any.dispatcher`. The four candidates are now spliced in as a prelude, gated on
+the compunit both using NativeCall and naming `trait_mod:<is>` itself — the
+gate matters, because their mere presence flips
+`has_proto`/`has_multi_candidates` and would route every other unknown `is`
+trait in the file through custom dispatch. Their bodies are empty on purpose:
+mutsu applies all four natively at declaration time
+(`register_native_call_sub`), and `registration_sub.rs` already excludes them
+from the custom-`trait_mod:<is>` loop for that reason, so these candidates are
+never what applies the trait — they exist to make the export surface
+introspectable, as rakudo's do.
+
+Two lifetime bugs surfaced with them. A prelude splice declared as a `multi`
+registers under arity-suffixed keys (`GLOBAL::name/2`, and the chained `__mN`
+slots), never under the bare `GLOBAL::name` the single-sub path records, so the
+prelude bookkeeping missed it entirely: `reinstate_module_functions` dropped the
+candidates as importer-scoped `GLOBAL::` aliases the first time a block scope
+unwound. And Raku re-runs `sub EXPORT` on *every* later import of an
+already-loaded module, by which time the importer's scope holds none of the
+module's own lexicals — `Map.new('NativeCall' => NativeCall, …)` quietly
+degraded to the bareword string `"NativeCall"`, which then shadowed the real
+package. The remembered hook now carries the module-scope env it was first
+called in, and both call paths anchor to the module's own compunit.
+
+## 4. A `Signature` literal lost its parameters across the precompilation cache
+
+This one is independent of `EXPORT` and reproduces in plain Raku:
+
+```raku
+# lib/SigRT.rakumod
+unit module SigRT;
+multi sub mm(Routine $r, :$native!) is export { 'native' }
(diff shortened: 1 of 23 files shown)


tokuhirom:
The `test` job's bundled-library gate failed on `Log::Async 16-imports2.rakutest`. Fixed in `50286a8` — and the interesting part is that the file was **passing for the wrong reason** before.

## What was actually happening

Its `sub EXPORT` reads `@*ARGS` to pick a log level. On `main` the EXPORT re-run saw the *importer's* `@*ARGS`, and mutsu matched the module's `--log=` branch against the test's `--level=trace`:

```
Use of Nil in string context
  in sub EXPORT at modules/Log-Async/lib/Log/Async.rakumod line 148
ok 1 - trace will be logged
```

`$<opt>` was `Nil`, so `$level` came out a `WhateverCode` (`* ≥ <nothing>`) — which is callable, so `$args<level>(TRACE)` happened to work.

Running the hook in the module's own scope (this PR) is what rakudo does, and there `@*ARGS` is empty at `BEGIN` time, so no branch matches and `$level` stays the plain enum value `WARNING`. Measured: **rakudo produces exactly the same `Loglevels` value there**, and its assertion still passes — because in Raku, calling an enum *value* is the same coercion as calling the enum type:

```raku
enum Lv <<:TRACE(1) DEBUG INFO>>;
my $v = DEBUG;
say $v(TRACE);   # rakudo: Lv::TRACE     mutsu (before): No such method 'CALL-ME' for invocant of type 'Lv'
say $v(2);       # rakudo: Lv::DEBUG     mutsu (before): same error
say Lv(1);       # rakudo: Lv::TRACE     mutsu (before): Lv::TRACE  <- only this half existed
```

`vm_call_on_value` already routes an invoked `Package` to the bare-name call path, under the comment *"invoking a type object is a coercion, not a `CALL-ME` call"*. An enum value is a `ValueView::Enum`, so it matched neither that arm nor the `Instance | Package` one and fell through to the `CALL-ME` fallback. The new arm makes the value form agree with the type form.

Pinned by `t/enum-value-call-coercion.t`, which passes under real `raku` unchanged. One pre-existing divergence is deliberately left alone because it lives in the shared coercion helper rather than the new arm: `Lv('DEBUG')` answers `Lv::DEBUG` in mutsu where rakudo answers `Nil`.

## Re-verification

- `scripts/battery-testsuite.sh`: `Log::Async 16-imports2.rakutest` **PASS**. The only rows still regressed locally are the six `DBIish` mysql ones, which need a live MySQL server this container has no service for — they passed in CI's own run of the same gate.
- `make test` **PASS** (3901 files, 41404 tests). `make lint` clean in all four configurations. `make roast` byte-identical to the pre-change run: green apart from the three container-only files `docs/agent-environments.md` documents.

---

Issue #7692, opened 2026-09-09
A nested module's imported `our` variable and class leak into the using scope


tokuhirom:
Split off 2026-09-09 from #7612 while fixing it. #7612's own repro turned out to
be the NativeCall *prelude splice* (a process-global `GLOBAL::` registration
mutsu makes because it has no `NativeCall.rakumod`), fixed there by scoping the
splice's visibility to the compunits it was spliced into. These two leaks are
real, are a different shape, and survive that fix.

Measured against rakudo 2026.07 while investigating #7612. Note that a nested
module's imported **`sub`** does NOT leak — that direction is already correct,
scoped by the `module_registered_functions` delta #7580 introduced. Only
variables and types do.

## Repro

```raku
# lib/L7Var.rakumod
unit module L7Var;
our $var-x is export = 42;
class ClsZ is export { method hi() { 'hi' } }
sub op-user() is export { 1 }

# lib/L7Mid.rakumod
unit module L7Mid;
use L7Var;
sub mid-p() is export { $var-x ~ '/' ~ ClsZ.new.hi }
```

```raku
use lib 'lib';
use L7Mid;                              # imports mid-p and nothing else
say 'mid-p: ',       mid-p();           # both: 42/hi
say 'var leaked: ',  (try $var-x).defined;
say 'class leaked: ',(try ClsZ.new.hi).defined;
say 'sub leaked: ',  (try op-user()).defined;
```

| | mutsu | raku |
| --- | --- | --- |
| `mid-p()` | `42/hi` | `42/hi` |
| `$var-x` in the using scope | `True` | compile error, `Variable '$var-x' is not declared` |
| `ClsZ` in the using scope | `True` | compile error, `Undeclared name: ClsZ` |
| `op-user()` in the using scope | `False` (correct) | not declared |

## Root cause

A module body runs in the **caller's env** (`load_module_inner` → `run_block`),
so the aliases `import_module` installs for the module's own lexical scope land
directly in whatever env triggered the load — the top-level script's, for a
file-scope `use`. The two loops at the end of `runtime/runtime_module_exports.rs`
show the asymmetry:

- the **sub** loop writes into `registry.functions` under `{target_pkg}::{name}`,
  which the `module_registered_functions` delta and `pop_import_scope`'s
  keep-rules then scope correctly;
- the **vars** loop ends in a bare `self.env.insert(target, value)` with
  `target` the unqualified `$var-x`, and nothing removes it when the nested load
[...]

PR #7764, opened 2026-09-09
fix(runtime): scope a transitively-`use`d module's package name to its own importer

tokuhirom:
`use Outer;` — where `Outer.rakumod` itself begins `use Inner;` — left the bare name `Inner` resolvable in the importing file, which never asked for it and where rakudo reports an "Undeclared name".

```raku
# lib/Inner.rakumod:  unit module Inner; class InnerClass is export { }
# lib/Outer.rakumod:  use Inner; unit module Outer; class OuterClass is export { }

use Outer;
OuterClass.new;          # correct: Outer exported it
Inner::InnerClass.new;   # rakudo: Could not find symbol.  mutsu: resolved
```

This is the last piece of #7555 item 1's divergence (b). The alias half of that divergence — the classes and roles a transitively-loaded module declares, and the variables and subs it exports — was closed by #7743 (issue #7692), which scopes a module's imports to its own compunit. Re-measured against `main` at d7a6906, exactly one binding still escaped:

| bare name in the importer | rakudo v2026.07 | mutsu at d7a6906 | with this PR |
| --- | --- | --- | --- |
| `OuterClass` | `OuterPkg::OuterClass` | same | same |
| `OuterPkg` | `OuterPkg` | same | same |
| `InnerClass` | MISSING | MISSING (#7743) | MISSING |
| `InnerRole` | MISSING | MISSING (#7743) | MISSING |
| **`InnerPkg`** | **MISSING** | **`InnerPkg`** | **MISSING** |

## Root cause

`OpCode::RegisterPackage` binds a `unit module X` under its bare name with a plain `env.insert`, and a module body runs in the **caller's** env (`load_module_inner` → `run_block`) rather than in an env of its own. So `Outer`'s own `use Inner;` left the `Inner` package binding in whatever frame triggered the load — the importing file, for a compile-time `use` at file scope. #7743's cleanup restores the caller's plain bindings and undoes the aliases `import_module` recorded; a package name is neither of those, so nothing reached it.

It is dropped now at the end of the load that introduced it, unless it names the module being loaded or something nested under it. A direct `use Inner;` is unaffected — that load keeps its own package name, exactly as rakudo puts it in the importing scope's `MY::`.

### Why the registry diff is consulted, not just the name

The obvious ownership test — "does the qualified name start with the module's own name" — is not sufficient. A module file with no `unit` declarator can register a class under an unrelated name: `roast/packages/RT128156/lib/RT128156/Top1.rakumod` declares a plain `class Top1`, and roast's `S10-packages/precompilation.t` tests 40-41 require it in the *loading* scope's `MY::`. A name-prefix test alone drops it and turns those two green tests red — measured, `make roast` caught it.

## What is deliberately still divergent

A module's `constant`s and enum values reach the importer by the same route and are left alone. Dropping them needs somewhere else for the module's own top-level subs to read them: `module_scope_lexicals` is keyed by package and a module's plain `sub` runs under `GLOBAL`, so removing the env binding turns `Log::Async`'s `sub trace { ... :level(TRACE) ... }` into a bareword `Str` (measured — `t/log-async-battery.t` aborts at test 2). mutsu also keeps module package names in `GLOBAL::` where rakudo keeps them in the importing compunit's `MY::`; this scopes their *visibility* to match without moving that storage. Both belong to the module-lexical scoping campaign #7555 describes.

## Tests

`t/module-transitive-use-does-not-leak-types.t` (new, 9 assertions) passes verbatim under rakudo v2026.07 as well as under mutsu. It covers what #7743 fixed alongside this one, so the whole scoping contract stays pinned from one file:

- a transitively-used class, role and package are all unresolvable in the importer;
- what the used module itself exports still resolves;
- the declaring module's own subs and its classes' methods still resolve the short names it imported;
- the sibling-package NativeCall shape (`unit module Foo::Native; class Handle`, referenced bare from a sibling class's method) still works;
- the two-hop case — a module that `use`s a module that `use`s a third — whose method bodies resolve bare names through their own class's package chain rather than through `env`, so the env path alone does not cover it.

## Verification

- `make lint` — clean in all four configurations (default clippy, `jit` off, wasm32 lib, rustdoc).
- `make test` — PASS, 3919 files / 41648 tests.
- `make roast` — no regression. The only failures are the three container-caused ones `docs/agent-environments.md` records, with exactly the documented signatures: `6.c/S32-io/file-tests.t` and `S16-filehandles/filetest.t` (this container runs as `uid 0`, so the `chmod`-based assertions are meaningless) and `S32-io/IO-Socket-Async.t` (sandboxed network).

Part of #7555 (item 1, divergence (b)).

https://claude.ai/code/session_01ARX7Pe67YipLQtH6Sb1hmA

---
_Generated by [Claude Code](https://claude.ai/code/session_01ARX7Pe67YipLQtH6Sb1hmA)_

### DIFF
--- a/news/2026-09/transitive-module-package-name-no-longer-leaks.md
+++ b/news/2026-09/transitive-module-package-name-no-longer-leaks.md
@@ -1,0 +1,61 @@
+# A transitively-`use`d module's package name no longer leaks into the importer
+
+`use Outer;` — where `Outer.rakumod` itself begins `use Inner;` — left the bare
+name `Inner` resolvable in the importing file, which never asked for it and
+where rakudo reports an "Undeclared name":
+
+```raku
+# lib/Inner.rakumod:  unit module Inner; class InnerClass is export { }
+# lib/Outer.rakumod:  use Inner; unit module Outer; class OuterClass is export { }
+
+use Outer;
+OuterClass.new;                # correct: Outer exported it
+Inner::InnerClass.new;         # rakudo: Could not find symbol.  mutsu: resolved
+```
+
+This is the last piece of issue #7555 item 1's divergence (b). The alias half of
+that divergence — the *classes and roles* a transitively-loaded module declares,
+and the variables and subs it exports — was closed by #7743 (issue #7692), which
+scopes a module's imports to its own compunit. Measured against that fix, one
+binding still escaped: the module's own package name.
+
+## Root cause
+
+`OpCode::RegisterPackage` binds a `unit module X` under its bare name with
+`env.insert`, and a module body runs in the **caller's** env
+(`load_module_inner` → `run_block`) rather than in an env of its own. So when
+`Outer`'s body executes its `use Inner;`, the `Inner` package binding lands in
+whatever frame triggered the load — the importing file, for a compile-time `use`
+at file scope. #7743's cleanup restores the caller's plain bindings and undoes
+the aliases `import_module` recorded, but a package name is neither: it is a new
+env key nobody registered as an import.
+
+It is now dropped at the end of the load that introduced it, unless it names the
+module being loaded or something nested under it. A direct `use Inner;` is
+unaffected — that load keeps its own package name, exactly as rakudo puts it in
+the importing scope's `MY::`.
+
+The registry diff is consulted alongside the name test, because a module file
+with no `unit` declarator can register a class under a name unrelated to its own:
+`roast/packages/RT128156/lib/RT128156/Top1.rakumod` declares a plain
+`class Top1`, which roast's `S10-packages/precompilation.t` requires to be
+visible in the loading scope's `MY::`. A name test alone drops it and fails that
+test.
+
+Pinned by `t/module-transitive-use-does-not-leak-types.t`, which passes 9/9
+verbatim under rakudo v2026.07 as well as under mutsu, and covers the parts
+#7743 fixed as well as this one — including the two-hop case (a module that
+`use`s a module that `use`s a third), whose method bodies resolve bare names
+through their own class's package chain rather than through `env`.
+
+## What is still divergent
+
+A module's `constant`s and enum values reach the importer the same way and are
+deliberately left alone. Removing them needs somewhere else for the module's own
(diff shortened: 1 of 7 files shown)


tokuhirom:
CI status on `31f5317`: the **`test`** job failed, and the failure is not this PR's.

```
error: could not fetch 691d0959ff8ea0a7c491167b366746cdfc0fb0be
       from https://git.sr.ht/~bduggan/raku-terminal-ansi
...
=== summary: 284/304 test files pass (2 excluded) ===
  REGRESSION: whitelisted 'Terminal::ANSI  01-basic.rakutest' did not pass this run
  REGRESSION: whitelisted 'Terminal::ANSI  02-capture.rakutest' did not pass this run
  REGRESSION: whitelisted 'Terminal::ANSI  03-call.rakutest' did not pass this run
  REGRESSION: whitelisted 'Terminal::ANSI  05-input.rakutest' did not pass this run
  REGRESSION: whitelisted 'Terminal::ANSI  99-meta.rakutest' did not pass this run
```

The gate could not **check out** the Terminal::ANSI battery from `git.sr.ht` — an external host — so all five of its whitelisted files were reported as regressed for want of a checkout, not for want of passing. No mutsu code ran for them. Every other battery is accounted for, and the five REGRESSION rows are exactly the five files of the dist that failed to fetch. (The `Log::Async` `FAIL` rows in the same output are pre-existing non-whitelisted rows, not regressions — the gate lists them every run.)

This PR's diff is `src/runtime/run_modules.rs`, five files under `t/`, and one under `news/`. It touches nothing in `scripts/battery-testsuite.sh`, `batteries.lock`, or `.github/`, and cannot affect a `git fetch` from sourcehut. There is no fix to port: the remedy is the host being reachable again.

Locally, on this exact tree, `make test` passes (3919 files / 41648 tests) and `make roast` shows no regression — its only failures are the three container-only ones `docs/agent-environments.md` records (`uid 0` `chmod` assertions ×2, sandboxed network ×1).

Re-running the failed job once to confirm, per the flake protocol. If it fails identically a second time, the sourcehut fetch is durably down and that is what is blocking this PR, not the change.

---

PR #7798, opened 2026-09-09
fix(proxy): an is-rw routine returns its Proxy as a container

### LINKED ISSUE #7748: `Proxy.new(FETCH/STORE)` cannot be assigned to — the whole type is unusable
`Proxy` is a documented core type whose entire purpose is to be bound to a
name and then read/written through the `FETCH`/`STORE` methods. In mutsu the
`STORE` side never runs: assigning through the bound name throws.

## Repro (the `Type/Proxy.rakudoc:17` synopsis example)

```
sub double() is rw {
    my $storage = 0;
    Proxy.new(
        FETCH => method ()     { $storage * 2    },
        STORE => method ($new) { $storage = $new },
    )
}
my $doubled := double();
$doubled = 4;
say $doubled;
```

- raku: `8`
- mutsu: `Cannot assign to an immutable value`

## Why it is `todo:deep`

`Proxy` is not a method that can be added in isolation — it is a *container*
whose read and write go through user code. Making it work means the assignment
path and the read path both have to recognise a Proxy-backed binding, i.e. it
touches the same container/`ContainerRef` machinery as the itemization work
(see #7542) rather than one dispatch site. It should be designed against that,
not bolted onto the assignment error path.

`Proxy` is also how several ecosystem modules implement lazy or computed
attributes, so this is a compatibility blocker, not a curiosity.

## Acceptance

- The repro prints `8`.
- `FETCH` is called on read (including in interpolation and as a sub argument)
  and `STORE` on assignment, each exactly once per access.
- Binding (`:=`) and `is rw` return both work; a Proxy returned from a method
  behaves the same as from a sub.
- A `t/` pin covering read, write, read-after-write, and use as an `is rw`
  routine return value.

tokuhirom:
`Proxy`'s whole purpose is to be bound to a name and then read and written through `FETCH`/`STORE`, and the `Type/Proxy.rakudoc:17` synopsis did not work at all:

```raku
sub double() is rw {
    my $storage = 0;
    Proxy.new(
        FETCH => method ()     { $storage * 2    },
        STORE => method ($new) { $storage = $new },
    )
}
my $doubled := double();
$doubled = 4;
say $doubled;      # raku: 8   mutsu: Cannot assign to an immutable value
```

## Root cause

Raku decontainerizes an ordinary `sub`'s return, so a `Proxy` built in its body reaches the caller already FETCHed. `is raw`, `is rw` and an explicit `return-rw` each suppress that: the caller receives the `Proxy` itself and the FETCH happens at the next *use*. That is what makes the synopsis work — the `:=` binds a container, so the later `=` has a `STORE` to run.

The return path tested `is raw` alone (`cf_auto_fetch = !cf.is_raw`), and the code-object path had the rule inverted outright (`auto_fetch = data.is_rw`, so an `is rw` routine was the *only* shape that FETCHed). Every `is rw` return was therefore resolved at the call, the binding held a plain value, and the assignment died as immutable.

## The fix

State the decision once per carrier, reading the whole ADR-0067 rule (`is rw`, `is raw`, **or** `return-rw`):

- **`Interpreter::routine_is_rw_capable` / `method_is_rw_capable`** wherever a resolved def was already in hand. This also brings the nine method-return auto-FETCH sites under the rule, so a `Proxy` returned from a method behaves as one returned from a sub. The two sites that no longer hold the def get `class_method_is_rw_capable`, an MRO walk that runs only once a `Proxy` is actually in hand.
- **`CompiledFunction::returns_container()`** for the compiled path. It carries no body AST, so it gets a new `uses_return_rw` precomputed at declaration time — which makes `sub f() { return-rw Proxy.new(...) }` bind a container too.
- **`SubData::returns_container()`** for the code-object path. It answers *container* for anything it cannot prove decontainerizes: a bare or pointy block returns raw in Raku anyway, and an anonymous `sub` is built from an `Expr::AnonSubParams`, which records `is_rw` but has no field for `is_raw` — so `sub () is raw { ... }` is indistinguishable there and keeps the pre-existing no-FETCH answer.

## Three bugs this uncovered, fixed here

**`try` did not evaluate its value.** With the container now reaching the caller, `(try dying-fetch-proxy()).defined` would FETCH *outside* the `try` and escape it. Rakudo evaluates a genuine `try`'s value inside the protected region — a throwing FETCH is caught and the `try` yields `Nil`, while an untroubled one still hands the `Proxy` back (`my $p := try f(); $p.VAR.^name` is `Proxy`). `TryCatch` now probes its body's value the same way, which also fixes the same escape for an `is raw` routine, where it was already reachable.

**A captured `Proxy` never ran `STORE`.** An assignment inside a block or a named sub reaches the variable by name (`SetGlobal`), never through a local slot, and that path replaced the captured container with the plain value instead of running `STORE` — the write silently vanished and the program carried on:

```raku
my $cell = 0;
my $p := Proxy.new(FETCH => method () { $cell }, STORE => method ($n) { $cell = $n });
my $b = { $p = 2 }; $b();      # raku: $cell is 2;  mutsu: still 0
sub s1() { $p = 3 }; s1();     # raku: $cell is 3;  mutsu: still 0
```

The store now checks for a `Proxy` before both write-throughs on that path (the compunit-lexical store and the generic `ContainerRef` one), each of which would otherwise rebind the name rather than write through it. This was live on `main` independently of the return rule; it stayed hidden because a bound name almost never held a real `Proxy` before.

**A subscript did not decontainerize a `Proxy` receiver.** This is XML's shape — `XML::Element::AT-POS` is `is rw` and returns a `Proxy` over the node list — and it is what the bundled-library gate caught on the first push (`XML/proxies.rakutest` and `XML/example.rakutest` fell below their baseline):

```raku
$xml[1][0].string      # No such method 'string' for invocant of type 'XML::Element'
```

The *second* subscript arrives holding the `Proxy` itself and indexed it as though it were a one-element list. `Index` FETCHes a `Proxy` receiver now, which is the decont raku performs on a subscript's invocant. Fusing the two levels on the *write* side (`$obj[i][j] = v` where `AT-POS` is `is rw`) had never worked, on `main` either — the chained-subscript store read the inner collection through the accessor and then had no way to descend past it — so when the inner instance has no `ASSIGN-POS`/`ASSIGN-KEY`, its own rw accessor is now asked for the location and the value stored through that.

## Tests

- New pin `t/proxy-rw-routine-return-bind.t` — 28 tests, verified to pass identically under rakudo: read, write and read-after-write through the binding; `FETCH`/`STORE` counted exactly once per access, including in interpolation and as a sub argument; a `Proxy` returned from a method; a plain `sub` still decontainerizing its return; `return-rw` without either trait; the `try` probe; a chained subscript reading and writing through an `is rw` `AT-POS`; and the three captured-`Proxy` store shapes.
- `t/thread-shared-scalar-visibility.t` loses its "STORE through a captured Proxy is visible" `todo` — a `start` block's write is one of the captured-name stores the second fix repairs.
- `cargo fmt`, `make lint` (all four configurations), `make test` (41816 tests), `make roast` and `scripts/battery-testsuite.sh` all run locally. The only failures are the documented container-only ones: roast's `6.c/S32-io/file-tests.t` + `S16-filehandles/filetest.t` under `uid 0` and `S32-io/IO-Socket-Async.t` under the network sandbox (exact subtest ranges as `docs/agent-environments.md` records), and the battery gate's six DBIish MySQL files with no `libmysqlclient` installed. XML is clean in the gate.

Closes #7748.

https://claude.ai/code/session_01Djo5f9tJ1zhSi7agc1NRMK

### DIFF
--- a/news/2026-09/proxy-rw-routine-return-container.md
+++ b/news/2026-09/proxy-rw-routine-return-container.md
@@ -1,0 +1,92 @@
+# `Proxy` returned from an `is rw` routine now reaches the caller as a container
+
+`Proxy` is the documented core type whose whole purpose is to be bound to a
+name and then read and written through its `FETCH`/`STORE` methods. The
+`Type/Proxy.rakudoc` synopsis — an `is rw` sub returning a `Proxy`, bound with
+`:=` and then assigned to — did not work in mutsu at all:
+
+```raku
+sub double() is rw {
+    my $storage = 0;
+    Proxy.new(
+        FETCH => method ()     { $storage * 2    },
+        STORE => method ($new) { $storage = $new },
+    )
+}
+my $doubled := double();
+$doubled = 4;
+say $doubled;      # raku: 8   mutsu: Cannot assign to an immutable value
+```
+
+## Root cause: the wrong half of the decontainerization rule
+
+Raku decontainerizes a routine's return value, so a `Proxy` built inside an
+ordinary `sub` reaches the caller already FETCHed. `is raw` and `is rw` both
+suppress that: the caller receives the `Proxy` itself and the FETCH happens at
+the next *use*. That is what makes the synopsis work — the `:=` binds a
+container, so the later `=` has a `STORE` to run.
+
+mutsu's return path tested `is raw` alone (`cf_auto_fetch = !cf.is_raw`), and
+the code-object path had the rule inverted outright (`auto_fetch = data.is_rw`,
+so an `is rw` routine was the *only* shape that FETCHed). Every `is rw` return
+was therefore resolved at the call, the binding held a plain value, and the
+assignment died as immutable.
+
+The decision is now stated once per carrier and reads the whole ADR-0067 rule
+— `is rw`, `is raw`, **or** an explicit `return-rw`:
+
+- ADR-0067's existing `Interpreter::routine_is_rw_capable` /
+  `method_is_rw_capable` wherever a resolved def was already in hand;
+- `CompiledFunction::returns_container()` for the compiled path, which carries
+  no body AST and so gets a new `uses_return_rw` computed once at declaration
+  time (so `sub f() { return-rw Proxy.new(...) }` binds a container now too);
+- `SubData::returns_container()` for the code-object path.
+
+`SubData` answers *container* for anything it cannot prove decontainerizes: a
+bare or pointy block returns raw in Raku anyway, and an anonymous `sub` is built
+from an `Expr::AnonSubParams`, which records `is_rw` but has no field for
+`is_raw` — so `sub () is raw { ... }` is indistinguishable there from a plain
+one and keeps the pre-existing no-FETCH answer.
+
+## Two bugs the fix uncovered
+
+**`try` did not evaluate its value.** With the container now reaching the
+caller, `(try dying-fetch-proxy()).defined` would FETCH *outside* the `try` and
+escape it. Rakudo evaluates a genuine `try`'s value inside the protected region
+— a throwing FETCH is caught and the `try` yields `Nil`, while an untroubled one
+still hands the `Proxy` back (`my $p := try f(); $p.VAR.^name` is `Proxy`). The
+`TryCatch` op now probes its body's value the same way, which also fixes the
+same escape for an `is raw` routine, where it was reachable before this change.
+
+**A captured `Proxy` never ran `STORE`.** An assignment inside a block or a
+named sub reaches the variable by name (`SetGlobal`), never through a local
+slot, and that path replaced the captured container with the plain value instead
+of running `STORE` — the write silently vanished and the program carried on:
+
+```raku
+my $cell = 0;
+my $p := Proxy.new(FETCH => method () { $cell }, STORE => method ($n) { $cell = $n });
+my $b = { $p = 2 }; $b();      # raku: $cell is 2;  mutsu: still 0
+sub s1() { $p = 3 }; s1();     # raku: $cell is 3;  mutsu: still 0
+```
+
+The store now checks for a `Proxy` before both write-throughs on that path (the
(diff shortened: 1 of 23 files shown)


tokuhirom:
## `jit-stress` red on `roast/S17-procasync/kill.t` test 7 — not this PR's, re-running once

`jit-stress` failed with exactly one real failure ([job](https://github.com/tokuhirom/mutsu/actions/runs/34378809774/job/102558297989)); everything else in that job's roast phase was green, and the crash report it uploaded is the allowlisted deliberate `strdup(0)` one.

```
roast/S17-procasync/kill.t  (Wstat: 256 (exited 1) Tests: 8 Failed: 1)
  Failed test:  7
```

Test 7 is `doesn't-hang { … 4 × Proc::Async … .kill … }, :10wait, '.kill kills when multi-procs kill in multi-promises'` — a wall-clock budget over four spawned processes killed from four promises. Its failure mode *is* a contended runner, and `S17-*` under heavy parallel load is the category CLAUDE.md already documents as load-sensitive. It is not in `flaky-tests.txt`, so the retry harness never re-ran it ("No quarantine retries were needed in this job").

Evidence it is not this PR's:

- **8 clean local runs in the exact jit-stress configuration** (`MUTSU_JIT=on MUTSU_JIT_THRESHOLD=2`, release binary): 5× `kill.t` serially, plus 3 full sweeps of all 99 whitelisted `S17-*` files at `-j4` — 1603 tests, PASS every time.
- **`jit-stress` was green on the previous head of this same branch** (4b8c2dc), whose only difference is the subscript-decont commit — and that commit touches `Index`'s receiver and the chained-subscript store, neither of which has any path into `Proc::Async.kill`.
- `make roast` and `make test` are green locally on this head, apart from the three documented container-only files.

Re-running the failed job once, per the flaky-triage rule. If it comes back red on the same subtest I will treat it as real and root-cause it rather than re-run again.

---

Issue #7803, opened 2026-09-09
A `unit module`'s exported `sub` is not callable via `Module::sub-name()` qualified-call syntax


tokuhirom:
Found while extending `t/module-transitive-use-does-not-leak-types.t` for #7797.

## Repro

```raku
# t/lib/TransitiveLeakOuter.rakumod (existing fixture)
unit module TransitiveLeakOuter;
sub outer-uses-inner() is export { ... }
```

```raku
use lib 't/lib';
use TransitiveLeakOuter;
say TransitiveLeakOuter::outer-uses-inner();
```

mutsu:

```
Could not find symbol '&outer-uses-inner' in 'TransitiveLeakOuter'
```

`raku` resolves this fine (a `use`d module's exported sub is always callable package-qualified, in addition to bare). This reproduces identically on `main` (verified via `git stash` before any #7797 changes were applied), so it is unrelated to that issue's package-qualified-name visibility gate — a plain `sub NAME() is export { ... }` (not `our sub`) inside a `unit module` apparently never gets registered so that `Module::NAME(...)` call syntax can find it, even from a compunit that legitimately `use`d the module.

Contrast: `our sub NAME is export { ... }` (used by `t/lib/InnerConst.rakumod`-style fixtures elsewhere) does not hit this — only the plain `sub ... is export` shape seems affected. Not fully root-caused; whoever picks this up should start by comparing how `register_sub_decl_with_metadata`/`register_exported_sub` register a plain `sub is export` vs an `our sub is export` inside a `unit module` body, and why only one form ends up reachable through `has_function`/`resolve_function_with_types` under the qualified key.

## Affected files (starting points)

- `src/runtime/run_modules.rs` (`load_module_inner`) — module body registration
- `src/runtime/registration_sub.rs` — `register_sub_decl_with_metadata`
- `src/vm/vm_call_func_ops.rs` (`exec_call_func_op_inner`) — where the qualified call lookup fails

## Verification bar

A regression test calling `Module::exported-plain-sub()` (not `our sub`) from a compunit that `use`d the module, expecting it to work exactly as the bare call does.

tokuhirom:
Claiming: claude/mutsu-issue-7803-qifb5e

---
_Generated by [Claude Code](https://claude.ai/code)_

tokuhirom:
[...]

Issue #7612, opened 2026-09-08
A nested module's import is visible to the USING scope (`GLOBAL::` is flat)


tokuhirom:
Split off 2026-09-08 from #7580, which fixed the *disappearance* of a nested
module's import. This is the opposite edge of the same flat-registry model and
is **pre-existing**, not introduced by that fix: it reproduces on plain
top-level `use` with no block anywhere.

## Repro

```raku
# t/lib/BlockUseNestedInner.rakumod: unit module BlockUseNestedInner; use NativeCall; ...
# t/lib/BlockUseNestedOuter.rakumod: unit module BlockUseNestedOuter; use BlockUseNestedInner; ...

use lib 't/lib';
use BlockUseNestedOuter;
say defined(::('&nativecast'));
# raku:  False
# mutsu: True
```

`BlockUseNestedOuter` exports `outer-probe` and nothing else, yet the routine
that a module *two levels down* imported for its own body is resolvable in the
using scope.

## Root cause

mutsu's routine registry is flat and keyed by name. A `unit module`'s body runs
at `current_package() == GLOBAL`, so its own `use X` registers `GLOBAL::<name>`
— the same key a top-level import would use. There is no per-compilation-unit
lexical pad to hold "this alias belongs to `BlockUseNestedInner`'s scope only",
so once the entry is in the registry it is visible everywhere.

#7580 had to retain exactly these entries (dropping them left the module
half-loaded, since `loaded_modules` is never rolled back), which makes the
block-scoped case consistent with the already-leaking plain case. Fixing the
leak is the other half and needs real lexical scoping for imports, not another
retain rule.

## Why it is deep

The fix is a scoping model, not a predicate: routine resolution would have to
consult the *compilation unit* a routine is being resolved from, so
`GLOBAL::nativecast` installed while `BlockUseNestedInner` was loading is
reachable from `BlockUseNestedInner`'s routines and from nowhere else. That
touches module loading, `pop_import_scope`, `reinstate_module_functions`, and
every by-name routine lookup. It should be designed (an ADR) before it is coded.

## Check when fixing

- The repro above, and the same with `{ ... }` around the `use`.
- `roast/S11-modules/lexical.t` (`{ use Foo }` must leave `foo()` unresolvable)
  and `roast/S11-modules/require.t` test 10 (`{ require NoModule <&bar>; }` must
  not leak `&bar`), the two tests that pin the direction that already works.
- `t/block-use-keeps-nested-module-imports.t` and
[...]

PR #7808, opened 2026-09-09
fix: gate package-qualified names by which compunit `use`d the package

### LINKED ISSUE #7797: Package-qualified names of a transitively-loaded module resolve from a compunit that never `use`d it
The storage half of #7555 item 1's divergence (b), left open deliberately by
#7764 and again by #7791 (point 3 of that issue's "What a fix would have to
establish"). The **visibility** half is now closed — #7743 (imports, classes,
roles), #7764 (package names), #7791 (`constant`s and enum values) — so no *bare*
name of a transitively-loaded module reaches the importer any more. Every one of
those names is still reachable **package-qualified**, from any compunit in the
process.

## Repro

```raku
# lib/InnerConst.rakumod
unit module InnerConst;
constant INNER-CONST = 42;
enum InnerEnum <ALPHA BETA>;
class InnerClass is export { }
our sub inner-reads-const is export { INNER-CONST ~ '/' ~ ALPHA.key }
```
```raku
# lib/OuterConst.rakumod
use InnerConst;
unit module OuterConst;
constant OUTER-CONST = 7;
```
```raku
# probe.raku -- the importer says only `use OuterConst`, never `use InnerConst`
use MONKEY-SEE-NO-EVAL;
use OuterConst;
sub probe($code) {
    my $r = try EVAL $code;
    say $code.fmt('%-34s') ~ ' => ' ~ ($! ?? 'MISSING' !! ($r // 'Nil').gist);
}
probe($_) for
    'OuterConst::OUTER-CONST',
    'InnerConst::INNER-CONST',
    'InnerConst::ALPHA.key',
    'InnerConst::InnerClass.new.^name',
    'InnerConst::InnerEnum.^name',
    'InnerConst::inner-reads-const()';
```

| expression | raku v2026.07 | mutsu 5721e1e |
| --- | --- | --- |
| `OuterConst::OUTER-CONST` (the module it *did* `use`) | 7 | 7 |
| `InnerConst::INNER-CONST` | MISSING | **42** |
| `InnerConst::ALPHA.key` | MISSING | **ALPHA** |
| `InnerConst::InnerClass.new.^name` | MISSING | **InnerConst::InnerClass** |
| `InnerConst::InnerEnum.^name` | MISSING | **InnerEnum** |
| `InnerConst::inner-reads-const()` | MISSING | **42/ALPHA** |

It is not one symbol kind: constants, enum values, enum types, classes and subs
are all reachable. Practical consequence — **a program that forgets a `use` runs
fine under mutsu and dies under rakudo**, as long as something else in the
dependency graph loaded that module. That is the same "green for the wrong
reason" hazard #7714's closing note flagged.

### Control (run it first — the naive probe gives a false negative)

The same file written with `::('InnerConst::InnerClass')` instead of `EVAL`
reports MISSING under *both* implementations, which reads like correct scoping
and is not: `::()` simply does not find a class this way. The control is the same
probe from a file that **does** `use InnerConst`:

| expression, importer did `use InnerConst` | raku | mutsu |
| --- | --- | --- |
| `InnerConst::INNER-CONST` | 42 | 42 |
| `InnerConst::ALPHA.key` | ALPHA | ALPHA |
| `InnerConst::InnerClass.new.^name` | InnerConst::InnerClass | InnerConst::InnerClass |
| `InnerConst::InnerEnum.^name` | InnerEnum | InnerEnum |
| `InnerConst::inner-reads-const()` | 42/ALPHA | 42/ALPHA |

An exact match, the rakudo-side `InnerEnum` `.^name` quirk included. So mutsu is
right about *what* these names mean and wrong only about *who can see them* —
which is what makes this a scoping problem and not a resolution one.

## Root cause

mutsu has no per-compunit symbol table for package symbols. They live in two
process-global stores:

- **`our`-scoped variables and `constant`s** — `self.env`, under their
  fully-qualified key. The end-of-load cleanup in `load_module_inner` is explicit
  about leaving them: *"Restore every plain caller binding … Qualified package
  globals and newly-created module names are intentionally left alone"*
  (`src/runtime/run_modules.rs:982-987`). Every cleanup added by the three
  visibility slices filters on `!key.contains("::")` for exactly this reason.
- **Classes and roles** — the single `registry().classes` / `registry().roles`
  maps, keyed by qualified name, with no notion of who may see an entry.

rakudo instead installs a `use`d package into the importing compunit's `MY::`,
so a compunit that never asked for `InnerConst` has no path to it at all.

## Why it is large

- **It needs a store that does not exist yet.** The three visibility slices could
  each *delete* an env binding at end of load; this one cannot delete anything —
  the symbols must stay resolvable for their own compunit and for whoever `use`d
  them. It needs a per-compunit "packages visible here" set consulted by
  qualified-name resolution, i.e. a new dimension on two of the hottest lookup
  paths.
- **The compunit identity exists but its coverage is unproven.**
  `unit_of_source` (`src/vm/vm_helpers.rs:28`, `src/runtime/compunit_scope.rs:63`)
  and the `unit_lexicals` store (`src/runtime/mod.rs:2582`) are the precedent, but
  they are consulted from a handful of sites. A qualified-name gate has to be
  right for OTF-compiled method bodies, `EVAL`, a runtime `require`, and the
  per-thread clones (`src/runtime/runtime_thread.rs:691-692`) — anywhere the
  "which compunit is running" answer is currently approximate, this turns a wrong
  answer into a hard "Could not find symbol".
- **Every previous slice of this divergence looked like a one-line ownership test
  and was not.** #7764 records the case: a name-prefix test alone drops a class a
  no-`unit` module registered under an unrelated name
  (`roast/packages/RT128156/lib/RT128156/Top1.rakumod`, which
  `S10-packages/precompilation.t` tests 40-41 require in the *loading* scope's
  `MY::`) and turns two green roast tests red.
- **Tightening this will expose code that currently gets away with a missing
  `use`** — mutsu's own bundled batteries and `t/` suite included. That is the
  point of the change, but it means the blast radius is "however many places in
  the tree rely on the leak", which is unknown until it is tried.

## Affected files

- `src/runtime/run_modules.rs` — `load_module_inner`, the end-of-load cleanup
  (the `!contains("::")` filters at 841-846 and 982-987)
- `src/runtime/types/type_registry.rs` — qualified-name resolution
  (`has_type_direct`, `lookup_in_package_chain`)
- `src/runtime/compunit_scope.rs`, `src/vm/vm_helpers.rs:28` — the existing
  compunit identity
- `src/runtime/mod.rs:2582` — `unit_lexicals`, the per-compunit store to model a
  new one on
- `src/runtime/runtime_thread.rs:691-692` — per-thread clones of the tables

## What a fix would have to establish

1. Where the "packages visible from this compunit" set lives, and whether it is
   keyed by `unit_of_source` or by something that also survives OTF compilation
   and thread spawn.
2. What a *miss* does. rakudo's answer is a compile-time "Undeclared name"; mutsu
   resolves qualified names at runtime, so the honest equivalent is a runtime
   `X::NoSuchSymbol` — which turns today's silent success into a loud failure in
   any code that relies on the leak.
3. Whether the class/role registry gets the same gate as `env`, or whether types
   are scoped by a different mechanism.
4. How much of the existing tree depends on the leak. Measure that first — a
   survey of `t/`, the bundled batteries and roast under an experimental gate is
   the cheap way to size the change before committing to it.

## Verification bar

`t/module-transitive-use-does-not-leak-types.t` (27 assertions after #7791) pins
the visibility half and passes verbatim under rakudo; extend it rather than
adding a parallel file. `t/log-async-battery.t` and
`roast/S10-packages/precompilation.t` are the two regressions previous slices of
this divergence actually hit.

tokuhirom:
## Summary

Closes #7797.

- A package-qualified name (`Pkg::name`) reached a class, role, enum,
  `our`-scoped constant/variable, or sub of ANY module ever loaded into the
  process, even from a compunit that never `use`d/`need`d/`require`d that
  module itself — only transitively, through a module it DID `use`. Rakudo
  installs a `use`d package into the *importing* compunit's `MY::` only, so
  such a name is undeclared there. This closes the package-qualified half of
  the divergence left open by #7743/#7764/#7791 (which already closed the
  bare-name half).
- Two new per-compunit stores — `package_declaring_units` (which compunit a
  `use`d top-level package belongs to) and `compunit_visible_packages` (which
  such packages a given compunit earned visibility to by using them
  directly) — populated at the end of a successful module load (both the
  first load and a re-`use` of an already-loaded module, which used to skip
  the grant entirely). `Interpreter::qualified_name_visible_here` consults
  both, walking the same `EVAL`-parent chain `Interpreter::prelude_visible_here`
  already uses for the analogous NativeCall-prelude-splice visibility gate
  (#7612). Wired into the two chokepoints where a source-written qualified
  name resolves: the bareword term path (`exec_get_bare_word_op`) and the
  qualified function-call path (`exec_call_func_op_inner`).
- Getting the gate right also exposed a real, separate bug: while a module's
  own top-level mainline runs (`load_module_inner`'s `run_block`, which
  pushes no routine frame), `Interpreter::executing_unit_sym` can
  misattribute "who is executing" to a stale, unrelated routine frame still
  on `routine_stack` (e.g. `DBIish.install-driver` doing a dynamic
  `require ::($module)`). `module_loading_unit_stack` + the new
  `executing_unit_sym_for_module_load` fix that for both the visibility gate
  and the importer-capture in `load_module_inner`.
- A package whose only declaring compunit is unknown (a same-compunit
  `package Foo { }` block, or a script's own top-level `unit module Foo`,
  neither of which goes through `load_module_inner`) stays permissive — this
  gate only ever restricts cross-compunit reach, never a compunit's view of
  its own declarations.

Filed #7803 for an unrelated pre-existing gap found while extending the
pinned test: a `unit module`'s plain `sub NAME() is export` is not callable
via `Module::NAME()` qualified-call syntax at all (reproduces identically on
`main`, unrelated to this change).

## Test plan

- [x] Extended `t/module-transitive-use-does-not-leak-types.t` (the existing
      pin for the bare-name half of this divergence) with package-qualified
      assertions, per the issue's verification bar, rather than adding a
      parallel file.
- [x] `cargo fmt --all`
- [x] `make test` — full `t/` suite, 3934 files / 41,855 assertions — all pass
- [x] `make roast` — 1436 files / 218,939 assertions; the only failures
      (`roast/6.c/S32-io/file-tests.t`, `roast/S16-filehandles/filetest.t`,
      `roast/S32-io/IO-Socket-Async.t`) are the documented container-only
      discrepancies (running as `uid 0`, sandboxed network — see
      `docs/agent-environments.md`), not regressions from this change.
- [x] Manually verified the issue's own repro table matches `raku` exactly,
      before and after `use InnerConst;` is added to the importer.

https://claude.ai/code/session_01JvKe6HdquqVhYX1fJYSeMD

---
_Generated by [Claude Code](https://claude.ai/code/session_01JvKe6HdquqVhYX1fJYSeMD)_

### DIFF
--- a/news/2026-09/package-qualified-name-transitive-use-leak.md
+++ b/news/2026-09/package-qualified-name-transitive-use-leak.md
@@ -21,41 +21,84 @@ into the *importing compunit's* `MY::` only, so a compunit that never `use`d
 `InnerConst` itself has no path to `InnerConst::anything`, even through a
 module it did `use` that in turn `use`d `InnerConst`.
 
+## The visibility rule turned out to have three parts, not one
+
+The naive fix — a compunit sees `Pkg::x` only if it `use`d `Pkg` itself —
+passes the issue's own repro and the local `t`/`roast` suites, but fails
+`scripts/battery-testsuite.sh` (the gate that runs 17 real, vendored
+distributions' own upstream test suites): it broke Cro::HTTP, HTTP::UserAgent,
+XML, OpenSSL, and DBIish's mysql driver, all in the same shape. Chasing that
+down against real `raku` (not just the issue's own repro) turned up two more
+genuinely-permissive rules mutsu had been relying on by accident:
+
+1. **Direct grant.** A compunit that `use`/`need`/`require`d `Pkg` directly
+   may reference `Pkg::anything` — the rule the issue asks for.
+2. **Package-ancestor.** Code running with current package `Foo` may
+   reference `Foo::Bar::anything` even without `use`ing `Foo::Bar` itself —
+   confirmed against the real, vendored `OpenSSL` module: `unit class
+   OpenSSL`'s `method new` calls `OpenSSL::Ctx::SSL_CTX_new(...)` without
+   ever `use`ing `OpenSSL::Ctx` (only transitively, through its own `use
+   OpenSSL::SSL`). A minimal repro with an unrelated current package
+   (`unit class Outer; use Bar; method probe { Baz::greet() }`, where
+   `Bar.rakumod` does `use Baz;`) correctly fails under real `raku` — the
+   difference is the shared name, not the `use` chain.
+3. **Same top-level namespace.** A compunit that `use`d *anything* under a
+   given top-level `::`-segment may reference *any other* package under that
+   same segment — confirmed with `IO::Socket::Async::SSL.rakumod` (no `unit`
+   declarator, so rule 2 cannot apply): a top-level `my constant ... =
+   OpenSSL::Version::version_num() ...;` resolves under real `raku` even
+   though the file's own `use` list never mentions `OpenSSL::Version`, only
+   sibling packages (`OpenSSL`, `OpenSSL::Bio`, `OpenSSL::Ctx`, ...) under the
+   same `OpenSSL::` prefix.
+
+None of this is written down anywhere; it was reverse-engineered by running
+real `raku` against both the failing vendored modules and minimal repros
+until each rule's exact boundary was pinned down.
+
 ## The fix
 
-Two new per-compunit tables, mirroring the existing `prelude_declaring_units`/
-`prelude_visible_here` mechanism built for the analogous NativeCall-prelude
-visibility gate (#7612):
+`Interpreter::qualified_name_visible_here` checks all three, in order, at the
+two chokepoints where a source-written qualified name resolves — the
+bareword term path (`exec_get_bare_word_op`) and the qualified function-call
+path (`exec_call_func_op_inner`):
 
-- `package_declaring_units`: which compunit a `use`/`need`/`require`d
-  top-level package belongs to (first `::`-segment granularity, matching the
-  coarseness the three earlier bare-name slices already use).
-- `compunit_visible_packages`: which such packages a given compunit earned
-  visibility to, by `use`ing them *directly* — populated at the end of a
-  successful module load, whether that's the first load (`load_module_inner`)
-  or a re-`use` of an already-loaded module (which used to skip the grant
-  entirely, since nothing new needs registering there for the OLDER
-  bare-name/class-alias mechanisms).
+- `current_package_is_ancestor_of` (rule 2) — a pure string ancestor-walk of
+  `self.current_package()`, no bookkeeping needed.
+- `package_granted_in_unit_chain` (rule 3) — a coarse top-level-segment
+  membership check against `compunit_visible_packages`.
+- `longest_declared_package_prefix` + `package_visible_in_unit_chain` (rule
+  1, plus a compunit's unconditional view of its own declarations) — an
+  exact-prefix match against `package_declaring_units`.
 
-`Interpreter::qualified_name_visible_here` consults both, walking the `EVAL`
-parent chain exactly as `prelude_visible_here` does, and gates the two
-chokepoints where a source-written qualified name resolves: the bareword term
-path (`exec_get_bare_word_op` — constants, enum values/types, classes, roles)
-and the qualified function-call path (`exec_call_func_op_inner`).
+Two new per-compunit tables populated at the end of every successful module
+load (mirroring `prelude_declaring_units`/`prelude_visible_here`, the
+existing NativeCall-prelude visibility gate from #7612):
 
-## The harder half: which compunit is "running" during a module load
+- `package_declaring_units`: **full package name** → declaring compunit.
+  Deliberately NOT truncated to a first `::`-segment: a real multi-file
+  distribution routinely has several unrelated compunits sharing a namespace
+  prefix (`XML::Entity` and `XML::Element` are separate `unit class`-scoped
+  files, both under `XML::`) — keying this by `"XML"` let whichever loaded
+  first claim the whole prefix and made every sibling's OWN qualified
+  self-reference to its own name look foreign (`XML::Entity.rakumod`
+  referencing `XML::Entity.new` inside its own body).
+- `compunit_visible_packages`: for each compunit, both the full names AND the
+  top-level segments of everything it `use`d directly (rules 1 and 3) —
+  populated on first load and, since a re-`use` of an already-loaded module
+  skips `load_module_inner` entirely, on the already-loaded fast path too.
 
-Getting the *visibility* gate right exposed a second, sharper bug in how
-mutsu decides "which compunit is executing right now" while a module's own
+## Which compunit is "running" during a module load
+
+Getting all of this right also exposed a sharper, separate bug in how mutsu
+decides "which compunit is executing right now" while a module's own
 top-level mainline is running. `Interpreter::executing_unit_sym` prioritizes
 the topmost `routine_stack` frame — correct for an ordinary call, but wrong
 for a module body, which runs via `run_block` and pushes no frame of its own.
 `DBIish.install-driver('SQLite')` does a dynamic `require ::($module)` from
-inside a method; the loaded module's own top-level `use NativeLibs;` (and a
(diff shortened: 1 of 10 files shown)


tokuhirom:
The `test` check failed on `t/whenever-callback-recompile-semantics.t` subtest 5 ("a nested whenever inside the callback fires once per value"), with an emit-order mismatch (`10 30 20` instead of `10 20 30`).

Not this PR's failure — this PR touches only module-load/qualified-name visibility, no concurrency code. Verified locally: the subtest is a genuine race (three async `whenever`/`Promise.keep` continuations racing with no ordering guarantee in the test), reproducing on `main` (7470841) too — 20 local repeats each: main 5/20 fail, this branch 9/20 fail. Different enough sample to look like noise, not a regression (an earlier full-suite run on this branch, under different system load, passed the file cleanly). Filed #7811 to track and fix the flaky test itself; no fix exists yet.

Re-running the `test` check once per the flaky-failure allowance.

---

tokuhirom:
Root-caused and fixed the `test` job failure (the "bundled library gate" — `scripts/battery-testsuite.sh`, which runs 17 real vendored distributions' own upstream test suites; this is not exercised by `make test`/`make roast`).

The naive fix (visible only if the compunit itself `use`d the package) broke Cro::HTTP, HTTP::UserAgent, XML, OpenSSL, and DBIish's mysql driver — all in the same shape. Verified against real `raku` (not just this issue's own repro) that the actual visibility rule has three parts:
1. Direct `use` grant (what the issue asks for).
2. Package-ancestor: code running with current package `Foo` may reference `Foo::Bar::anything` without `use`ing `Foo::Bar` itself — confirmed with the real vendored `OpenSSL` module (`OpenSSL.rakumod`'s `method new` calls `OpenSSL::Ctx::SSL_CTX_new` without ever `use`ing `OpenSSL::Ctx`, only transitively).
3. Shared top-level namespace: a compunit that `use`d anything under a top-level `::`-segment may reference any other package under that segment — confirmed with `IO::Socket::Async::SSL.rakumod`, which reads `OpenSSL::Version::version_num()` in a top-level `constant` without ever `use`ing `OpenSSL::Version` itself.

Also fixed a real bug the first version of the fix exposed: `package_declaring_units` was keyed by a truncated first `::`-segment, so two unrelated files sharing a namespace prefix (`XML::Entity`/`XML::Element`, both under `XML::`) collided — whichever loaded first claimed the prefix, making the other's own qualified self-reference to its own name look foreign.

Locally validated: `make test` (41,855 assertions), `make roast` (clean except the three documented container-only discrepancies), and `scripts/battery-testsuite.sh` (`GATE PASSED: every whitelisted bundled-library test still passes` — 289/312 files, 2 excluded) all green. Pushed. Re-running the failed `test` check now.

Full writeup in `news/2026-09/package-qualified-name-transitive-use-leak.md`.

---

tokuhirom:
The main `test` job (including the bundled-library gate) is green. `jit-stress` and `gc-stress` each failed on a single, unrelated subtest:

- `jit-stress`: `roast/S32-list/seq.t` test 38 — `Seq` iteration, unrelated to this PR's module-loading/qualified-name-visibility diff. Reproduces 0/5 locally under identical `MUTSU_JIT=on MUTSU_JIT_THRESHOLD=2` env vars (release binary).
- `gc-stress`: `roast/S17-procasync/kill.t` test 7 — a process/signal-kill test in the `S17-procasync` family, explicitly documented in `CLAUDE.md` as occasionally flaky under heavy parallel load.

Both stress jobs use deliberately aggressive settings (`MUTSU_JIT_THRESHOLD=2`, `MUTSU_GC_EVERY_CANDIDATE=1024`) precisely to surface timing-sensitive issues; neither failure touches code this PR changes. Re-running both once.

---

PR #7810, opened 2026-09-09
Fix `need` suppressing exports its own compunit imports via `use`

### LINKED ISSUE #7805: `need Foo` loses the compunit's own imports for `Foo`'s methods
A module loaded with `need` (which loads without importing) cannot resolve, from
its own methods, the subs its own compunit imported. `use` is unaffected.

Found by running the bundled-library gate under the vendored `Test` module
(`MUTSU_REAL_TEST=1`) on `main` at 5721e1e. It is **not** a `Test` bug — the
native TAP provider merely masks it, because there `diag` is a builtin that
resolves anywhere.

## Repro

```raku
# lib/CT.rakumod
use Test;

unit class CT;

method run { diag "from a method"; }
```
```raku
# t2.raku
need CT;
CT.new.run;
```

| | result |
| --- | --- |
| `raku -I lib t2.raku` (v2026.07) | `# from a method` |
| `mutsu -I lib t2.raku` (native provider) | `# from a method` |
| `MUTSU_REAL_TEST=1 mutsu -I lib t2.raku` | **`Unknown function: diag`** (`in sub run at lib/CT.rakumod line 5`) |

Two controls that do **not** reproduce, so the trigger is precise:

- `use CT;` instead of `need CT;` — works under all three, even with no
  `use Test` in the main program.
- `need CT;` where the main program *also* says `use Test;` — works.

So the failing shape is specifically: the compunit imported a sub, the loading
program imported nothing, and the resolution happens inside a method of that
compunit's `unit class`.

## Impact on the gate

Four whitelisted `DBIish` files fail on this and nothing else:
`35-pg-common`, `55-oracle-common`, `70-sqlcipher-memory`, `71-sqlcipher-common`.
Each is a two-line file that does

```raku
need DBIish::CommonTesting;
DBIish::CommonTesting.new(...).run-tests;
```

and `DBIish::CommonTesting` is `use Test; use DBIish; unit class DBIish::CommonTesting;`
whose `method run-tests` opens with `diag "Testing DBDish::$.dbd"`
(`modules/DBIish/lib/DBIish/CommonTesting.rakumod:1,48-49`). Under the vendored
provider the method dies there before emitting a plan, so the file reports
`FAIL(ok=0/?)`.

### This corrects #7555

#7555 records these rows under *"The `DBIish` rows are environment, not this"* —
"they need live database servers and cannot be judged from a container without
them", with the next step being to run the provider A/B on a runner that has the
services. That A/B has now been run on a container that has **no** database
servers at all, and they still move:

| DBIish file | native | vendored |
| --- | --- | --- |
| `35-pg-common` | PASS | FAIL(ok=0/?) |
| `55-oracle-common` | PASS | FAIL(ok=0/?) |
| `70-sqlcipher-memory` | PASS | FAIL(ok=0/?) |
| `71-sqlcipher-common` | PASS | FAIL(ok=0/?) |
| the six `2x-mysql-*` rows | FAIL | FAIL |

They pass under the native provider *because* there is no server: DBIish's own
`connect-or-skip` calls `skip-rest 'prerequisites failed'`, which satisfies the
whole 109-test plan. So no server is needed to judge them — only the provider
A/B. (The six `2x-mysql-*` rows really are environment: this container has no
`libmysqlclient`, so `NativeCall: symbol 'mysql_init' not found` kills them
identically under both providers.)

## Not a regression from the module-scoping campaign

Bisected: it already fails at 398009f, the commit **before** #7743
(`d7a6906`), so #7743 / #7764 / #7791 did not introduce it. It is a
pre-existing `need` gap that only the vendored provider makes visible.

## Where to look

`need` is a load-without-import, so the module's compunit gets no `import_module`
record in the loading scope, and the machinery that lets a module's own routines
resolve their imports lexically — `module_scope_lexicals` /
`module_imported_lexical_names` (`src/runtime/mod.rs:2412-2436`), consulted via
`lookup_in_running_package` (`src/runtime/types/type_registry.rs:504`) — is
populated from `module_imported_names`, which `import_module` accumulates. The
`use` path works, so the likely gap is that the `need` path skips whatever
records those names against the module. `lookup_in_running_package` also
deliberately skips a frame package of `GLOBAL`, which is worth checking for a
method frame here.

## Verification bar

A `t/` pin in the shape of the repro above (it passes verbatim under rakudo), and
the four DBIish rows going green under `MUTSU_REAL_TEST=1` in
`scripts/battery-testsuite.sh`.

tokuhirom:
## Summary

- `need Foo;` sets `suppress_exports` for the whole duration of loading `Foo`, so `Foo`'s own `is export` subs never register as importable into the loading scope — that is the whole point of `need`. But the flag was a single unscoped boolean, so it stayed set for every nested `use` statement `Foo`'s own body executed too. A `Foo.rakumod` that itself said `use Bar;` therefore had `Bar`'s exports silently dropped from `exported_subs`, so `Foo`'s own `use Bar;` found nothing to import, and any of `Foo`'s methods calling a `Bar` routine died with `Unknown function`.
- `Interpreter::use_module_with_tags` (`src/runtime/runtime_module.rs`) now suspends `suppress_exports` for the duration of its own nested load, restoring the ambient value afterward — an explicit `use` always wants ordinary export semantics, regardless of whether it happens to run inside a `need`-loaded compunit's own mainline.
- Found via the bundled-library gate running under the vendored `Test` module (`MUTSU_REAL_TEST=1`): four whitelisted `DBIish` files (loaded through `need DBIish::CommonTesting;`, whose own `method run-tests` opens with `diag "..."` and which itself does `use Test;`) failed with `Unknown function: diag`. Verified locally (no database server needed) that all four now pass under the vendored provider: `35-pg-common`, `55-oracle-common`, `70-sqlcipher-memory`, `71-sqlcipher-common` (109/109 each).

Closes #7805.

## Test plan

- [x] New regression test `t/need-preserves-nested-use-imports.t` (+ fixture modules under `t/lib-need-nested-import/`): a `need`-loaded class whose own `use` of a sibling module must resolve from one of its methods. Confirmed it reproduces the bug on the pre-fix code and passes after the fix. Verified the same shape passes verbatim under `raku`.
- [x] `cargo fmt --all`
- [x] `make test` (release build, `cargo test`, `prove t/`) — all green (`Files=3935, Tests=41851`)
- [x] `make roast` — green except the three container-only failures documented in `docs/agent-environments.md` (`file-tests.t`/`filetest.t` under `uid 0`, `IO-Socket-Async.t`'s sandboxed-network timeout); no other regressions

https://claude.ai/code/session_01UBj19tmmC6B4jK5cvRTLhH

---
_Generated by [Claude Code](https://claude.ai/code/session_01UBj19tmmC6B4jK5cvRTLhH)_

### DIFF
--- a/news/2026-09/need-suppresses-own-nested-use-exports.md
+++ b/news/2026-09/need-suppresses-own-nested-use-exports.md
@@ -1,0 +1,28 @@
+# `need Foo;` no longer suppresses exports `Foo` imports for itself
+
+`need Foo;` loads a compunit without importing its exports into the caller —
+implemented by setting `suppress_exports` for the whole duration of the load
+so `Foo`'s own `is export` declarations are never registered as importable.
+That flag was a single unscoped boolean, though, so it stayed set for every
+nested `use` statement `Foo`'s own body executed too. A `Foo.rakumod` that
+itself said `use Bar;` therefore had `Bar`'s exports silently dropped: `Bar`'s
+`is export` subs never reached `exported_subs`, so `Foo`'s own `use Bar;`
+found nothing to import, and any of `Foo`'s methods calling a `Bar` routine
+died with `Unknown function`.
+
+`Interpreter::use_module_with_tags` now suspends `suppress_exports` for the
+duration of its own nested load, restoring the ambient value afterward. An
+explicit `use` always wants ordinary export semantics, regardless of whether
+it happens to run inside a `need`-loaded compunit's mainline — only the
+directly `need`ed module's own exports should stay unregistered.
+
+This was found via the bundled-library gate running under the vendored `Test`
+module (`MUTSU_REAL_TEST=1`), where four whitelisted `DBIish` files (used
+through `need DBIish::CommonTesting;`) called `diag` from a method of that
+compunit and hit exactly this gap — `DBIish::CommonTesting` itself does
+`use Test;`. All four now pass under the vendored provider with no database
+server present. Pinned by `t/need-preserves-nested-use-imports.t`, a minimal
+`need`-loaded class whose own `use` of a sibling module must resolve from one
+of its methods.
+
+Closes #7805.
--- a/src/runtime/runtime_module.rs
+++ b/src/runtime/runtime_module.rs
@@ -262,7 +262,19 @@ impl Interpreter {
         // resolve with ITS OWN (usually absent) selectors, not the outer ones.
         let (module, dist_selectors) = Self::split_dist_selectors(module);
         let saved = std::mem::replace(&mut self.pending_dist_selectors, dist_selectors);
+        // `suppress_exports` is set for the whole duration of an enclosing
+        // `need` load (see `need_module`), so its own compunit's `is export`
+        // subs never get registered as importable. But an explicit `use`
+        // nested inside that compunit's body (e.g. `need CT;` where
+        // `CT.rakumod` itself says `use Test;`) must still register and
+        // import Test's exports normally -- otherwise CT's own methods can
+        // never resolve `diag` via `module_imported_lexical_names`, even
+        // though CT's own mainline genuinely imported it (#7805). `use`
+        // always wants ordinary export semantics regardless of an ambient
+        // `need`, so suspend the flag for exactly this nested load.
+        let saved_suppress_exports = std::mem::replace(&mut self.suppress_exports, false);
         let result = self.use_module_with_tags_inner(module, tags);
+        self.suppress_exports = saved_suppress_exports;
         self.pending_dist_selectors = saved;
         // `load_module` consumes `pending_use_export_args`; clear any residue
         // here so a native/pragma/already-loaded path (which never reaches
--- a/t/lib-need-nested-import/NestedImportExporter.rakumod
+++ b/t/lib-need-nested-import/NestedImportExporter.rakumod
@@ -1,0 +1,2 @@
+unit module NestedImportExporter;
+sub nested-import-marker() is export { 'nested-import-ok' }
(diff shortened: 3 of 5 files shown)


tokuhirom:
The `test` check failed in its "Bundled-library test suites" step:

```
```

This is a transient network failure fetching the `Terminal::ANSI` battery from `git.sr.ht` — unrelated to this PR's diff (which only touches `need`/`use` module-loading, not the battery infra or `batteries.lock`). The `test` job on `main` succeeded moments earlier (run for #7807's merge, `head_sha 6d301019`), so this is not a standing base-branch failure. Re-running the failed job once to confirm it's a one-off.

---

tokuhirom:
Re-run cleared the Terminal::ANSI network flake, but the `test` job failed again on a different, unrelated cause this time: `t/whenever-callback-recompile-semantics.t` test 5, in the `prove t/` step.

This file was added by `627307a` ("perf: cache the compiled chunk for a whenever callback instead of recompiling per value", merged just before this branch was cut) and has nothing to do with this PR's diff (module `use`/`need` export scoping). Ran it locally against `target/release/mutsu` 5 times: 2 failures, 3 passes — genuinely non-deterministic, consistent with a `whenever`/timing-sensitive flake rather than a regression from this change. Not yet in `flaky-tests.txt`; leaving that ledger update to whoever owns #7801/`627307a` rather than touching it from this unrelated PR. Re-running the failed job once more.

---

tokuhirom:
Update: this same `t/whenever-callback-recompile-semantics.t` subtest 5 failure hit two more PRs independently (#7808, #7814) and is already tracked as a genuine mutsu scheduling bug in #7811 (cross-promise continuation ordering under the pooled worker threads, unrelated to any of our diffs — someone is actively investigating there). Not something to fix from this PR. Will keep retrying/watching `test` here until it lands green, either by luck (it's intermittent, not deterministic) or once #7811's fix merges to `main`.

---

PR #7814, opened 2026-09-09
fix: a module's custom EXPORT and own package name survive an EVAL'd first load

### LINKED ISSUE #7806: Regression (#7764): a module first loaded inside an `EVAL` loses its package name and its `EXPORT::` stash in the importer
`NativeLibs t/01-basic.t` went from **23/23 to 6/23** under the vendored `Test`
module. Bisected to #7764 — my own change, which scoped a transitively-`use`d
module's package name to its importer. CI is unaffected (the native provider is
the default and the file passes 23/23 there), but it undoes what #7714 achieved
and blocks the vendored-`Test` flip #7555 tracks.

Found by running `scripts/battery-testsuite.sh` under `MUTSU_REAL_TEST=1` on
`main` at 5721e1e.

## Repro

```raku
# t.raku, run with -I <mutsu>/modules/NativeLibs/lib
use Test;
use-ok 'NativeLibs:v<0.0.9>';
use NativeLibs:ver<0.0.9>;
for <NativeCall NativeCall::EXPORT NativeCall::EXPORT::ALL NativeLibs NativeLibs::Loader> -> $n {
    say "# $n: " ~ (::($n) ~~ Failure ?? 'MISSING' !! 'ok');
}
```

| name | native | vendored |
| --- | --- | --- |
| `NativeCall` | ok | ok |
| `NativeCall::EXPORT` | ok | **MISSING** |
| `NativeCall::EXPORT::ALL` | ok | **MISSING** |
| `NativeLibs` | ok | **MISSING** |
| `NativeLibs::Loader` | ok | ok |

`01-basic.t`'s tests 7-23 are `NCexports::{$_}:exists` over
`my \NCexports = ::('NativeCall::EXPORT::ALL')`, so all seventeen fail at once.

## Bisect

Debug builds, the repro above under `MUTSU_REAL_TEST=1`:

| commit | `NativeCall::EXPORT::ALL` |
| --- | --- |
| `bbe8f60` (#7714 merge) | ok |
| `d7a6906` (#7743 merge) | ok |
| `a90f9a1` (first parent of the #7764 merge) | ok |
| **`f305977` (#7764 merge)** | **MISSING** |
| `a75c551` (before #7791) | MISSING |
| `5721e1e` (current main) | MISSING |

So neither #7743 nor #7791 is involved; #7764 is.

## Why the vendored provider is the trigger

`use-ok` under the vendored `Test.rakumod` is an `EVAL`, so `NativeLibs` is first
loaded from inside that EVAL's compunit. #7764 removes, at the end of each load,
a `Package`-valued env binding whose target is neither owned by the module being
loaded nor a class/role registered during that load
(`leaked_packages`, `src/runtime/run_modules.rs:929-948`). The next line's real
`use NativeLibs` is then a no-op — the module is already loaded — so nothing
re-installs the bindings into the main scope.

The native `use-ok` is Rust and does not isolate the load, which is why only the
vendored provider shows it. Note `NativeLibs` itself going MISSING: #7764's own
exemption is "a package this module declares itself stays", and it does stay —
in the EVAL's scope, which is gone.

The suspected shape of the fix is that a `use` of an **already-loaded** module
must still install that module's package bindings into the importing scope,
rather than returning early. `registered_here` only diffs `classes` and `roles`,
so an `EXPORT::`/`EXPORT::ALL` stash is not exempt from the removal either;
whether it should be is the second question.

## An independent gap this exposed — do not "fix" the test by restoring the leak

`NativeCall::EXPORT::ALL` is reachable in rakudo after a **plain** `use NativeLibs`
with no `use-ok` anywhere (NativeLibs' custom `sub EXPORT` re-exports `NativeCall`,
and its stash comes with it). mutsu resolves it in **neither** provider without
`use-ok`:

```raku
use NativeLibs:ver<0.0.9>;
say ::('NativeCall::EXPORT::ALL') ~~ Failure ?? 'MISSING' !! 'ok';
# raku: ok        mutsu native: MISSING        mutsu vendored: MISSING
```

So before #7764 this file was green under the native provider for the wrong
reason: the native `use-ok`'s leak supplied what rakudo gets from the custom
export. That is the third instance of the pattern #7714's closing note warned
about. A fix should make the plain-`use` case work, not merely restore the
leaked binding.

## Verification bar

`NativeLibs t/01-basic.t` at 23/23 under **both** providers, plus a `t/` pin for
the plain-`use` case above (which passes verbatim under rakudo). `t/module-transitive-use-does-not-leak-types.t`
(27 assertions) must stay green — it is what #7764 and #7791 pinned, and the
naive fix of dropping the `leaked_packages` removal would break it.

tokuhirom:
## Summary

`#7764` taught the module loader to drop a `Package`-valued env binding left over from a module's own transitive `use` statements once that module's load finished (so a file that never `use`d the transitive dependency directly could not resolve it bare). That fix uncovered two bugs of its own, both surfacing specifically when a module is FIRST loaded from inside an `EVAL` (`Test`'s own `use-ok` is `EVAL ( "use $code" )`, so `use-ok 'M'; use M;` hits both):

1. **A custom `sub EXPORT` ran against the wrong scope.** `sub EXPORT` is part of the module's own closure, so a bareword term inside it (e.g. NativeLibs' `Map.new('NativeCall' => NativeCall, ...)`) must resolve against what the module's own mainline could see. mutsu instead ran `EXPORT` against the (already-restored) importing scope's env, by which point `NativeCall` had already been stripped back out (owned by no one but the module's own transitive `use NativeCall;`). The bareword silently degraded to the plain string `"NativeCall"`, which then shadowed the real package for every importer — reproducible with a **plain `use`, no `EVAL` involved at all** (the "independent gap" the issue called out). Fixed by snapshotting the module's own env right after its body finishes running (before the load's restoration strips it back down), and running `EXPORT` — both on first load and on a later re-`use`'s rerun — against that snapshot instead of the caller's env.

2. **A module's own bare package-name binding wasn't tracked for reinstatement.** `module_package_globals` (and its `reinstate_module_package_globals` counterpart) exists precisely to put a module's bindings back when a scope loses them wholesale, but it only ever recorded `::`-qualified env keys — never the module's OWN bare name (a `unit module Foo;` binds bare `"Foo"` in env). So when a module's first load happened inside a nested call frame whose own env overlay is discarded on return (a sub wrapping an `EVAL`, exactly `use-ok`'s shape), the module's bare name vanished with that frame while `loaded_modules` kept it recorded as loaded. A later real `use` of the same module is then a no-op reuse that never re-installs the missing binding, so the module's own name became permanently unresolvable (the `NativeLibs: MISSING` row in the issue).

Reproduced end-to-end against the real vendored `modules/NativeLibs` module (the issue's own repro) under both the native and vendored `Test` providers — all five symbols (`NativeCall`, `NativeCall::EXPORT`, `NativeCall::EXPORT::ALL`, `NativeLibs`, `NativeLibs::Loader`) now resolve, matching `raku`.

## Test plan

- [x] New regression tests, verified to fail without the fix and pass with it:
  - `t/module-export-sub-sees-module-scope-not-caller.t` (bug 1, no EVAL needed)
  - `t/module-first-loaded-in-eval-keeps-own-package-name.t` (bug 2)
  - Minimal fixtures under `t/lib/Issue7806/`, shaped after NativeLibs' actual `use NativeCall; sub EXPORT(|) {...} unit module NativeLibs:...;` structure.
- [x] `t/module-transitive-use-does-not-leak-types.t` (the #7764/#7791 pin, 27 assertions) still green.
- [x] `t/module-loaded-in-eval-keeps-imports.t`, `t/nativecall-export-stash.t`, `t/nativecall-module-compat.t`, `t/use-ok-real-load.t` and other related existing tests still green.
- [x] `cargo fmt --all` — no changes needed.
- [x] `make lint` — all four configurations (default clippy, `jit`-off clippy, wasm32 clippy, rustdoc) clean.
- [x] `make test` — full suite green (3936 files, 41854 tests).
- [x] `make roast` — only the three documented container-specific failures (`file-tests.t`, `filetest.t` — `uid 0` bypasses permission bits; `IO-Socket-Async.t` — sandboxed network), no new regressions.

Closes #7806.

---

https://claude.ai/code/session_01Hm9kvDefwL8iisjAkrmstZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hm9kvDefwL8iisjAkrmstZ)_

### DIFF
--- a/news/2026-09/module-export-sub-and-own-package-name-across-eval.md
+++ b/news/2026-09/module-export-sub-and-own-package-name-across-eval.md
@@ -1,0 +1,43 @@
+# A module's custom EXPORT and its own package name survive an EVAL'd first load
+
+`#7764` taught the module loader to drop a `Package`-valued env binding left
+over from a module's own transitive `use` statements once that module's load
+finished, so a file that never `use`d the transitive dependency directly
+could not resolve it bare. That fix uncovered two bugs of its own, both
+surfacing specifically when a module was FIRST loaded from inside an `EVAL`
+(`Test`'s own `use-ok` is `EVAL ( "use $code" )`, so `use-ok 'M'; use M;` hits
+both):
+
+1. **A custom `sub EXPORT` ran against the wrong scope.** `sub EXPORT` is
+   part of the module's own closure, so a bareword term inside it (e.g.
+   NativeLibs' `Map.new('NativeCall' => NativeCall, ...)`) must resolve
+   against what the module's own mainline could see. mutsu instead ran
+   `EXPORT` against the (already-restored) importing scope's env, by which
+   point `NativeCall` had already been stripped back out (owned by no one
+   but the module's own transitive `use NativeCall;`). The bareword silently
+   degraded to the plain string `"NativeCall"`, which then shadowed the real
+   package for every importer — reproducible with a plain `use`, no `EVAL`
+   involved at all. Fixed by snapshotting the module's own env right after
+   its body finishes running (before the load's restoration strips it back
+   down) and running `EXPORT` — both on first load and on a later re-`use`'s
+   rerun — against that snapshot instead of the caller's env.
+
+2. **A module's own bare package-name binding wasn't tracked for
+   reinstatement.** `module_package_globals` (and its
+   `reinstate_module_package_globals` counterpart) exists precisely to put a
+   module's bindings back when a scope loses them wholesale, but it only ever
+   recorded `::`-qualified env keys — never the module's OWN bare name (a
+   `unit module Foo;` binds bare `"Foo"` in env). So when a module's first
+   load happened inside a nested call frame whose own env overlay is
+   discarded on return (a sub wrapping an `EVAL`, exactly `use-ok`'s shape),
+   the module's bare name vanished with that frame while `loaded_modules`
+   kept it recorded as loaded. A later real `use` of the same module is then
+   a no-op reuse that never re-installs the missing binding, so the module's
+   own name became permanently unresolvable. Fixed by also keeping the
+   module's own bare name in the tracked set.
+
+Both are pinned: `t/module-export-sub-sees-module-scope-not-caller.t` for the
+first, `t/module-first-loaded-in-eval-keeps-own-package-name.t` for the
+second, using minimal fixtures under `t/lib/Issue7806/` shaped after
+NativeLibs' actual `use NativeCall; sub EXPORT(|) {...} unit module
+NativeLibs:...;` structure. Closes #7806.
--- a/src/runtime/run_modules.rs
+++ b/src/runtime/run_modules.rs
@@ -857,6 +857,13 @@ impl Interpreter {
             // scope already declared.
             let hidden_toplevel = self.hide_toplevel_global_routines();
             let result = self.run_block(&stmts);
+            // Snapshot the env exactly as the module body left it, before any
+            // of the restoration below (the `leaked_packages` removal, the
+            // `saved_plain_env` restore, `unit_lexicals` extraction) strips
+            // module-local bindings back out. `apply_module_export` needs
+            // this: `sub EXPORT` is part of the module's own closure and must
+            // resolve everything the mainline could (see its doc comment).
+            let module_body_env = self.env.clone();
             self.pending_rw_writeback_sources
                 .truncate(saved_pending_rw_writeback_len);
             self.strict_mode = saved_strict_mode;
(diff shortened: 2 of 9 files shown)


tokuhirom:
CI's `test` job failed on `t/whenever-callback-recompile-semantics.t` subtest 5 ("a nested whenever inside the callback fires once per value"). This is unrelated to this PR's diff (module-load/`EXPORT` handling only) — the test exercises `Supply`/`whenever`/`Promise` concurrency timing, a completely different subsystem, and is not touched by any file this PR changes.

Confirmed non-deterministic locally: 5 runs of `prove -e target/release/mutsu t/whenever-callback-recompile-semantics.t` on this branch passed 3/5 and failed 2/5 (same subtest 5 each time), so it's a genuine timing flake rather than a regression from this change. It isn't currently in `flaky-tests.txt`; adding it there would need the full evidence-gathering process in `docs/flaky-test-policy.md`, which is out of scope for this PR, so I'm not touching that file here. Re-running the failed job once.

---

Issue #7824, opened 2026-09-10
A nested `whenever` inside a supply callback emits out of order under CPU contention


tokuhirom:
## Summary

When a `supply` block's `whenever` callback registers a **nested** `whenever` on a `Promise` and
then keeps that promise, the values emitted by the nested bodies can arrive at the tap **out of
source order** under CPU contention. Rakudo keeps them in order under the same load.

A `supply` block serializes its emissions; mutsu appears to serialize the *outer* callback but let
each nested `whenever` body emit on whatever thread keeps its promise, so value 3's nested body can
beat value 2's to the tap.

Found while running the full `t/` suite on a 4-core container: it is the only failure in
`t/whenever-callback-recompile-semantics.t` (test 5), which landed with 627307a
("perf: cache the compiled chunk for a whenever callback instead of recompiling per value", #7820).
The test is not wrong — it asserts the order Rakudo actually produces.

## Minimal repro

```raku
my $sup = Supplier.new;
my @fired;
my $done = Promise.new;
supply {
    whenever $sup -> $v {
        my $p = Promise.new;
        whenever $p { emit $v * 10 }
        $p.keep;
    }
}.tap({ @fired.push($_); $done.keep if @fired.elems == 3 });
$sup.emit($_) for 1..3;
await Promise.anyof($done, Promise.in(5));
say @fired;
```

## Measured, 2026-09-10, release build, 4 CPU burners running alongside

| | result |
| --- | --- |
| `raku` (v2026.07) | `[10 20 30]` — **8/8 runs** |
| `mutsu` | `[10 20 30]` 12/15, **`[10 30 20]` 3/15** |

Idle, mutsu is stable (`t/whenever-callback-recompile-semantics.t` passes 5/5 in isolation), which
is why CI has not caught it yet. Under load:

```
$ MUTSU_BIN=target/release/mutsu scripts/flake-repro.sh -n 20 -l 4 \
    t/whenever-callback-recompile-semantics.t
t/whenever-callback-recompile-semantics.t   pass=14  fail=6   FLAKY (6/20 failed)
```

with the failure always this shape:

```
not ok 5 - a nested whenever inside the callback fires once per value
# expected: '10 20 30'
#      got: '10 30 20'
```

## Why this is a bug and not a quarantine candidate

Per `docs/flaky-test-policy.md`, a wrong answer is never quarantined. The number of emitted values
[...]

Issue #7819, opened 2026-09-09
t/ contains too much files


tokuhirom:
we need to refactor this directory

tokuhirom:
Claiming: claude/directory-nesting-design-kp9mpo

Splitting this into two PRs so the 3,938-file rename does not conflict with every in-flight PR that adds a `t/` file:

1. **This branch — design + infrastructure, no file moves.** `docs/t-directory-layout.md` (sixteen subject categories, two-level depth cap, globally-unique basenames), `prove -r` in the `make test` recipe and all three CI TAP steps, a new `scripts/check-t-layout.sh` guard, and the rule updates in `CLAUDE.md` / `AGENTS.md` / the skills.
2. **A follow-up branch — the mechanical `git mv` sweep**, driven by a checked-in classifier script plus an override list, plus the `flaky-tests.txt` path rewrite.

---
_Generated by [Claude Code](https://claude.ai/code)_

tokuhirom:
Step 1 of 2 is up: #7826 — the design (`docs/t-directory-layout.md`), the `prove -r` + `scripts/check-t-layout.sh` infrastructure, and the rule updates in `CLAUDE.md` / `AGENTS.md` / the skills. No test files move in it.

Step 2, the mechanical `git mv` sweep, follows once that lands.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #7811, opened 2026-09-09
`t/whenever-callback-recompile-semantics.t` subtest 5 (nested `whenever`) is flaky: emit order races


tokuhirom:
Found while investigating a red `test` CI check on PR #7808 (unrelated change — module-load/qualified-name visibility, no concurrency code touched).

## Repro

```
prove -e 'target/release/mutsu' t/whenever-callback-recompile-semantics.t
```

Subtest 5 ("a nested whenever inside the callback fires once per value") fails intermittently:

```
not ok 5 - a nested whenever inside the callback fires once per value
# expected: '10 20 30'
#      got: '10 30 20'
```

The relevant code:

```raku
my $sup3 = Supplier.new;
my @fired;
my $done = Promise.new;
my $sup3out = supply {
    whenever $sup3 -> $v {
        my $p = Promise.new;
        whenever $p { emit $v * 10 }
        $p.keep;
    }
};
$sup3out.tap({ @fired.push($_); $done.keep if @fired.elems == 3 });
$sup3.emit($_) for 1..3;
await Promise.anyof($done, Promise.in(5));
is @fired, [10, 20, 30], ...;
```

Each `$sup3.emit($_)` schedules a nested `whenever $p { emit $v * 10 }` whose actual `emit` runs asynchronously once `$p.keep` resolves it — nothing in the test (or, per rakudo semantics, the language) orders those three async continuations against each other, so the three `emit`s can interleave in any order. This is a genuine race in the test's own assumption, not (as far as investigated) a mutsu scheduling bug — verified on both `main` (7470841) and the #7808 branch, both showing a nonzero failure rate across 20 local repeats (main: 5/20, #7808 branch: 9/20 — different enough to look like noise on a small sample, not a regression: the earlier full debug-build `t/` suite run on the #7808 branch, under different system load, passed this file cleanly).

## Suggested fix

Either:
- Make the assertion order-insensitive (`is @fired.sort, [10, 20, 30].sort, ...`) if per-value ordering isn't actually guaranteed here, or
- If rakudo's `whenever`/`Promise.keep` DOES guarantee ordering here (needs checking against `raku`), find and fix the actual mutsu scheduling bug that lets these races interleave.

Whoever picks this up should start by checking `raku`'s own behavior on this exact snippet (repeated several times) to determine which of the two applies.

tokuhirom:
[...]

PR #7826, opened 2026-09-10
docs: define a nested layout for t/ and land the infrastructure for it

tokuhirom:
`t/` had grown to **3,948 flat `.t` files**, which is past the point where a directory listing is a usable index: `ls t/` is unreadable, and neither a human nor an agent can answer "is there already a test for this?" without a full-text grep over the tree.

This is the **design half** of that refactor. **It moves no existing test files.** The ~3,948 `git mv`s land in a separate, mechanical PR, because a rename that size is unreviewable if it also carries policy changes — and because the policy has to be on `main` first, or every in-flight PR that adds a `t/` file conflicts with it.

Refs #7819.

## The layout

[`docs/t-directory-layout.md`](https://github.com/tokuhirom/mutsu/blob/claude/directory-nesting-design-kp9mpo/docs/t-directory-layout.md) is the new authority for placement; `CLAUDE.md` and `AGENTS.md` point at it rather than restating the list.

- **Sixteen top-level subject categories as a closed set**: `lang`, `types`, `collections`, `control`, `routines`, `oo`, `regex`, `grammar`, `exceptions`, `io`, `modules`, `concurrency`, `nativecall`, `rakuast`, `vm`, `tooling`.
- No `.t` at `t/` top level; at most two levels below `t/`.
- **Basenames stay globally unique, and a migrated file keeps its basename exactly** — `t/regex-backtrack.t` → `t/regex/regex-backtrack.t`, not `t/regex/backtrack.t`. The redundancy is the point: several hundred `t/<name>.t` references in `docs/`, `news/` and code comments are prose rather than links, so nothing rewrites them mechanically. An unchanged, unique basename keeps every one resolvable by one `git grep`, keeps the migration checkable by name, and keeps the tools that index tests by basename rather than path (`scripts/test-module-sweep.sh`'s flat work directory above all) working untouched.
- Placement goes by **what the test would catch if it broke, not the syntax it happens to use**: a test that writes `for @a { ... }` to prove a closure captures the loop variable by reference is a `vm/` writeback test, not a `control/` loop test.
- `lib/`, `fixtures/` and `packages/` hold no tests; the nine grandfathered `lib-*` / `*-lib` directories must not grow.

Tests reach fixtures by a path relative to the repository root (`-I t/lib`) and `prove` runs from the root, so nesting a test does **not** change how it loads anything — no test body needs editing for the migration.

## Infrastructure

Without this, the migration PR would silently break things.

- **`prove` does not descend into subdirectories without `-r`**, so the `make test` recipe and all three CI TAP steps (`test`, `gc-stress`, `jit-stress`) now pass it. Inert on the current flat tree: `prove --dry` and `prove -r --dry` returned byte-identical file lists.
- **`scripts/check-t-layout.sh`** (new; `make check-t-layout`, a `make test` prerequisite and a CI step) enforces the machine-checkable half: known categories only, the two-level depth cap, declared subcategories, no `.t` under a support directory, and globally-unique basenames. Each rule was verified by deliberately violating it. The "no `.t` at `t/` top level" rule sits behind a `MIGRATED` flag so the guard lands and starts protecting the other five **before** the move rather than after it.
- **`scripts/test-module-sweep.sh`** discovered tests with `ls t/*.t`; it now uses `find`. It is the one consumer that would otherwise have quietly swept nothing once files moved.

## One test, placed in a category — the layout proves itself

`t/concurrency/whenever-nested-promise-emit-order.t` is an independent regression pin for the supply-block reaction ordering fixed on `main` by 9d59f4c / a43f1d1 (#7811). It complements `t/supply-serialize-fifo.t` by driving the guarantee through more values, and by checking that a `whenever` on a *bare* promise — which has no enclosing order and stays on the plain pooled path — still fires.

It sits in a category directory rather than at top level, so all three CI TAP jobs now prove `prove -r` really picks up a nested test:

- `prove -r --dry` lists it; plain `prove --dry` does not.

That turns the central assumption of this PR from a claim into a verified fact.

## A guard bug the merge surfaced

`make roast` leaves an empty `t/spec/S22-package-format` behind. Git does not track it (it contains no files), so `git status` is clean, but `check-t-layout.sh` flagged it as an unknown category — i.e. a second `make test` after a `make roast` would fail on a scratch artifact. The guard now judges only directories that actually contain a `.t`; a real stray test is still caught, both there and by the per-file check.

## Verification

- `cargo fmt --all --check` clean.
- `make check-t-layout` — OK.
- `prove --dry` vs `prove -r --dry` — byte-identical, re-confirmed after merging current `main`.
- `make lint`, `make test` and `make roast` — re-running on the merged state; I'll report all three.
- `scripts/flake-repro.sh -n 20 -l 4` on `t/concurrency/whenever-nested-promise-emit-order.t` and `t/whenever-callback-recompile-semantics.t`: both `pass=20 fail=0 DETERMINISTIC PASS` against current `main`.

### Note on an earlier revision of this branch

An earlier push carried a concurrency fix of my own for the same ordering bug (filed as #7824). It was **wrong**: running the nested body inline on the resolving thread fixed the order but deadlocked `Cro::HTTP`'s middleware pipeline, whose resolver must make progress for the body's own await to complete — caught by the battery gate, which my local `make test`/`make roast` do not cover. `main` meanwhile landed a better fix for the same diagnosis (#7811): a `SupplyTicket` reserved on the resolving thread, with the sequencer beside the group lock rather than inside it, which also avoids a Log::Async throughput regression. That commit is dropped from this branch; `main`'s fix is what this PR now builds on, and #7824 is closed as a duplicate of #7811.

Note also that `make test` **exits 0 even on `Result: FAIL`** — its recipe is a `( … ) | tee` pipeline with no `pipefail`, so `make` reports `tee`'s status. Not changed here, but worth knowing when reading a green `make test`.

https://claude.ai/code/session_01RmggmzmJQfQkywq1QU3Cb9

### DIFF
--- a/.agents/skills/rakuast-implementation/SKILL.md
+++ b/.agents/skills/rakuast-implementation/SKILL.md
@@ -7,7 +7,7 @@ metadata:
 
 # RakuAST implementation
 
-Use this skill for work involving `src/rakuast/`, `t/rakuast*.t`,
+Use this skill for work involving `src/rakuast/`, `t/rakuast/`,
 the RakuAST campaign-overview issue, or RakuAST-specific ADR/design work.
 It covers the RakuAST-specific investigation and implementation details; use
 the repository's normal ticket and PR workflow for branching, validation, and
@@ -53,7 +53,7 @@ converter.
 
 ## Test the contract
 
-Add a focused `t/rakuast-<slice>.t` test with three kinds of assertions:
+Add a focused `t/rakuast/rakuast-<slice>.t` test with three kinds of assertions:
 
 - read direction: measured node classes, accessors, and relevant field
   omission/presence;
--- a/.agents/skills/test-util-workout/SKILL.md
+++ b/.agents/skills/test-util-workout/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: test-util-workout
-description: Pick one Test::Util function from roast's Test-Helpers package, write a t/ test for it, and fix the interpreter until it passes, ending in a merged PR. Use when asked for a "Test::Util workout" or to work through the Test::Util helper functions.
+description: Pick one Test::Util function from roast's Test-Helpers package, write a t/tooling/ test for it, and fix the interpreter until it passes, ending in a merged PR. Use when asked for a "Test::Util workout" or to work through the Test::Util helper functions.
 metadata:
   short-description: Land one Test::Util helper end-to-end
 ---
@@ -16,13 +16,14 @@ learn the expected behaviour.
 ## Workflow
 
 1. Read `roast/packages/Test-Helpers/lib/Test/Util.rakumod` for the list of exported functions.
-2. Check `t/` for the existing coverage (`test-util-*.t`, `is-run.t`, ...) to see which
-   functions already have tests.
+2. Check `t/tooling/` for the existing coverage (`test-util-*.t`, `is-run.t`, ...) to see
+   which functions already have tests. The `Test` module and roast's `Test::Util` helpers
+   live in the `tooling/` category (`docs/t-directory-layout.md`).
 3. Pick **one** unimplemented or undertested function. Once chosen, do **not** switch to a
    different one mid-task.
-4. Write `t/<function-name>.t` exercising it with several cases: basic usage, edge cases, and
-   combined checks.
-5. Run it: `timeout 30 target/debug/mutsu t/<function-name>.t`.
+4. Write `t/tooling/<function-name>.t` exercising it with several cases: basic usage, edge
+   cases, and combined checks.
+5. Run it: `timeout 30 target/debug/mutsu t/tooling/<function-name>.t`.
 6. Fix the interpreter until it passes. When the spec is unclear, check with `raku -e '<code>'`
    (the `install-raku` skill gets `raku` if the container has none) and consult `raku-doc/`.
 7. Run `make test` and `make roast` to check for regressions, then read the results out of
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -197,7 +197,7 @@ jobs:
         # widen the budget rather than quarantining a real (if heavy) test.
         # Kept at 45 on the release binary too: the budget exists for the
         # heaviest file under contention, and headroom there is free.
-        run: prove -j4 --merge -e 'scripts/run-t-test.sh' t/
+        run: prove -r -j4 --merge -e 'scripts/run-t-test.sh' t/
         env:
           MUTSU_BIN: target/release/mutsu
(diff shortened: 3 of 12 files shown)


tokuhirom:
## CI: `test` job failed on a pre-existing bug, not on this diff

**Failing check:** `test` — `t/whenever-callback-recompile-semantics.t`, `Failed 1/6 subtests` (test 5), `Result: FAIL`. Every other completed check is green (`lint-configs`, `wasm-e2e`, `miri`, `changes`, `label`); `gc-stress` and `jit-stress` are still running.

**Why it is not this PR's.** This diff contains **zero `.rs` files** — it is documentation plus `prove -r` and two shell scripts. The one behavioural change, adding `-r`, is provably inert on the current flat tree: `prove --dry` and `prove -r --dry` return byte-identical 3,944-file lists, so the same files run in the same order as before. There is no path from this diff to a `whenever`/`Promise` scheduling outcome.

It is a genuine, pre-existing concurrency bug, filed before this failure as #7824 and now root-caused there:

- `whenever <Promise>` nested inside another `whenever` body registers via `shared.on_resolve` (`src/runtime/subtest.rs:650`).
- `$p.keep` hands that promise's waiters to `dispatch_waiters` (`src/value/value_async.rs:193`), which calls `worker_pool::submit` **once per resolved promise**.
- `worker_pool` is multi-worker, so three `$p.keep` calls become three independent pooled tasks. Ordering is preserved *within* one promise's waiter list and lost *across* promises — exactly the observed `[10 30 20]` instead of `[10 20 30]`.

Measured (release build, 4 CPU burners): `raku` v2026.07 `[10 20 30]` 8/8; `mutsu` `[10 20 30]` 12/15, `[10 30 20]` 3/15. Idle, the file passes 5/5. `scripts/flake-repro.sh -n 20 -l 4` on it: `pass=14 fail=6`. The test landed on `main` in 627307a (#7820) and asserts the order Rakudo actually produces, so the test is right and mutsu is wrong.

**No fix exists yet to port.** The two-line fix — take the `supplier_serialize_group` guard the outer emit path already holds (`native_supplier_methods.rs:629`) in the nested-Promise arm too — would narrow the window but not close it: a mutex gives mutual exclusion, not FIFO order, and three pooled tasks can still take it in any order. Rakudo's semantics need the supply block to dispatch `whenever` bodies through a **sequence-ordered queue**, which is a design change to supply-block dispatch and belongs in its own PR with its own full-suite run — not bolted onto a layout PR. The proposed scope is written up on #7824.

Standing down on it here, and spending the one permitted re-run once `gc-stress`/`jit-stress` finish, so the re-run covers everything that actually failed. Not quarantining it: a wrong answer is never a quarantine candidate (`docs/flaky-test-policy.md`).

---

Issue #7554, opened 2026-09-08
Make the vendored upstream `Test` module the default provider


tokuhirom:
The vendored upstream `Test.rakumod` (`modules/Rakudo-Core/lib/`, source hash
`f34dec45d52ad099c37f42fdbd93e277`, 953 lines of ordinary Raku) **runs verbatim
under mutsu**. What is still open is making it what a bare `use Test` resolves
to. Today that is opt-in behind `MUTSU_REAL_TEST=1`; the native TAP provider in
`runtime/test_functions.rs` is still the default.

The flip was attempted on 2026-09-07/08 and **withdrawn** — see "Why it was
withdrawn" below. This file is the entry point for resuming it.

## What is already done

The campaign's real product was never the flip: it was the ~130 general
interpreter fixes the module forced along the way. The module exercises phasers,
`EVAL`, `callframe`, custom traits, `nqp::` ops, subtests and lazy evaluation, so
every gap it hit was a gap in mutsu, not in `Test`. Those fixes are all landed
and are independent of which provider is default. This is BATTERIES.md rung 2 in
its intended form — grow the interpreter until the real module runs verbatim,
rather than reimplement it natively.

## The completion criteria, and where they stand

Measured on a 4-core box, 2026-09-07/08, release build unless stated.

**1. No correctness regression in the roast and `t/` suites — MET.** The roast
dual-provider sweep (`scripts/roast-test-module-sweep.sh`, whitelist, both
providers, 4 jobs) ended at 1431 pass under both, 0 regressed, 0 pass-only-under-
real, 4 fail under both (container-environment files). Under the vendored module
as default: `prove -j4 t/` 3836 files / 40528 tests PASS, and `make roast` 1436
files / 218939 tests with only the known container-environment failures.

**2. The per-file timeout class is closed — MET.** `roast/S03-buf/write-int.t`
(~93 000 assertions) against the 30 s budget: 11.4 s idle, 14.6 s under
`prove -j4` alongside the five next-heaviest files — a 51% margin, measured
directly on 4 cores rather than scaled from a 12-core box.

**3. `Test::Util` still composes — MET.** roast loads it from
`roast/packages/Test-Helpers/`, exercised by the sweep under both providers.

**4. The `Bundled-library test suites` gate — NOT MET.** This is the blocker.

## Why it was withdrawn

The gate (`scripts/battery-testsuite.sh`, run inside CI's `test` job) reports
[...]

PR #7833, opened 2026-09-10
refactor: move every t/ test into its category directory

### LINKED ISSUE #7819: t/ contains too much files

tokuhirom:
The second half of #7819, following #7826 which landed the layout, the `prove -r` discovery and the `check-t-layout` guard. This is the move.

`t/` had 3,948 flat `.t` files — past the point where a directory listing is an index. The largest directory in the tree is now **193** files.

Closes #7819.

## The shape

Ten of the sixteen categories carry a second level. One level was not enough: it left `collections` at 534, `oo` at 533 and `routines` at 492, which is no more browsable than the flat directory was. For calibration, **roast's own largest directory is 70 files**. After the split the largest subcategory is `t/routines/signature/` at 167.

```
t/collections/{array,hash,lazy-seq,range-pair,set-bag-mix,subscript,transform}/
t/concurrency/{promise,supply,thread-lock}/
t/lang/{adverbs,operators,parsing,quoting}/
t/modules/{batteries,compunit,import-export}/
t/oo/{attribute,class,construct,method,mop,role,trait}/
t/regex/{match,subst,syntax}/
t/routines/{call,closure,dispatch,signature}/
t/types/{coercion,enum-subset,numeric,string,temporal}/
t/vm/{binding,codegen,frames,scope,writeback}/
t/{control,exceptions,grammar,io,nativecall,rakuast,tooling}/
```

**The subcategory boundaries are the most arguable part of this PR.** They all live in one `SUBRULES` table in `scripts/migrate-t-layout.py`; re-running the script regenerates the tree, so they are cheap to re-cut if you would draw them differently.

## How to review a 4,000-file rename

Don't read the diff — re-run the rules and diff their output:

```
python3 scripts/migrate-t-layout.py --check    # every placed file is where the rules say
make check-t-layout                            # categories, depth cap, unique basenames
```

`scripts/migrate-t-layout.py` performed the sweep and stays in the tree, which is what makes the result reproducible rather than a one-off sort. It places a file from an explicit `OVERRIDES` map, then an ordered `RULES` list (first regex wins, most specific subject first), then the per-category `SUBRULES` table. If any file matches nothing it reports and refuses to apply, so the tree cannot end up half-placed. **Six files needed a hand-written override; the rules placed the other 3,942.**

`--check` is wired into `make check-t-layout`, so the layout stays reproducible rather than drifting.

## Commit 1 is a pure rename; commits 2 and 3 are the cost

The move itself is **3,949 renames with zero insertions and zero deletions**. But it was not a pure rename overall, and the reason is the most useful thing this turned up.

### 129 tests addressed fixtures relative to their own file

```raku
use lib $?FILE.IO.parent.add('lib').Str;                            # -> t/lib, before
use lib $*PROGRAM.parent(2).add("roast/packages/Test-Helpers/lib");
```

Both only ever resolved correctly because every test sat directly in `t/`. One level down the first points at `t/<category>/lib` and the second at `t/`, so those tests could not find `Test::Util` and **aborted before running a single assertion** — 12 files reporting `exited 255, Tests: 0`, and 721 fewer tests in the suite.

They now use root-relative literals (`use lib 't/lib'`), which is depth-independent because `prove`, `make test` and CI all run from the repository root — and is what the 100+ tests already saying `use lib 't/lib'` had been doing all along. Same treatment for the non-`use lib` cases: EVALFILE targets, a vendored module path, and an `-I` list built from `$*PROGRAM.parent(2)`.

Left alone deliberately: `t/fixtures/use-lib-dynamic-sibling/probe.rakutest`, whose whole purpose is to exercise `use lib $*PROGRAM.sibling(...)`, and which does not move.

**`docs/t-directory-layout.md` now states the rule** — address a fixture from the repository root, never from the test file — replacing its earlier claim that no test body would need editing. That claim was true of the 100+ tests I sampled and false for these 129; it is the one thing this migration cost something to learn.

### Four files named a path rather than loading one

Three asserted on their **own** filename (`$?FILE`, `callframe.file`) and now match on the basename, which stays unique and stable even if a category is re-cut. One Rust integration test, `tests/proto_method_body_compiled_once.rs`, executes a specific `t/` file by path. That one aborts `make test` before `prove` runs, which is why the first full-suite attempt never reached the TAP files at all.

## Verification

| gate | result |
| --- | --- |
| `make test` | **PASS** — 3,949 files, **41,978 tests** |
| `make roast` | only the three documented container failures (`docs/agent-environments.md`), unchanged across four runs |
| `make lint` | exit 0, all four configurations |
| `cargo fmt --all --check` | clean |
| `make check-t-layout` | OK (guard + `--check`) |

The test count is the number that matters: **41,978 both before and after**, versus 41,257 while the fixture paths were broken. Every repaired file is running its full plan again rather than silently short-planning.

Filenames are unchanged throughout — `t/regex-backtrack.t` became `t/regex/syntax/regex-backtrack.t`, not `.../backtrack.t` — so the several hundred prose `t/<name>.t` references in `docs/` and `news/` still resolve by `git grep`, and were deliberately not rewritten.

https://claude.ai/code/session_01RmggmzmJQfQkywq1QU3Cb9

---
_Generated by [Claude Code](https://claude.ai/code/session_01RmggmzmJQfQkywq1QU3Cb9)_


tokuhirom:
## CI `test` failure was a latent CI bug, now fixed in `2da6f85a6`

Six files (`t/io/io-handle-*`, `t/io/io-cathandle-read-dispatch.t`, `t/io/dir-seq-map-gist.t`, two `t/collections/lazy-seq/*`) died at their first `spurt` with `exited 255` / "planned N but ran 0".

**Not the move, and not the tests.** Two dozen tests write scratch files to a repository-relative `tmp/`, which is gitignored and therefore absent from a fresh checkout. `make test` starts with `@mkdir -p tmp`, and so do the `gc-stress` and `jit-stress` TAP steps — **the `test` job's TAP step was the only one of the four that did not.** It got away with it because the flat `t/` ordering happened to run something that created the directory before anything needed it; nesting reshuffled `prove`'s walk and removed that luck.

The evidence is clean:

- **`gc-stress` passed on this exact commit**, running the identical 3,950-file suite. The only relevant difference between the two steps is that `mkdir -p tmp`.
- `dir-seq-map-gist.t` failed *only* tests 6-7 — precisely where it stops exercising `dir` and starts writing to `tmp/`.
- Removing `tmp/` locally reproduced the CI symptom exactly, and restoring it made the same `prove -r -j4 -e scripts/run-t-test.sh` command pass:

```
=== without mkdir (CI before fix):   Files=2, Tests=0   Result: FAIL
=== with mkdir -p tmp (after fix):   Files=2, Tests=7   Result: PASS
```

The fix adds `mkdir -p tmp` to that step so it no longer depends on execution order. No test file is changed.

---

tokuhirom:
## Update: the TAP suite is green; the remaining blocker is a `main` regression

The `mkdir -p tmp` fix worked — **`prove -r -j4` over all 3,950 files reports `Result: PASS`**, and so does the roast step in the same job. `gc-stress`, `lint-configs`, `wasm-e2e`, `miri` and `changes` are all green.

What now fails is the **battery gate**, on three libraries this PR does not touch:

```
REGRESSION: whitelisted DateTime::Parse/01-basic.t no longer passes
REGRESSION: whitelisted Encode/01-basic.t no longer passes
REGRESSION: whitelisted Log::Async/08-use.rakutest no longer passes
```

**It is not this PR's.** `git diff origin/main -- src/ modules/ batteries.lock batteries-whitelist.txt batteries-exclude.txt` is **empty** — every input the gate exercises is byte-identical to `main`. I reproduced the identical three regressions locally on that identical code.

Root-caused and filed as #7837: `EVAL $code, context => $caller-context` ignores the context, so the vendored `Test.rakumod`'s string form of `throws-like` (`Test.rakumod:640`) evaluates in the Test module's scope instead of the caller's and cannot see the test file's imports:

```
    not ok 2 - right exception type (Encode::X::Encode::Unknown)
    # Got:      X::AdHoc
    # Exception message: Could not find symbol 'Encode::decode'
```

Minimal repro with a side-by-side oracle divergence is in the issue. The same file passes under the native provider (`MUTSU_REAL_TEST=`), which is why this only appeared when da21ca091 (#7554) made the vendored module the default — a couple of hours ago, after this branch's work began.

**No fix exists to port**, and fixing `EVAL`'s context handling is an interpreter change well outside a directory move, so I am not widening this PR with it. Not re-running either: it is deterministic, not a flake, and a re-run would only burn a cycle.

This PR is blocked on #7837 landing. Everything else on it is green.

---

PR #7834, opened 2026-09-10
fix(supply): reserve a SupplyTicket for an already-resolved promise too

### LINKED ISSUE #7831: t/supply-serialize-fifo.t test 3 fails ~17% of runs on main: an already-kept promise skips the SupplyTicket sequencer
## Summary

`t/supply-serialize-fifo.t` — the pin added by 9d59f4c for #7811 — **fails about 1 run in 6 on `main`, on an idle machine**. Since the `t/` TAP suite is fatal in CI, this intermittently reddens every PR.

The #7811 fix orders supply-block reactions by reserving a `SupplyTicket` on the resolving thread. But it only does so on the path where the promise is still `Planned`. A promise that is **already resolved** when the nested `whenever` registers bypasses the sequencer completely — which is exactly the case the file's own test 3 was written to cover.

## Root cause

`SharedPromise::on_resolve_in_supply_group` (`src/value/value_async.rs:291`):

```rust
if state.status == "Planned" {
    state.waiters.push((waiter, group));   // group kept -> dispatch_waiters reserves a SupplyTicket
    false
} else {
    // ... already Kept/Broken ...
    drop(state);
    waiter(status, result, output, stderr);  // <-- runs NOW, on this thread, with no ticket
    true
}
```

The `group` argument is simply dropped in the `else` branch. Only `dispatch_waiters` reserves tickets, and it is never reached for an already-resolved promise. So such a reaction never enters the sequencer and races against the ticketed ones.

## Repro

`t/supply-serialize-fifo.t` test 3 is already the minimal case — it alternates the two paths deliberately:

```raku
whenever $src -> $v {
    my $p = Promise.new;
    $p.keep if $v %% 2;        # even: resolved BEFORE the whenever registers -> no ticket
    whenever $p { emit $v }
    $p.keep unless $v %% 2;    # odd: resolved AFTER  -> ticketed
}
```

## Measured on `main` (release build, idle machine, 2026-09-10)

```
30 runs: pass=25 fail=5   (~17%)
```

Always test 3, always the same shape — every even value ahead of every odd one, i.e. the two paths cleanly separated rather than interleaved:

```
not ok 3 - already-kept and later-kept promises interleave in value order
# expected: '1 2 3 4 5 6'
#      got: '2 4 6 1 3 5'
```

Tests 1 and 2 (all promises kept after registration, so all ticketed) pass consistently — consistent with only the un-ticketed path being broken.

Reproduced on a branch whose `src/` is byte-identical to `main` (`git diff origin/main -- src/` empty), so this is not a merge artifact.

## Suggested fix

Reserve a ticket in the already-resolved branch too, so both paths enter the group through the same sequencer, rather than one running inline ahead of the queue. Worth checking whether running it inline on the registering thread is safe at all here: doing that for the *pending* path is what deadlocked `Cro::HTTP`'s middleware pipeline when I tried it (its resolver must make progress for the body's own await to finish), so the inline branch may want to hand off to a pooled worker like `dispatch_waiters` does rather than execute user code on the caller's thread.

Not quarantining it in `flaky-tests.txt`: this is a wrong answer, not a non-deterministic-by-design assertion, and the policy is explicit that wrong answers are never quarantined.

cc the #7811 work — this is the remaining half of that fix, not a regression from it.

tokuhirom:
## Problem

`t/supply-serialize-fifo.t` test 3 — the pin added for #7811 — fails about one run in six on `main`. Measured here on an idle release build: **11 of 30 runs failed**. Since the `t/` TAP suite is fatal in CI, this reddens unrelated PRs at random.

It is a wrong answer, not a non-deterministic-by-design assertion, and not a regression from #7811 — it is the half of that fix that was never written.

```
not ok 3 - already-kept and later-kept promises interleave in value order
# expected: '1 2 3 4 5 6'
#      got: '2 4 6 1 3 5'
```

## Root cause

#7811 orders a supply block's reactions by having the thread that *resolves* a promise reserve the reaction's place in the block's sequencer (a `SupplyTicket`) before the pooled worker that will run it has been woken.

But `SharedPromise::on_resolve_in_supply_group` only did that on the path where the promise is still `Planned`. The ticket is taken in `dispatch_waiters`, and a promise that is **already** resolved when the nested `whenever` registers never reaches `dispatch_waiters` — its `group` argument was simply dropped, and the reaction ran inline on the registering thread, ahead of every reaction already ticketed and waiting for a worker.

So the two paths were ordered against each other by luck. That is exactly what the test alternates between (even values keep the promise *before* the nested `whenever` sees it, odd values *after*), which is why the wrong answer came out cleanly separated rather than interleaved.

## Fix

The already-resolved branch now reserves its ticket too, on the registering thread — the ordering point, since that thread is running the enclosing supply block's reactions in order — and hands the body to a pooled worker exactly as `dispatch_waiters` does. Both paths enter the block through the one sequencer, in the order the reactions were created.

Redeeming the ticket *inline*, keeping the body on the caller's thread, is not an option: that thread is typically running the enclosing supply block's own reaction and so already holds the group lock, so parking it behind an earlier ticket — whose worker is itself blocked on that lock — would deadlock. This mirrors the reason the *pending* path hands off to a worker rather than running user code on the resolver, which is what deadlocked `Cro::HTTP`'s middleware pipeline when it was tried there.

Contract change: `on_resolve_in_supply_group` now returns `false` for an already-resolved promise when a group is given, because the reaction genuinely has not run yet. Plain `on_resolve` (no group) keeps its synchronous fast path and its `true`. No caller reads the return value.

## Verification

- Before: 11 / 30 runs failed (release, idle machine). After: **0 / 40**.
- `t/supply-serialize-fifo.t` grows a fourth test that keeps *every* promise before its nested `whenever` registers, so the already-resolved path is exercised on its own rather than only in alternation with the ticketed one. `raku` agrees with all four tests.
- `cargo fmt --all`, `make lint` (all four configurations), `make test` (3949 files, 41979 tests, PASS).
- `make roast`: the only failures are the three documented container-only ones — `roast/6.c/S32-io/file-tests.t` and `roast/S16-filehandles/filetest.t` (this container runs as `uid 0`, which bypasses the permission bits they assert on) and `roast/S32-io/IO-Socket-Async.t` (sandboxed network), matching the shapes and test numbers listed in `docs/agent-environments.md` exactly.

Closes #7831.

---

### DIFF
--- a/news/2026-09/already-resolved-promise-supply-ticket.md
+++ b/news/2026-09/already-resolved-promise-supply-ticket.md
@@ -1,0 +1,53 @@
+# An already-resolved promise now takes a SupplyTicket too
+
+`t/supply-serialize-fifo.t` test 3 — the pin added for #7811 — failed about one
+run in six on `main` (11 of 30 on an idle release build here), which made the
+fatal `t/` suite redden PRs at random. The failure was always the same shape:
+
+```
+# expected: '1 2 3 4 5 6'
+#      got: '2 4 6 1 3 5'
+```
+
+Not noise, and not a regression from #7811: it is the half of that fix that was
+never written.
+
+#7811 orders the reactions of a supply block by having the thread that resolves
+a promise reserve the reaction's place in the block's sequencer (a
+`SupplyTicket`) before the pooled worker that will run it has even been woken.
+But `SharedPromise::on_resolve_in_supply_group` only did that on the path where
+the promise was still `Planned`: the ticket is taken in `dispatch_waiters`, and
+a promise that is **already** resolved when the nested `whenever` registers
+never reaches `dispatch_waiters`. Its `group` argument was simply dropped, and
+the reaction ran inline on the registering thread — ahead of every reaction
+already ticketed and queued for a worker. The two paths were therefore ordered
+against each other by luck, which is exactly what the test alternates between
+(even values keep the promise before the nested `whenever` sees it, odd values
+after), and why the wrong answer came out cleanly separated rather than
+interleaved.
+
+The already-resolved branch now reserves its ticket too, on the registering
+thread — the ordering point, since that thread is running the enclosing supply
+block's reactions in order — and hands the body to a pooled worker exactly as
+`dispatch_waiters` does. Both paths enter the block through the one sequencer,
+in the order the reactions were created.
+
+Redeeming the ticket inline instead, and keeping the body on the caller's
+thread, is not an option: that thread is typically running the enclosing supply
+block's own reaction and so already holds the group lock, so parking it behind
+an earlier ticket — whose worker is itself blocked on that lock — deadlocks.
+This mirrors the reason the *pending* path hands off to a worker rather than
+running user code on the resolver, which is what deadlocked `Cro::HTTP`'s
+middleware pipeline when it was tried there.
+
+Consequence for the API contract: `on_resolve_in_supply_group` now returns
+`false` for an already-resolved promise when a group is given, because the
+reaction genuinely has not run yet. Plain `on_resolve` (no group) keeps its
+synchronous fast path and its `true`.
+
+40 consecutive runs of the file are green after the fix. `t/supply-serialize-fifo.t`
+grows a fourth test that keeps *every* promise before its nested `whenever`
+registers, so the already-resolved path is exercised on its own rather than only
+in alternation with the ticketed one.
+
+Closes #7831.
--- a/src/runtime/react_whenever.rs
+++ b/src/runtime/react_whenever.rs
@@ -530,7 +530,10 @@ impl Interpreter {
             // what keeps N nested `whenever <Promise>` bodies -- one created
             // per value by an outer `whenever`, each on its own promise -- in
             // the order their promises were kept, instead of in whichever
-            // order N pooled workers happened to wake up in (#7811).
+            // order N pooled workers happened to wake up in (#7811). A promise
+            // already kept by the time this runs reserves its place right
+            // here, on this thread, so it queues behind the bodies created
+            // before it instead of running inline ahead of them (#7831).
             shared.on_resolve_in_supply_group(
                 Box::new(move |status, result, _output, _stderr| {
                     if status == "Kept" {
--- a/src/runtime/supply_promise.rs
+++ b/src/runtime/supply_promise.rs
@@ -283,9 +283,11 @@ impl Interpreter {
     /// and the tap's `quit` handler see it.
     ///
     /// Callers must run this only after registering the taps for the rewritten
-    /// markers. The waiter runs on whichever thread resolves the promise (or
-    /// synchronously here when it already has), so it drives a thread clone of
-    /// this interpreter — the same pair `promise_chain_method` uses for `.then`.
+    /// markers. The waiter runs on a pooled worker — woken by whichever thread
+    /// resolves the promise, or, when it is already resolved, reserving its
+    /// place in the group here and running just as asynchronously (#7831) — so
+    /// it drives a thread clone of this interpreter, the same pair
+    /// `promise_chain_method` uses for `.then`.
     pub(crate) fn arm_pending_promise_whenevers(&mut self) {
         for (promise, supplier) in std::mem::take(&mut self.pending_promise_whenever_arms) {
             // The stand-in's serialize group was recorded when the rewritten
@@ -293,7 +295,10 @@ impl Interpreter {
             // Handing it to `on_resolve` makes the resolving thread reserve
             // this reaction's place in the supply block before the pooled
             // worker that runs it is even woken, so N promises resolved in a
-            // row reach the block in resolution order (#7811).
+            // row reach the block in resolution order (#7811). An already
+            // resolved promise reserves its place here instead, so it queues
+            // behind the reactions created before it rather than running
+            // inline ahead of them (#7831).
             let group = match supplier.view() {
                 ValueView::Instance { attributes, .. } => {
                     supplier_id_from_attrs(&attributes.as_map())
--- a/src/value/value_async.rs
+++ b/src/value/value_async.rs
@@ -266,6 +266,8 @@ impl SharedPromise {
     /// already resolved, the callback runs synchronously on the caller's
     /// thread (matching the existing `.then`/`.andthen`/`.orelse` fast path
     /// for an already-kept/broken promise) and this returns `true`.
+    /// (A supply-block reaction is the exception — see
+    /// [`Self::on_resolve_in_supply_group`].)
     /// Otherwise the callback is queued and this returns `false`: some
     /// future `keep`/`try_keep`/`break_with`/`try_break` call will run every
     /// queued waiter, **in registration order**, on a single dedicated
@@ -287,6 +289,19 @@ impl SharedPromise {
     /// the supply block in that order, instead of in whichever order their two
     /// workers happened to wake up in. See [`SupplyTicket`].
     ///
+    /// An **already-resolved** promise takes its ticket here instead, on the
+    /// registering thread, and its reaction is handed to a pooled worker just
+    /// like a queued one. Both paths therefore enter the block through the one
+    /// sequencer, in the order the reactions were created. Running the reaction
+    /// inline, as [`Self::on_resolve`] does for a plain waiter, would let it
+    /// jump ahead of every reaction already ticketed and waiting for a worker
+    /// (#7831); redeeming the ticket inline instead is not an option either,
+    /// because this thread is typically running the enclosing supply block's
+    /// own reaction and so already holds the group lock, so parking it behind
+    /// an earlier ticket — whose worker is itself blocked on that lock — would
+    /// deadlock. So this returns `false` (the reaction has not run yet) even
+    /// when the promise was already resolved.
+    ///
     /// [`SupplyTicket`]: crate::runtime::native_methods::SupplyTicket
     pub(crate) fn on_resolve_in_supply_group(
         &self,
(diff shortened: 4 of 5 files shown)


tokuhirom:
## CI: the `test` job is red, and the failure is inherited from `main`

The failing check is **`test`**, at its bundled-library gate (`scripts/battery-testsuite.sh`), not at the TAP suite or roast:

```
```

**This is not this PR's failure.** Three independent pieces of evidence:

1. **Red on the base branch, identically.** `main` at `a3fb606` — the commit this PR's merge ref is built on — [fails the same `test` job](https://github.com/tokuhirom/mutsu/actions/runs/34429929944) with the *same three files* and the same `287/311` summary. A PR's checks run the merge ref, so this PR inherits it.
2. **Green at this branch's actual parent.** This branch is off `597d8ee`, whose own `main` run had `test` **passing** (its only red job was `jit-stress`). So the battery breakage entered `main` after `597d8ee` — i.e. in `dbcd043` (#7830, vendored `Test` as the default provider), `abcad7f` (#7808), or `a3fb606` — none of which this branch contains.
3. **All three pass locally on this branch.** `MUTSU_BIN=target/release/mutsu scripts/battery-testsuite.sh` here gives `PASS` for `DateTime::Parse 01-basic.t`, `Encode 01-basic.t` and `Log::Async 08-use.rakutest`. (The only local regressions are the six `DBIish` mysql files, which need a MySQL server this container has not got.)

The diff here touches one function, `SharedPromise::on_resolve_in_supply_group`, and `Encode`/`DateTime::Parse` do not use supplies or promises at all — so no path from this change reaches two of the three.

**No fix exists yet to port into this PR**, so there is nothing for me to carry over; widening this PR to chase a `main` breakage it did not cause would be the wrong move. I am not re-running the job either: the failure is deterministic on `main`, not a flake, so a re-run would return the same red.

Next step is on the `main` side. I am root-causing the battery breakage now and will file it as its own issue; once `main` is green again I will update this branch so CI re-runs against the fixed base. Watch stays on until this PR is merged.

---

tokuhirom:
## Root-caused: filed as #7836

The `main` breakage this PR inherits is now root-caused, with a minimal repro, and filed as **#7836**.

All three battery regressions are **one bug**: a package-qualified name fails to resolve inside a string `EVAL`'d from a *different* compunit than the one that `use`d the package. `Test`'s `throws-like` / `use-ok` EVAL their code-string from inside `Test.rakumod`, which never `use`d the module under test.

- `DateTime::Parse` → `Could not find symbol 'DateTime::Parse'`
- `Encode` → `Could not find symbol 'Encode::decode'`
- `Log::Async` → `Could not find symbol 'Terminal::ANSI::OO'`

The `MUTSU_REAL_TEST` toggle isolates it on one binary: the vendored `Test` (the default since #7830) fails, the native TAP provider passes. So #7830 exposed it and #7808's compunit gating is the mechanism. And within `throws-like`'s EVAL, the *imported* name resolves while the *package-qualified* one does not — same sub, same scope — which is the bug in one pair of assertions.

None of that is reachable from this diff, which touches `SharedPromise::on_resolve_in_supply_group` only.

I am **not** widening this PR to fix #7836. Once it lands and `main` is green, I will update this branch so CI re-runs against the fixed base. Watch stays on until then.

---

tokuhirom:
## The blocker is fixed: #7843

All three facets of #7836 are now fixed and **the bundled-library gate is green again** (`285/311`, all three regressions gone). #7843 is up with auto-merge on.

Facet 3 — the one I had left root-caused but unfixed — turned out to be the same idea as the other two. A routine records its declaring file as it is *created*, via `executing_source_file()`, which walks the routine stack for the innermost frame's `def_file`. A module's mainline runs through `run_block` and pushes **no routine frame**, so that walk goes straight past it to whatever routine is still on the stack underneath — a different compunit entirely when the load came from `EVAL "use $module"` inside another module's routine. `declaring_source_file()` applies the same module-load correction `executing_unit_sym_for_module_load()` already uses.

**#7843 carries this PR's commit.** Without the #7831 fix, `t/supply-serialize-fifo.t` test 3 fails 17–37% of runs and the `t/` suite is fatal in CI, so an unrelated red would have been a coin flip on every push there — and waiting for this PR to merge first was not an option while this PR was itself blocked by the gate. It no-ops the moment `main` carries it.

So whichever lands first is fine:

- **#7843 first** → this PR becomes already-merged content; I will close it pointing there, with the #7831 fix landed either way.
- **This PR first** → #7843's merge commit absorbs it with no conflict.

Either way #7831 is fixed on `main`. Watch stays on here until this is merged or closed.

---

tokuhirom:
## Rebased onto the fixed `main`; the blocker is gone

`main` went green while this was blocked: #7842 (`282bd7c`) + `d50f510` fixed the compunit-attribution bug (#7837) that had the bundled-library gate red, and #7844 (`1aab678`) installs `libmysqlclient` from the SessionStart hook.

That was the same bug I had filed as #7836 and fixed on a branch of my own. Since `main` now carries all three of those fixes, I closed **#7843** as redundant rather than reworking six files of conflicts, and **#7836** as a duplicate of #7837.

This PR is back to standing alone: rebased onto `484788a`, force-pushed as `3156261`, one commit, no conflicts. Re-verified on the new base:

- `t/supply-serialize-fifo.t`: **0 failures in 20 runs** (was 11/30 before the fix)
- `make test` — 3951 files, 41999 tests, **PASS**
- `make lint` (all four configurations), `cargo fmt --check` — clean
- `make roast` — clean but for the three documented container-only files
- `scripts/battery-testsuite.sh` — `286/311`, no regressions beyond the six `DBIish` mysql files this container still cannot run (its `libmysqlclient` predates the hook fix)

Auto-merge is on. Watch stays until merged.

---

tokuhirom:
## Rebased again for the `t/` migration; `check-t-layout` is blocked on #7845

`main` moved twice more while this was verifying. Rebased onto `f02f800f` as `b852d2dd` — no conflicts; git followed both renames, so the test edit landed on its new path `t/concurrency/supply/supply-serialize-fifo.t` and the source hunk followed `subtest.rs` → `react_whenever.rs`.

Re-verified on that tree:

- `t/concurrency/supply/supply-serialize-fifo.t`: **0 failures in 20 runs**
- Whole TAP suite via `prove -r`: **3952 files, 42022 tests, PASS**
- Release build clean
- The previous head (`3156261`) had already gone **green on all 7 CI jobs**, `gc-stress` and `jit-stress` included, and the commit's content is unchanged across these rebases

### One inherited red, not this PR's

`make check-t-layout` — a `make test` prerequisite **and** a CI step (`ci.yml:228`) — currently fails on `main` itself:

```
t/any-cool-method-not-found.t sits at t/ top level; every test belongs in a category directory
```

That file came from #7840, which was in flight while #7833 moved `t/` into categories; the two merged minutes apart. It is not in this PR's diff (`git diff --name-only origin/main...HEAD` does not list it).

**#7845 already fixes it** with the one `git mv`, so there is nothing for me to port here — widening this PR with someone else's file move would be the wrong call. This PR's `check-t-layout` goes green as soon as #7845 lands.

Auto-merge stays on; watch continues.

---

tokuhirom:
## CI: red at `t/ layout is valid`, and it is `main`'s, not this PR's — fix up as #7846

The `test` job failed at **step 15, `t/ layout is valid`**. Every step before it passed, **`prove t/` included**, so this PR's own change is fine; the job dies there and skips roast and the bundled-library gate entirely.

```
t/any-cool-method-not-found.t sits at t/ top level; every test belongs in a category directory
```

That file is not in this PR's diff — it is on `main`, added by `a8fa6878`. Two PRs crossed, and neither was wrong about its own diff: `a8fa6878` added the file at top level while `scripts/check-t-layout.sh` still had `MIGRATED=0` (so the rule was off), and #7833 then landed the migration and flipped `MIGRATED=1`. Neither PR's own CI run could see the combination. Reproduced locally on `origin/main` alone.

No fix existed to port — the only other open PR is #7835, unrelated — so I opened one: **#7846**, a pure `git mv` into `t/oo/method/`, the category `scripts/migrate-t-layout.py` names for it. `make check-t-layout` passes again (`3952 tests, all in category directories`, both checkers agreeing) and the test still passes 26/26. That unblocks every open PR, not just this one.

For the record, this PR is otherwise in good shape: the previous run on `3156261` was **green on all 7 jobs**, `gc-stress` and `jit-stress` included, and the current head `b852d2dd` is the same change rebased onto `f02f800f` (the rebase correctly followed `main`'s renames — the pinned test is now `t/concurrency/supply/supply-serialize-fifo.t`, and `src/runtime/subtest.rs` → `src/runtime/react_whenever.rs`). Locally after that rebase: 20/20 clean on the pinned test.

Not re-running the failed job: it is deterministic on the merge ref until #7846 lands, so a re-run returns the same red. Once #7846 merges I will refresh this branch so CI re-runs against a green base.

---

tokuhirom:
## Rebased onto a green `main`; the layout blocker is gone

#7845 landed the `t/` layout fix on `main` (`df5c51c0`) — `t/any-cool-method-not-found.t` is now at `t/oo/method/`, the same destination my #7846 had picked, so I closed that one as redundant rather than landing a no-op rename.

This branch is rebased onto `df5c51c0` and force-pushed as `73332efb`. Re-verified there:

- `make check-t-layout` — OK, both checkers (`3952 tests, all in category directories`)
- `prove -r t/` on the release binary — **3952 files, 42022 tests, PASS**
- the pinned `t/concurrency/supply/supply-serialize-fifo.t` — **0 failures in 20 runs**
- `cargo fmt --all --check` — clean

The rebase again followed `main`'s renames correctly: the pin lives at `t/concurrency/supply/supply-serialize-fifo.t` and `src/runtime/subtest.rs` is now `src/runtime/react_whenever.rs`.

For the record on the earlier red: it was step 15 `t/ layout is valid`, never this PR's change — every step before it passed, `prove t/` included. Rather than wait for the fix to land I should have ported it here immediately; that was the wrong call and cost a cycle. Moot now that `main` carries it.

Auto-merge is on; watch stays until merged.

---

PR #7835, opened 2026-09-10
fix: remove phantom user-instance name method

### LINKED ISSUE #7772: Every instance answers a phantom `.name` method that no class declared
`.name` resolves on any user-class instance even when the class declares no such
attribute — returning the `name` named argument passed at construction, or `Nil`.
raku throws `No such method`.

## Repro

```
class D {}; say D.new(:name("Fido")).name;   # mutsu Fido   raku No such method 'name' for invocant of type 'D'
class D {}; say D.new.name;                  # mutsu Nil    raku No such method
class D { has $.a }; say D.new(:a(1), :name("x")).name;   # mutsu Nil   raku No such method
```

## Controls that are CORRECT — this is specific to the name `name`

```
class D {}; say D.new(:x(1)).x;         # No such method 'x'     in BOTH
class D {}; say D.new(:value(7)).value; # No such method 'value' in BOTH
class C { has $.a }; say C.new(:a(1), :zzz(9)).zzz;  # No such method 'zzz' in BOTH
```

So unknown named arguments are *not* generally turned into attributes — good —
and unknown method names are *generally* rejected. It is `name` in particular
that has a fallback it should not have, presumably a leftover from a built-in
type's `.name` (enum values, `Attribute`, `Routine`, `IO::Path`… all legitimately
have one) leaking into the generic instance path.

## Why it matters

A phantom accessor is worse than a missing one: `$obj.name` silently answers
`Nil` where the program meant to fail, so a typo or a refactor that removes an
attribute goes unnoticed. It also shadows `Mu`'s "did you mean" suggestion.

## Doc-diff origin

`raku-doc/doc/Language/py-nutshell.rakudoc:582`, found by the error-parity
oracle mode added on 2026-09-09.

## Acceptance

- All three repro lines throw `No such method 'name' for invocant of type 'D'`.
- Types that legitimately have `.name` keep it: enum values, `Attribute`,
  `Routine`/`Sub`/`Method`, `IO::Path`, `Thread`, `Distro`/`Kernel`.
- A user class that declares `has $.name` still gets its own accessor.
- A `t/` pin covering all three: absent, declared, and built-in `.name`.

tokuhirom:
Closes #7772

User instances now report X::Method::NotFound for undeclared .name calls, while declared accessors and legitimate built-in name methods remain available.


tokuhirom:
Follow-up validation for b054baf3f:

- Local `make lint`, `make test`, and `make roast` pass.
- The bundled-library gate reports 287/311 passing files. zef and all Template::Mustache tests now pass.
- The only three failures are `DateTime::Parse/01-basic.t`, `Encode/01-basic.t`, and `Log::Async/08-use.rakutest`, exactly matching the latest `main` CI baseline run 34429929944.

No new bundled-library regression remains attributable to this PR; the required `test` check is blocked by the pre-existing main baseline.

Issue #7838, opened 2026-09-10
t/supply-serialize-fifo.t subtest 3 is non-deterministic on a 4-core box


tokuhirom:
## What happens

`t/supply-serialize-fifo.t`'s third subtest — "already-kept and later-kept promises interleave in value order" — fails on roughly half the runs in a 4-core remote container. It is **not** in `flaky-tests.txt`, so a run that rolls badly fails `make test` outright.

```
not ok 3 - already-kept and later-kept promises interleave in value order
# expected: '1 2 3 4 5 6'
#      got: '1 2 3 4 6 5'      <- most common
#      got: '1 2 4 6 3 5'
#      got: '2 4 6 1 3 5'
#      got: '2 1 3 4 5 6'
```

## Not a regression — measured on `main`

Two release binaries, same container (4 cores), the file run directly, 8 runs each:

| binary | result |
| --- | --- |
| `main` @ `a3fb606` | **5 / 8 fail** |
| a feature branch off it | 4 / 8 fail |

So the branch is not the cause; `main` itself is where this lives. The file was added by `9d59f4c` ("order a supply block's reactions by the order their sources produced them", #7811) and presumably passes reliably on the 12-core dev box, which is why CI has not caught it.

## Repro

```
cargo build --release
for i in $(seq 1 8); do
  ./target/release/mutsu t/supply-serialize-fifo.t | grep -E '^(not )?ok 3'
done
```

Subtests 1 and 2 (`run-nested(3)`, `run-nested(20)`) pass every time. Only the third shape fails — the one where a promise is kept *before* its nested `whenever` is registered for odd values and *after* for even ones:

```raku
whenever $src -> $v {
    my $p = Promise.new;
    $p.keep if $v %% 2;      # already kept when the whenever registers
    whenever $p { emit $v }
    $p.keep unless $v %% 2;  # kept after
}
```

The `got` values are the tell: `1 2 4 6 3 5` and `2 4 6 1 3 5` are the even (already-kept, so run synchronously on the registering thread) and odd (later-kept, so dispatched to a pooled worker) halves separating out. #7811 serialized the pooled-worker path against itself, but the synchronous already-kept path can still overtake an earlier value's pending continuation.

## Why it matters

[...]

PR #7839, opened 2026-09-10
refactor(test): retire the native Test TAP provider

### LINKED ISSUE #7566: Retire the native `Test` TAP provider
**Blocked on the flip.** This ticket only becomes actionable once the vendored
upstream `Test.rakumod` is what a bare `use Test` resolves to — see
`todo/deep/vendor-real-test-module-flip.md`, which was attempted on 2026-09-07/08
and withdrawn on four bundled-library regressions. Until then the native provider
is still the default and cannot be deleted; do not start here.

mutsu ships **two** TAP implementations today, with the vendored one reachable
through `MUTSU_REAL_TEST=1`. Keeping two production providers indefinitely is
exactly what the vendoring campaign existed to end, so the sequence is: fix the
four gate root causes, flip the default, then do the deletion below.

## What comes out

| | lines |
| --- | --- |
| `src/runtime/test_functions/` (7 files) | ~2 400 |
| `src/runtime/subtest.rs` | 700 |
| `src/vm/vm_native_test.rs` | 83 |

plus:

- `Interpreter::real_test_module_enabled()` (`src/runtime/runtime_module.rs`)
  and its seven call sites — `runtime_module.rs` (the `use Test` no-op and the
  `register_native_provider_exports` call), `builtins_operators_fallback.rs`,
  `calls.rs`, `accessors_resolve.rs`, `vm_call_dispatch.rs`,
  `vm_native_test.rs`. Each one becomes "the real module owns this", i.e. the
  branch it guards simply goes away.
- `TEST_MODULE_EXPORTS` and `register_native_provider_exports("Test", ...)`:
  the real module runs its own `is export` declarations, so nothing has to
  synthesize its export stash.
- The `__mutsu_backend_todo__:` marker in `run_roast_preprocess.rs` and its
  consumer in `call_helpers.rs` (`test_ok_with_diag`). Only the native provider
  ever understood it; the vendored path already emits the bare reason.
- `scripts/test-module-sweep.sh` and `scripts/roast-test-module-sweep.sh` —
  both are dual-provider comparisons with nothing left to compare against.
  Delete them with the switch, in the same change, so neither is left silently
  measuring one provider twice.
- `t/vendored-real-test-module.t`'s native half (subtests 3 and 4) and its
  `%*ENV<MUTSU_REAL_TEST>` steering. The file should keep asserting that the
  vendored module is what answers, without a second implementation to contrast
  with.
- `user_test_decl_beats_native` (`runtime/calls.rs`) and the second dispatch
  path it guards in `builtins_operators_fallback.rs`: both exist to let an
  imported declaration win over a native handler that will no longer exist.

## What must NOT come out with it

`Test::Util` is a **roast helper**, loaded from
`roast/packages/Test-Helpers/lib/Test/Util.rakumod`, not part of this provider.
Its own native overrides are a separate retirement, tracked by
`todo/tickets/retire-native-test-util-overrides.md`. `is_run`,
`doesn't-hang`, `make-temp-dir` and friends stay where they are.

## Why this is a `deep/` item and not a ticket

The seven gates are not independent: `calls.rs` and
`builtins_operators_fallback.rs` are two different dispatch entries into the
same handlers, `accessors_resolve.rs` synthesizes `&is-deeply`-style Routine
values that only exist because the natives are not declared subs, and
`runtime_module.rs` decides whether `use Test` is a module load at all. Removing
one at a time leaves the provider half-reachable in states nothing tests. It
wants one change, validated by `make test` + `make roast` on the default
provider, with `cargo test --lib` for the in-process TAP unit tests that today
assert native output shapes (`run_dist.rs`'s `rakudo_todo_*`, `runtime/mod.rs`'s
`test_more_tests_arg_emits_plan`) — those assertions collapse to the single
remaining provider's output.

## Before starting

Run both sweeps one last time on the current `main` and record the result in
the retirement's news entry: it is the last moment the two providers can be
compared, and the numbers are what justify deleting the loser.

---
*Migrated verbatim from `todo/deep/retire-the-native-test-provider.md` on 2026-09-08, when the in-repo `todo/` directory was retired in favour of GitHub issues. See [`docs/issue-workflow.md`](https://github.com/tokuhirom/mutsu/blob/main/docs/issue-workflow.md).*

tokuhirom:
mutsu shipped **two** TAP implementations. Since the flip (#7554) a bare `use Test` has loaded rakudo's own `Test.rakumod` (vendored to `modules/Rakudo-Core/lib/`), and the Rust one survived only behind `MUTSU_REAL_TEST=0`, as the escape hatch the dual-provider sweeps drove. This deletes it, and the switch with it: there is one `Test` under mutsu now, and it is the module rakudo ships.

**+516 / −4718 across 84 files.**

Closes #7566

## The last comparison

Both sweeps were run one final time on `main` at `a3fb606` (release build, same container), since the two providers can never be compared again:

| sweep | pass under both | regressed under the real `Test` | passes only under the real `Test` | fail under both |
| --- | --- | --- | --- | --- |
| `scripts/test-module-sweep.sh` (3 950 `t/` files) | 3 926 | **0** | 1 | 23 |
| `scripts/roast-test-module-sweep.sh` (1 436 whitelisted roast files) | 1 432 | **0** | 0 | 4 |

Nothing the native provider could do is lost. The "fail under both" rows are sweep artifacts, not failures: the `t/` sweep copies every file into one flat working directory, which breaks the files that resolve fixtures by their own path (`use lib`, `$*PROGRAM.parent(2)`, the whole `repo-*` / `use-dist-*` / `nativecall-*-in-module` family), and the roast sweep forces `MUTSU_PRECOMP=0`, which is exactly what `S10-packages/precompilation.t` measures. The other three roast rows are the documented container-environment set.

Both sweep scripts go with the provider they compared — a dual-provider sweep with one provider silently measures the same half twice.

## What comes out

| | lines |
| --- | --- |
| `src/runtime/test_functions/` (7 files) | 2 933 |
| `src/vm/vm_native_test.rs` | 83 |
| the subtest halves of `src/runtime/subtest.rs` | 153 |
| the two sweep scripts | 211 |

plus everything that existed only to keep two providers apart:

- `Interpreter::real_test_module_enabled()` and its seven gates — the `use Test` no-op in `runtime_module.rs`, the two dispatch entries (`exec_call_sanitized` and `call_function_fallback`), the `&is-deeply`-style Routine synthesis in `accessors_resolve.rs`, `is_interpreter_handled_function`, the VM's `try_native_test_function`, and the fudge preprocessor's provider branch;
- `user_test_decl_beats_native`, whose only job was to let an imported declaration win over a native handler;
- `TEST_MODULE_EXPORTS` and `register_native_provider_exports("Test", …)` — the real module runs its own `is export` declarations, so `t/export-default-stash.t` now passes because `Test` reaches `Test::EXPORT::DEFAULT` the way every source module always did, not because mutsu pre-populated it;
- the `__mutsu_backend_todo__:` marker and its consumer in `test_ok_with_diag`;
- `collect_eval_imported_function_names` and the whole EVAL imported-function preseed chain it fed;
- `Stmt::Subtest` / `OpCode::SubtestScope` / `exec_subtest_scope_op` — a dedicated statement form for `subtest NAME => { … }` that ran the native subtest machinery. It was **already unreachable**: `subtest_stmt` parses its name with `expression`, which consumes the `=>` itself, so the rule never matched and the form has been compiling to an ordinary `Call` all along.

### The TAP state collapsed with it

`tap_state.rs` held the provider's bookkeeping, and the dead-code lint says what is left: the plan counter, the pending-`todo` ranges, the subtest stack and the bail-out flag have no producer any more — rakudo's `Test` keeps all of that in its own Raku-level state. `active()` and the shared cross-thread counter survive.

That took three consumers with it, all of which were asking a permanently-zero `subtest_depth`:

- `OutputSink::emit` / `emit_stderr` lose their `subtest_active` parameter, and `Interpreter::subtest_active()` goes with them — a subtest's output is indented by the module's own output handles now;
- `Thread.start` and the thread-clone path no longer suppress immediate stdout for "a thread spawned inside a subtest";
- `finish()`'s trailing TAP footer is deleted. Verified against the vendored module's own `END` phaser: `plan 2; ok 1, "a"` still exits 255 with `# You planned 2 tests, but ran 1`, a failing assertion still exits 1, and `bail-out` still prints `Bail out!` and exits 255.

## Two things surfaced along the way, and are fixed here

**1. The parse-time export list has to stay, for a measured reason.** Deleting `TEST_MODULE_EXPORTS` also deleted the parser's fast path for `use Test`, sending `register_module_exports` down `find_and_scan_module` — a full parse of the vendored module's 953 lines, once per process, on top of the runtime's own (precompilation-cached) load. Measured on release, `mutsu -e 'use Test; plan 1; ok 1, "x"'` went **35 ms → 93 ms**, and every `t/` and roast file pays it. So the list is kept, moved to `parser::stmt::simple::module_exports::TEST_EXPORTS` where it belongs, with a comment saying it is a speed shortcut for a bundled module rather than a native provider's export surface. Startup is back at parity (36 ms).

Unlike the constant it replaces it cannot drift: `test_exports_match_the_vendored_module` re-derives it from `Test.rakumod` on every `cargo test`. That guard immediately found three names the old list was missing (`MONKEY-SEE-NO-EVAL`, `exit-ok`, `trait_mod:<is>`) and two it invented (`force_todo` / `force-todo`, which rakudo's `Test` does not export).

**2. A real bug in `collect_eval_user_sub_names`.** It took everything after the last `::`, so a multi candidate keyed `Pkg::name/arity…` reached the parse-time preseed as `is/2` rather than `is`. EVAL'd code calling any imported multi in listop form (`is [$sub()], [42], 'desc'`) therefore parsed its first argument as a subscript. This was masked for `Test` by the hardcoded list and is live for every other module's multis; `eval_q_bracket_statement_list_runs_declaration_then_assertion` is the pin.

## What did NOT come out

`Test::Util` is a roast helper (`roast/packages/Test-Helpers/lib/Test/Util.rakumod`), not part of this provider — `is_run`, `doesn't-hang`, `make-temp-dir` and friends are untouched. `Interpreter::is_test_function_name` stays too, but as a *recognition* list rather than a dispatch table: the static known-routine tables (`runtime/undeclared_routines.rs`, the ADR-0065 analysis frontend) use it so a unit they never run does not report `is-deeply` as undeclared, and the bare-word resolver uses it to route a hyphenated zero-argument call through function dispatch.

## Verification

Run locally on this branch before opening:

| gate | result |
| --- | --- |
| `cargo fmt --all` | clean |
| `make lint` (4 configurations) | exit 0, zero warnings |
| `cargo test` | 1016 lib tests + workspace, all pass |
| `prove -r t/` (release) | Files=3950, Tests=41989 — **1 failing file**, see below |
| `make roast` | Files=1436, Tests=218939 — **3 failing files**, the documented container set |

The three roast failures are exactly `docs/agent-environments.md`'s list, matching it row for row: `roast/6.c/S32-io/file-tests.t` (`Failed: 4`, tests 6-8/10, `uid 0`), `roast/S16-filehandles/filetest.t` (`Failed: 25`, `uid 0`), `roast/S32-io/IO-Socket-Async.t` (exit 124, ran 17, sandboxed network).

The one `t/` failure is `t/supply-serialize-fifo.t` subtest 3, and it is **not this change**: measured 8 runs each on the same container, `main` @ `a3fb606` fails it **5/8** and this branch 4/8. It is a pre-existing non-deterministic ordering hole in the #7811 fix, filed separately as #7838 with the evidence; it is not in `flaky-tests.txt`, so CI here may roll red on it.

https://claude.ai/code/session_019NLxutW34JwgK9bmKFpgeF

---
_Generated by [Claude Code](https://claude.ai/code/session_019NLxutW34JwgK9bmKFpgeF)_


tokuhirom:
## CI: the `test` job is red, and it is not this PR

The failing check is **`test` → `Bundled-library test suites`**. It is red on the **base branch too**, at the same step, on the same three files.

`main` @ `a3fb606` — the commit this PR branches from — [run 34429929944, job 102723932395](https://github.com/tokuhirom/mutsu/actions/runs/34429929944/job/102723932395):

```
```

This PR — [run 34436333056, job 102742082056](https://github.com/tokuhirom/mutsu/actions/runs/34436333056/job/102742082056):

```
=== summary: 288/311 test files pass (3 excluded) ===
```

Same three files; this branch passes **288/311 against `main`'s 287/311**, one file more, not fewer. Every other step of the `test` job is green on both, and `lint-configs`, `gc-stress`, `jit-stress` and `wasm-e2e` are green here.

Reproduced locally against both binaries built from the same container, running the gate's own invocation:

```
$ cd <perl6-encode@f61acc36> && mutsu -I $ROOT/modules/Encode/lib t/01-basic.t
not ok 2 - right exception type (Encode::X::Encode::Unknown)
# Got:      X::AdHoc
# Exception message: Could not find symbol 'Encode::decode'
```

Byte-identical output from a `main` build and from this branch. `raku` passes the same file 3/3.

## Root cause, and no fix exists yet

`EVAL $code, context => $ctx` records the *lexically enclosing* compunit as the EVAL unit's parent (`builtins_eval_misc.rs:396`, `note_eval_unit_parent(unit_sym, self.current_unit)`), ignoring `context`. #7797's qualified-name gate then judges the snippet against the grants of the compunit that called `EVAL`. Rakudo's `Test.rakumod` — what `use Test` loads since #7554 — EVALs `throws-like`'s code string with `context => $caller-context`, so the snippet inherits `Test`'s grants and any `throws-like 'Pkg::sub(...)'` dies with `Could not find symbol`.

It is not a `Test` problem: the same failure reproduces with two ordinary modules and no `Test` anywhere. Filed as #7841 with that repro and a proposed patch. I searched open issues and PRs and there is no fix to port into this branch, so I am standing down on it here rather than widening a provider-deletion PR with an unrelated EVAL-scoping change.

Not re-running the job: the failure is deterministic and `main`'s own CI log already settles the question a re-run would have asked.

This PR stays watched until it is green and merged; the blocker is #7841 landing on `main`.

---

Issue #7836, opened 2026-09-10
main is red: a package-qualified name does not resolve inside an EVAL made from another compunit (Test's throws-like/use-ok), breaking 3 batteries


tokuhirom:
## Summary

`main` is currently **red on the `test` job**, at the bundled-library gate. Every PR inherits it, because a PR's checks run the merge ref:

```
=== summary: 287/311 test files pass (3 excluded) ===
  REGRESSION: whitelisted 'DateTime::Parse	01-basic.t' did not pass this run
  REGRESSION: whitelisted 'Encode	01-basic.t' did not pass this run
  REGRESSION: whitelisted 'Log::Async	08-use.rakutest' did not pass this run
GATE FAILED: a bundled library regressed below its recorded baseline.
```

All three have **one root cause**: a **package-qualified name** fails to resolve inside a string `EVAL`'d from a *different* compunit than the one that `use`d the package. `Test`'s `throws-like` / `use-ok` / `lives-ok` EVAL their code-string argument from inside `Test.rakumod`, which never `use`d the module under test.

- `DateTime::Parse 01-basic.t` → `Could not find symbol 'DateTime::Parse'`
- `Encode 01-basic.t` → `Could not find symbol 'Encode::decode'`
- `Log::Async 08-use.rakutest` → `Could not find symbol 'Terminal::ANSI::OO'`

## Where it came from

| main commit | `test` job |
|---|---|
| `597d8ee` (#7826) | **green** (only `jit-stress` red) |
| `dbcd043` (#7830 — vendored `Test` as the default provider) | run cancelled |
| `abcad7f` (#7808 — gate package-qualified names by which compunit `use`d the package) | run cancelled |
| `a3fb606` | **red**, these three |

Two commits interact, and the toggle proves it — same binary, same script:

```
$ ./target/release/mutsu tmp/repro-min.raku            # vendored Test (default since #7830)
not ok 2 - throws-like resolves Widget::make too
    # Got:      Could not find symbol 'Widget::make'

$ MUTSU_REAL_TEST=0 ./target/release/mutsu tmp/repro-min.raku    # native TAP provider
ok 2 - throws-like resolves Widget::make too
```

So **#7830 exposed it** (it moved `throws-like`'s `EVAL` from native Rust into a real Raku compunit) and **#7808's compunit gating is the mechanism**. Neither is wrong on its own; the gate is just too strict for an `EVAL` that should compile in its *caller's* lexical scope.

## Minimal repro

[...]

PR #7840, opened 2026-09-10
fix: an undefined Any does not answer Cool's methods

### LINKED ISSUE #7773: A known method name called on an undefined value returns `(Any)` instead of throwing
Calling a method that exists on some other type, on an **undefined** value,
quietly answers `(Any)` instead of throwing `No such method`.

## Repro

```
my $x; say $x.comb(/\w+/);   # mutsu (Any)   raku No such method 'comb' for invocant of type 'Any'
```

The doc example that found it shows the shape that matters — the bad value
travels on into a data structure:

```
my $formatted-text;
my @words = "Abe", "Lincoln";
@words.push("said", $formatted-text.comb(/\w+/));
say @words;
```
- raku:  `No such method 'comb' for invocant of type 'Any'`
- mutsu: `[Abe Lincoln said (Any)]`

## Control that is CORRECT

```
my $x; say $x.no-such-method;   # No such method 'no-such-method' for invocant of type 'Any'   in BOTH
```

So the `No such method` path exists and works for a name nothing defines. What
goes wrong is a name that IS defined elsewhere (`comb` on `Cool`): the dispatch
apparently resolves it on the undefined invocant and returns a type object
rather than refusing.

## Doc rows this covers (2026-09-09 sweep)

`Language/objects.rakudoc:48`, and the `Language/perl-nutshell.rakudoc` cluster
at `:86`, `:200`, `:816`, `:1011`, `:1027`, `:1260`, `:1268`, which all call a
method on an undeclared (hence `Any`) variable and expect raku's diagnostic.

Found by the error-parity oracle mode added on 2026-09-09 — every one of these
blocks used to be discarded as "no oracle" because raku exits non-zero on them.

## Acceptance

- Both repros throw, with raku's message and exception type.
- Methods that *are* defined on `Mu`/`Any` for a type object still work
  (`.defined`, `.^name`, `.gist`, `.raku`, `.Str` where raku allows it) — the
  fix must not turn every type-object method call into an error.
- A `t/` pin covering: a `Cool` method on `Any`, an unknown method on `Any`,
  and the allowed `Mu` methods as controls.

tokuhirom:
`Any` is not a `Cool` — it is the other way round — so `uc`, `chars`, `comb`, `abs`, `substr` and the rest of that surface do not exist on an undefined invocant. raku throws; mutsu answered a **value**.

```
my $x; say $x.comb(/\w+/);   # raku: No such method 'comb' for invocant of type 'Any'
                             # mutsu: (Any)
```

## Root cause

By-name native dispatch. Its cascades recognize the method NAME, stringify the receiver and answer out of that string — and a type object stringifies to its gist. So `$x.comb(/\w+/)` combed the literal text `"(Any)"` and returned `("Any",)`. Likewise `.uc` → `"(ANY)"`, `.chars` and `.bytes` → `5`, `.IO` → an `IO::Path` for a file named `(Any)`.

That is worse than a missing method: the wrong value does not stop, it travels. The documentation example that turned it up (`Language/objects.rakudoc:48`, plus a cluster of seven rows in `perl-nutshell.rakudoc`) pushes it straight into an array.

```raku
my $formatted-text;
my @words = "Abe", "Lincoln";
@words.push("said", $formatted-text.comb(/\w+/));
say @words;                      # raku: dies.  mutsu: [Abe Lincoln said (Any)]
```

## Fix

This is [ADR-0051](https://github.com/tokuhirom/mutsu/blob/main/docs/adr/0051-type-ancestry-has-one-oracle-and-an-unresolved-method-throws.md)'s **P4 rule** — "an unresolved `Cool`-only method throws rather than stringifying the receiver" — applied to the one receiver shape its gates never saw. Those gates sit inside a `ValueView::Instance` arm, so the `Any`/`Mu` **type object** walked straight past them into the cascades.

`src/runtime/any_cool_method_gate.rs` is the missing arm, wired into **both** dispatch entries: `call_method_with_values` (the interpreter) and `try_native_method` (the opcode path, which reaches the cascades without passing through the first — the 0-arg `.uc` shape proved it needed its own gate).

The gated name set is the union of two independently raku-verified sources, and neither subsumes the other:

| source | contributes |
| --- | --- |
| ADR-0051's curated `cool_only_builtin_method` | names with no `Cool` row of their own — `bytes`, `lazy`, `race`, `Date`, `parse-base` |
| the generated row catalog (`Cool` rows − `Any`/`Mu` rows) | rows the list predates — `printf`, `is-prime`, `indices`, the native-int coercion family |

`Any.^can` — rakudo's **and mutsu's** — is empty for all 109 of them. That last point is the interesting one: mutsu's introspection already knew `Any` has no `comb`; only dispatch disagreed.

`match` and `split` are deliberately excluded. raku declares both on `Any`; their candidates merely refuse an *undefined* invocant, so raku answers `X::Multi::NoMatch` / a use-of-uninitialized warning for them, never `X::Method::NotFound`. Gating them would swap one wrong answer for another.

## Verification

| gate | result |
| --- | --- |
| `cargo fmt --all --check` | clean |
| `make lint` (all four configurations) | exit 0, zero warnings |
| `cargo test` | 1018 lib tests, all pass |
| `make test` | Files=3951, Tests=42018 — one failing file, see below |
| `make roast` | Files=1436, Tests=218939 — only the three documented container failures |

The three roast failures match `docs/agent-environments.md` row for row: `roast/6.c/S32-io/file-tests.t` (`Failed: 4`, tests 6-8/10, `uid 0`), `roast/S16-filehandles/filetest.t` (`Failed: 25`, `uid 0`), `roast/S32-io/IO-Socket-Async.t` (exit 124, sandboxed network).

The one `t/` failure is `t/supply-serialize-fifo.t` subtest 3, which is **not this change** — it is [#7838](https://github.com/tokuhirom/mutsu/issues/7838) / [#7831](https://github.com/tokuhirom/mutsu/issues/7831), measured at ~5/8 failures on `main` on a 4-core container and already fixed by the open [#7834](https://github.com/tokuhirom/mutsu/pull/7834). It passed on a direct re-run here.

`t/any-cool-method-not-found.t` (26 assertions, green under `raku` as well as mutsu) pins the repro and its exception's `.method`/`.typename`, the doc example including that nothing reaches the array, nine more `Cool` names, the `Mu` twin, an unknown name as the control that the ordinary `X::Method::NotFound` path is unchanged, `Any`'s own surface (`.defined`, `.gist`, `.raku`, `.^name`, `.WHAT`, `.WHICH`, `.elems`) and a defined `Str`/`Int` receiver.

A **184-name sweep** of `my $x; $x.<name>` against rakudo v2026.07 confirms the change introduces no new divergence: the seventeen names that still disagree are all pre-existing gaps of a different kind (a zero-argument call of a method that needs arguments, where raku raises `X::Multi::NoMatch`), and none of them is in the gated set.

Closes #7773

https://claude.ai/code/session_01LZT6GFNwTpdVShUWFZxeM8

---
_Generated by [Claude Code](https://claude.ai/code/session_01LZT6GFNwTpdVShUWFZxeM8)_

### DIFF
--- a/news/2026-09/any-is-not-a-cool-method-gate.md
+++ b/news/2026-09/any-is-not-a-cool-method-gate.md
@@ -1,0 +1,72 @@
+# A `Cool` method on an undefined `Any` throws instead of answering `(Any)`
+
+`Any` is not a `Cool` — it is the other way round — so `uc`, `chars`, `comb`,
+`abs`, `substr` and the rest of that surface simply do not exist on an
+undefined invocant. raku says so:
+
+```
+my $x; say $x.comb(/\w+/);   # No such method 'comb' for invocant of type 'Any'
+```
+
+mutsu answered `(Any)`.
+
+The mechanism was mutsu's by-name native dispatch. Its cascades recognize the
+method *name*, stringify the receiver and answer out of that string; a type
+object stringifies to its gist, so `$x.comb(/\w+/)` combed the literal text
+`"(Any)"` and returned `("Any",)`. `.uc` answered `"(ANY)"`, `.chars` answered
+`5`, `.bytes` answered `5`, `.IO` handed back an `IO::Path` for a file named
+`(Any)`.
+
+That is worse than a missing method, and the documentation example that turned
+it up (`raku-doc/doc/Language/objects.rakudoc:48`, plus a cluster of seven rows
+in `perl-nutshell.rakudoc`) shows why — the wrong value does not stop, it
+travels:
+
+```raku
+my $formatted-text;
+my @words = "Abe", "Lincoln";
+@words.push("said", $formatted-text.comb(/\w+/));
+say @words;                      # raku: dies.  mutsu: [Abe Lincoln said (Any)]
+```
+
+## The fix
+
+This is [ADR-0051](../../docs/adr/0051-type-ancestry-has-one-oracle-and-an-unresolved-method-throws.md)'s
+P4 rule — "an unresolved `Cool`-only method throws rather than stringifying the
+receiver" — applied to the one receiver shape its gates never saw. Those gates
+sit inside a `ValueView::Instance` arm, so the `Any`/`Mu` **type object** walked
+straight past them into the cascades. `src/runtime/any_cool_method_gate.rs` is
+the missing arm, wired into both dispatch entries: `call_method_with_values`
+(the interpreter) and `try_native_method` (the opcode path, which reaches the
+cascades without passing through the first).
+
+The gated name set is the union of two independently raku-verified sources, and
+neither subsumes the other: ADR-0051's hand-curated `cool_only_builtin_method`
+list carries names with no `Cool` row of their own (`bytes`, `lazy`, `race`,
+`Date`), and the generated row catalog carries rows the list predates
+(`printf`, `is-prime`, the native-int coercion family). `Any.^can` — rakudo's
+and mutsu's alike — is empty for every name in either. That last point is the
+interesting one: mutsu's introspection already knew `Any` has no `comb`, and
+only dispatch disagreed.
+
+`match` and `split` are deliberately excluded. raku declares both on `Any`;
+their candidates merely refuse an undefined invocant, so raku answers
+`X::Multi::NoMatch` / a use-of-uninitialized warning for them, never
+`X::Method::NotFound`. Gating them would swap one wrong answer for another.
+
+## Verification
+
+`t/any-cool-method-not-found.t` pins the repro, the doc example (including that
+nothing reaches the array), nine more `Cool` names, the `Mu` twin, an unknown
+name as the control that the ordinary `X::Method::NotFound` path is unchanged,
+and `Any`'s own surface (`.defined`, `.gist`, `.raku`, `.^name`, `.WHAT`,
+`.WHICH`, `.elems`) plus a defined `Str`/`Int` receiver. All 26 assertions pass
+under `raku` as well as mutsu.
+
+A 184-name sweep of `my $x; $x.<name>` against rakudo v2026.07 confirms the
+change introduces no new divergence: the seventeen names that still disagree are
+all pre-existing gaps of a different kind (a zero-argument call of a method that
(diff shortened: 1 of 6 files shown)


tokuhirom:
## CI: the `test` job is red on the bundled-library gate, and it is not this PR's

```
=== summary: 288/311 test files pass (3 excluded) ===
```

That is [#7836](https://github.com/tokuhirom/mutsu/issues/7836) verbatim — same three files, same message — filed against `main` at `a3fb606`, which is this PR's base. Every PR inherits it through the merge ref.

**The failures are a different mechanism from this diff.** Reproduced the gate locally on this branch (`BATTERIES_LOCK`/`BATTERIES_WHITELIST` narrowed to the three), then ran each failing file directly:

| file | exception message |
| --- | --- |
| `Encode/t/01-basic.t` | `Could not find symbol 'Encode::decode'` |
| `DateTime::Parse/t/01-basic.t` | `Could not find symbol 'DateTime::Parse'` |
| `Log::Async/t/08-use.rakutest` | `Could not find symbol 'Terminal::ANSI::OO'` |

All three are `X::AdHoc: Could not find symbol` raised inside `throws-like`'s cross-compunit `EVAL` — #7836's package-qualified-name gating. This PR raises `X::Method::NotFound` for a `Cool` method on an `Any`/`Mu` **type object**; `Encode::decode` is a package-qualified *sub*, not a method, and no `No such method` appears anywhere in the three failures.

Two further points against it being this PR's: the gate reports **288**/311 here versus **287**/311 in #7836's own report of `main`, i.e. one file more passing, not fewer; and `make roast` (1436 files, 218939 tests) is clean on this branch apart from the three documented container-only failures.

**No fix exists to port yet.** #7836 is claimed and actively being worked in another session (branch `claude/mutsu-issue-7836-evalscope`), with no PR pushed as of now, so there is no change to bring into this branch. Writing a second fix for it here would duplicate that claimed work and widen this PR.

Not re-running the job: the failure reproduces deterministically on this branch locally (not a flake), so a re-run buys nothing. Once #7836 lands on `main` I will merge `main` into this branch to re-run CI against the fixed base. Keeping this PR watched until then.

---

Issue #7837, opened 2026-09-10
`EVAL $code, context => CALLER::` ignores the context, breaking vendored throws-like (battery gate red on main)


tokuhirom:
## Summary

`EVAL`'s `context` named argument is ignored: the code is evaluated in the scope of the routine calling `EVAL`, not the context handed to it. That makes the vendored `Test.rakumod`'s string form of `throws-like` unable to see the caller's imports, and it is **failing the battery gate on `main` right now**:

```
REGRESSION: whitelisted DateTime::Parse/01-basic.t no longer passes
REGRESSION: whitelisted Encode/01-basic.t no longer passes
REGRESSION: whitelisted Log::Async/08-use.rakutest no longer passes
GATE FAILED: a bundled library regressed below its recorded baseline.
```

This surfaced when the vendored upstream `Test` module became the default provider (da21ca091, #7554). The native provider did not take this path, so the gap was invisible until the default flipped.

## Root cause

`modules/Rakudo-Core/lib/Test.rakumod:640`:

```raku
my $caller-context = $*THROWS-LIKE-CONTEXT // CALLER::; # Don't guess our caller context, know it!
...
EVAL $code, context => $caller-context;
```

mutsu evaluates `$code` in `Test.rakumod`'s own scope instead of the caller's, so any symbol the *test file* imported is unresolvable.

## Minimal repro

`lib/EvalHelper.rakumod`:

```raku
unit module EvalHelper;
use MONKEY-SEE-NO-EVAL;
sub eval-in-caller(Str $code) is export {
    my $caller-context = CALLER::;
    EVAL $code, context => $caller-context;
}
```

`probe.raku`:

```raku
use lib 'lib', 'modules/Encode/lib';
use EvalHelper;
use Encode;
say (try eval-in-caller('Encode::decode("nyi", buf8.new(97))')) // "died: {$!.^name}: {$!.message}";
```

| | output |
| --- | --- |
| `raku` v2026.07 | `died: Encode::X::Encode::Unknown: Unknown encoding nyi.` |
| `mutsu` | `died: X::AdHoc: Could not find symbol 'Encode::decode'` |

Rakudo resolves `Encode::decode` through the passed context; mutsu does not. Dropping the `context` argument makes **both** fail with a symbol error, which confirms the argument is the whole difference rather than some other scoping detail.

## What it looks like in the battery gate

`Encode/t/01-basic.t` is a two-line file whose only assertion is a string `throws-like`:

```
[...]

PR #7843, opened 2026-09-10
fix: run code in the compunit it was written in (EVAL context, END phasers, module routines)

### LINKED ISSUE #7836: main is red: a package-qualified name does not resolve inside an EVAL made from another compunit (Test's throws-like/use-ok), breaking 3 batteries

tokuhirom:
## Problem

`main` is red on the `test` job's bundled-library gate, so **every PR inherits it**:

```
  REGRESSION: whitelisted 'DateTime::Parse	01-basic.t'   # Could not find symbol 'DateTime::Parse'
  REGRESSION: whitelisted 'Encode	01-basic.t'            # Could not find symbol 'Encode::decode'
  REGRESSION: whitelisted 'Log::Async	08-use.rakutest'   # Could not find symbol 'Terminal::ANSI::OO'
```

Two commits interact to produce them. #7797 gated package-qualified names by which compunit `use`d the package; #7830 made the vendored upstream `Test` the default provider, which moved `throws-like`'s and `use-ok`'s `EVAL` out of native Rust into a real Raku compunit where that gate applies. Neither is wrong.

What was missing is one idea in three places: **several things that *run* "somewhere else" still belong, lexically, to the compunit that *wrote* them**, and the gate was reading the wrong one for each.

## Fix 1 — a context EVAL inherits the context's compunit

`EVAL $code, context => $ctx` compiles the snippet as if it stood at `$ctx`'s frame. mutsu honoured that for the *package* and for ADR-0037's `return` classification, but not for the compunit — and a compunit is part of a lexical scope just as much as a package is: which modules it `use`d is what decides the qualified names its code may write.

So the EVAL unit's parent was whichever compunit was *running* the `EVAL`. `Test.rakumod` passes `CALLER::` precisely to avoid that (*"Don't guess our caller context, know it!"*), and mutsu ignored it — compiling the snippet as if `Test` had written it, when `Test` never `use`s the module under test:

```
ok 1 - throws-like resolves the IMPORTED name
    # Got:      Could not find symbol 'Widget::make'
not ok 2 - throws-like resolves the QUALIFIED name
```

Same sub, same scope. A `CALLER::`/`CALLERS::` pseudo-stash now carries its frame's compunit (`STASH_ORIGIN_UNIT_ATTR`) beside the package and control-flow identity it already carried, and `EVAL` uses it as the EVAL unit's parent. A plain `EVAL` is unchanged — the running compunit *is* its caller's lexical scope.

**Fixes `DateTime::Parse` and `Encode`.**

## Fix 2 — an END phaser runs in the compunit that declared it

`EndPhaser` already recorded and restored the declaring **package**, for a reason its own comment states: END bodies run at exit, long after `current_package` returned to `GLOBAL`. The compunit needed the same treatment for the same reason — the strongest rule in `qualified_name_visible_here` is *"a compunit always sees a package it declares itself"*.

Invisible while the module is also visible from the main compunit, so it took an EVAL-loaded module to surface. `Log::Async`'s `END { Log::Async.instance.done ... }`, reached through `use-ok`, died — and **took the rest of the exit sequence with it**, so the script's own END phasers never ran either.

## Fix 3 — a module's own routines are stamped with the module's file

A routine records its declaring file as it is *created*, via `executing_source_file()`, which walks the routine stack for the innermost frame's `def_file`. A module's mainline runs through `run_block` and pushes **no routine frame**, so that walk goes straight past it to whatever routine is still on the stack underneath — a different compunit entirely when the load came from `EVAL "use $module"` inside another module's routine.

`enter_compilation_unit` then anchors `current_unit` from that stamp on every call, so `Terminal::ANSI::OO`'s own EXPORT could not resolve its own class. Measured at the registration site, the two available answers disagree exactly as predicted:

```
REGSUB name=EXPORT
  exec_sf = .../UseOkish.rakumod                 <- stale caller frame
  ?FILE   = .../Terminal/ANSI/OO.rakumod         <- correct
  rstack  = 2   modstack = (.../OO.rakumod, 2)   <- depths match
```

`declaring_source_file()` is the declaration-side twin of `executing_unit_sym_for_module_load()` and uses the same test for the same reason. A routine call since the mainline started means the frame-based answer is right again, which is what keeps the behaviour those five sites' comments were written to protect.

> An earlier attempt anchored `current_unit` at the `apply_module_export` call site instead. That does **not** work — `enter_compilation_unit` resets `current_unit` from the `CompiledFunction` on entry. Verified with the debugger and reverted rather than left as dead code.

## Verification

- **The gate is green again: `285/311`, all three regressions gone.** The only files still failing locally are the six `DBIish` mysql ones, which need a `libmysqlclient` this container has not got (`NativeCall: symbol 'mysql_init' not found ... dlsym failed`) and which CI installs.
- Three focused regression tests under `t/modules/`, **each verified against rakudo**.
- `cargo fmt --all`, `make lint` (all four configurations), `make check-t-layout`, `make test` (3953 files, 42004 tests, PASS), `make roast` (clean but for the three documented container-only files).

## Note on the merged commit

This branch also carries the #7831 supply-ticket fix from **#7834**. Without it `t/supply-serialize-fifo.t` test 3 fails 17–37% of runs and the `t/` suite is fatal in CI, so an unrelated red would be a coin flip on every push here. It no-ops the moment `main` carries it — if this lands first, #7834 becomes already-merged content and I will close it pointing here.

Closes #7836.

---

### DIFF
--- a/news/2026-09/already-resolved-promise-supply-ticket.md
+++ b/news/2026-09/already-resolved-promise-supply-ticket.md
@@ -1,0 +1,53 @@
+# An already-resolved promise now takes a SupplyTicket too
+
+`t/supply-serialize-fifo.t` test 3 — the pin added for #7811 — failed about one
+run in six on `main` (11 of 30 on an idle release build here), which made the
+fatal `t/` suite redden PRs at random. The failure was always the same shape:
+
+```
+# expected: '1 2 3 4 5 6'
+#      got: '2 4 6 1 3 5'
+```
+
+Not noise, and not a regression from #7811: it is the half of that fix that was
+never written.
+
+#7811 orders the reactions of a supply block by having the thread that resolves
+a promise reserve the reaction's place in the block's sequencer (a
+`SupplyTicket`) before the pooled worker that will run it has even been woken.
+But `SharedPromise::on_resolve_in_supply_group` only did that on the path where
+the promise was still `Planned`: the ticket is taken in `dispatch_waiters`, and
+a promise that is **already** resolved when the nested `whenever` registers
+never reaches `dispatch_waiters`. Its `group` argument was simply dropped, and
+the reaction ran inline on the registering thread — ahead of every reaction
+already ticketed and queued for a worker. The two paths were therefore ordered
+against each other by luck, which is exactly what the test alternates between
+(even values keep the promise before the nested `whenever` sees it, odd values
+after), and why the wrong answer came out cleanly separated rather than
+interleaved.
+
+The already-resolved branch now reserves its ticket too, on the registering
+thread — the ordering point, since that thread is running the enclosing supply
+block's reactions in order — and hands the body to a pooled worker exactly as
+`dispatch_waiters` does. Both paths enter the block through the one sequencer,
+in the order the reactions were created.
+
+Redeeming the ticket inline instead, and keeping the body on the caller's
+thread, is not an option: that thread is typically running the enclosing supply
+block's own reaction and so already holds the group lock, so parking it behind
+an earlier ticket — whose worker is itself blocked on that lock — deadlocks.
+This mirrors the reason the *pending* path hands off to a worker rather than
+running user code on the resolver, which is what deadlocked `Cro::HTTP`'s
+middleware pipeline when it was tried there.
+
+Consequence for the API contract: `on_resolve_in_supply_group` now returns
+`false` for an already-resolved promise when a group is given, because the
+reaction genuinely has not run yet. Plain `on_resolve` (no group) keeps its
+synchronous fast path and its `true`.
+
+40 consecutive runs of the file are green after the fix. `t/supply-serialize-fifo.t`
+grows a fourth test that keeps *every* promise before its nested `whenever`
+registers, so the already-resolved path is exercised on its own rather than only
+in alternation with the ticketed one.
+
+Closes #7831.
--- a/news/2026-09/eval-and-end-run-in-their-declaring-compunit.md
+++ b/news/2026-09/eval-and-end-run-in-their-declaring-compunit.md
@@ -1,4 +1,4 @@
-# A context EVAL, and an END phaser, run in the compunit they belong to
+# Code runs in the compunit it was written in: EVAL context, END phasers, module routines
 
 `main` went red on the `test` job's bundled-library gate with three regressions
 that all reported the same kind of failure — a package-qualified name that would
@@ -52,8 +52,8 @@ A `CALLER::`/`CALLERS::` pseudo-stash now carries its frame's compunit
 already carried, and `EVAL` uses it as the EVAL unit's parent. That fixes
 `DateTime::Parse` and `Encode`.
 
-`on_resolve`-style contract note: nothing else changes for a plain `EVAL`, which
-keeps inheriting the running compunit — that IS its caller's lexical scope.
+Nothing changes for a plain `EVAL`, which keeps inheriting the running
+compunit — that IS its caller's lexical scope.
 
 ## An END phaser runs in the compunit that declared it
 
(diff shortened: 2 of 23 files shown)


tokuhirom:
## Closing: `main` already carries all three of these fixes

While this PR's CI was running, #7842 and #7844 landed the **same three fixes** for #7837, which is the same bug this branch was opened for. Checked against `origin/main` at `484788a`:

| Fix here | On `main` |
|---|---|
| EVAL `context` → the context's compunit (`STASH_ORIGIN_UNIT_ATTR`) | ✅ `282bd7c` |
| `EndPhaser` records and restores the declaring compunit | ✅ `282bd7c` (`EndPhaser::unit`) |
| Module-load correction for the declaring source file | ✅ `282bd7c` + `d50f510`, as `executing_source_file_for_module_load()` |

`d50f510` even refines the third one the same way I would have (read `?FILE` rather than the unit stack), and `t/modules/eval-context-caller-compunit.t` covers the same ground as the tests here. The overlap is complete, which is also why this PR went `dirty`.

Forcing a duplicate through would mean resolving conflicts in six files to land code that is already there, so I am closing it rather than reworking it. Nothing here is lost: the behaviour and its tests are on `main`.

`1aab678` also installs `libmysqlclient` from the SessionStart hook, which fixes the six `DBIish` mysql failures this container was showing — a nice bonus for future agent sessions.

**The #7831 supply-ticket fix this branch carried is still needed** — `main` does not have it. It is back to standing on its own in **#7834**, rebased onto `484788a` and re-verified there.

---

PR #7846, opened 2026-09-10
fix(t): move the last stray top-level test into its category

tokuhirom:
## Problem

`main` is red on the `test` job, at its **`t/ layout is valid`** step — so every PR inherits it:

```
t/any-cool-method-not-found.t sits at t/ top level; every test belongs in a category directory
    scripts/migrate-t-layout.py names the category for a file.
```

Everything before that step passes, `prove t/` included; the job dies at step 15 and skips roast and the bundled-library gate entirely.

## Why it happened

Two PRs crossed, and neither was wrong about its own diff:

- `a8fa6878` ("fix: an undefined Any does not answer Cool's methods") added `t/any-cool-method-not-found.t` at `t/` top level. **That was correct when written** — `scripts/check-t-layout.sh` still had `MIGRATED=0`, so the top-level rule was off.
- #7833 then landed the migration itself, moving 3949 files into categories and flipping `MIGRATED=1`.

The new file simply landed on the other side of the flip, and nothing in either PR's own CI run could see the combination.

## Fix

A pure `git mv` into the category `scripts/migrate-t-layout.py` names for it:

```
t/any-cool-method-not-found.t → t/oo/method/any-cool-method-not-found.t
```

That matches the rule `docs/t-directory-layout.md` states — place by what the test would catch if it broke. The file is about method **dispatch** on an undefined invocant (`Any` not answering `Cool`'s methods, while declared accessors and real built-ins still resolve), not about the `Any`/`Cool` types themselves.

**No test content and no interpreter code changes.**

## Verification

```
$ make check-t-layout
t/ layout OK (3952 tests, all in category directories)
t/ layout matches the rules (3952 files)
```

Both checkers agree — the shell rules and `migrate-t-layout.py --check`. The test itself still passes, 26/26.

Found while driving #7834, which this unblocks along with every other open PR.

---


tokuhirom:
## Closing: #7845 landed the same move first

`main` at `df5c51c0` (#7845) already has the file where this PR puts it — the identical destination, independently chosen:

```
$ git ls-tree -r origin/main --name-only | grep any-cool-method-not-found
t/oo/method/any-cool-method-not-found.t
```

No top-level `.t` remains, and `make check-t-layout` passes on `main` as it stands. Nothing here to add, so closing rather than landing a no-op rename.

#7834, which this was opened to unblock, is rebased onto `df5c51c0` and green locally: `make check-t-layout` OK (both checkers), `prove -r t/` **3952 files / 42022 tests PASS**, and the pinned `t/concurrency/supply/supply-serialize-fifo.t` 20/20 clean.

---

Issue #7579, opened 2026-09-08
Tracking: the late-August call-path slowdown


tokuhirom:
**Tracking issue.** The work lives in the sub-issues below; this page keeps the history, the standing methodology, and the one warning that is easy to get wrong.

Audited and restructured 2026-09-09. The original body was a single list of seven items accumulated over a month; most had been closed in place, several of its "still open" descriptions no longer matched the code, and one was measurably wrong. It could not be worked in parallel — the claim protocol is per-issue, so two agents could not take two items — and it could not carry a `tier:*`. The full item-by-item audit is in [the audit comment](#issuecomment-5600935467) and below.

## Open work

| | |
| --- | --- |
| #7737 | The JIT-generated body is the second-largest symbol on `bench-fib` (13.8%) and nothing in the interpreter explains it. **Largest open item.** |
| #7738 | positional-light call: 35% of `bench-fib` is per-call frame bookkeeping spread across a dozen sites, not one hotspot. (Absorbs old items 1 and 5.) |
| #7739 | `CompiledCode::const_sym` is a per-call `OnceLock` side-table probe, 3.5% of `bench-fib`. (New; found by the audit.) |

Related but tracked separately: #7736 (`Symbol::intern` on the `Test` assertion path).

## The original finding

Between 2026-08-19 and 2026-08-31 a broad set of call-path-shaped benchmarks got slower while an unrelated set got faster. Daily medians of the bench-CI series (`git show origin/bench-data:bench-history.tsv`), 2026-08-17..19 vs 2026-08-31..09-02:

| regressed | | improved | |
| --- | --- | --- | --- |
| `hash-access+jit` | +37% | `bench-array+jit` | −40% |
| `bench-grammar-parse+jit` | +36% | `bench-ctor+jit` | −33% |
| `bench-fib+jit` | +34% | `bench-yaml-parse+jit` | −21% |
| `bench-tak+jit` | +23% | `method-call+jit` | −10% |
| `bench-hash+jit` | +19% | `bench-class+jit` | −9% |

`bench-startup` is flat over the window (+0.4 ms), so this is not startup cost.

The largest single step was found by bisection and fixed (`news/2026-09/adr0037-routine-frame-push-intern-cost.md`): #6720 added a `RoutineFrame` push to the light call paths whose five per-call `Symbol::intern`s cost 26% of `bench-fib`.

## ⚠️ Do NOT bisect the remainder — read this before trying

[...]

PR #7889, opened 2026-09-11
perf: resolve const_sym from an eager table instead of two OnceLocks

### LINKED ISSUE #7739: CompiledCode::const_sym is a per-call OnceLock side-table probe, 3.5% of bench-fib
New finding from the 2026-09-09 audit of #7579; not in that ticket's original list.

## Benchmark

```
cargo build --profile profiling
./target/profiling/mutsu benchmarks/bench-fib.raku
```

## Measurement

`bench-fib` under callgrind, at the merge of #7731. Total 1,288,756,087 Ir.

| symbol | Ir | share |
| --- | ---: | ---: |
| `src/opcode.rs:CompiledCode::const_sym` | 35,594,300 | 2.76% |
| `std/sync/once_lock.rs:CompiledCode::const_sym` | 10,169,836 | 0.79% |
| **total** | **45,764,136** | **3.55%** |

For scale, that is half of what `exec_call_func_op_inner` now costs (7.35%) after ADR-0066 removed its two hash probes.

## Where the time goes

`const_sym` is the fix that ADR-0037 Slice 1 introduced to stop the call path re-interning routine names per call — `Symbol::intern` is a thread-local string-keyed hash, and doing five of them per call cost ~26% of `bench-fib`. It replaced that with a lazily-built per-chunk side table:

```rust
let slots = self.const_syms.get_or_init(|| …);   // OnceLock
match slots.get(idx as usize) {
    Some(slot) => *slot.get_or_init(|| resolve(idx)),   // OnceLock per slot
    None => resolve(idx as usize),
}
```

That is strictly cheaper than an intern, and it worked. But the replacement is itself **two `OnceLock` acquires plus a bounds-checked index on every call**, and now that the interns either side of it are gone it is visible on its own.

This is the same shape as the finding in `news/2026-09/incdec-index-name-symbol.md` (#7731), one level further in: there the fix was to stop re-deriving a `Symbol` per execution when the call site could hold it. Here the `Symbol` is already memoized — the remaining cost is *reaching* the memo. ADR-0066's "What was actually built" section records the identical lesson from the other direction: a per-callsite inline cache reached through `OnceLock` → side table → index array → slot removed **no** retired instructions, because five dependent loads cost what two hash probes cost. Whatever replaces this must be **one load**, not a shorter pointer chase.

## Possible directions (unmeasured)

- Resolve the symbol **once per call site at finalize time** and store it in the opcode payload, so dispatch reads a field instead of probing a side table. `OpCode` has a 48-byte cap pinned by `opcode_size_guard`; `CallFunc`'s payload is `{u32, u32, Option<u32>}`, so there may be room — check before assuming.
- Or make `const_syms` a plain `Box<[Symbol]>` resolved eagerly at finalize, turning the two `OnceLock` acquires into one indexed load. Costs interning every string constant of every chunk up front, including ones never called — measure against `bench-startup`, which #7579 records as flat over the regression window and would be the thing this regresses.

Neither is obviously right; that is why this is a `todo:perf` and not a ticket.

## Why it is not a quick fix

The eager variant trades startup for steady-state and needs `bench-startup` as a control. The opcode-payload variant touches the compiler, `OpCode`'s size guard, and the JIT's `call_func` shim (which also calls `const_sym`). Both need the retired-instruction A/B discipline in #7579's method notes, and a same-binary A/B if cycles are quoted at all — this box's cross-build cycle noise is larger than the effect.

Per `docs/issue-workflow.md`, a `todo:perf` implementation agent runs **solo**.

tokuhirom:
## What

`CompiledCode::const_sym` reached its memoized `Symbol` through an outer
`OnceLock` holding a side table and a second `OnceLock` per slot. ADR-0037
Slice 1 built that to keep `Symbol::intern` off the call path, and it worked —
but once ADR-0037 and ADR-0066 had cleared the interns and hash probes either
side of it, the cost of *reaching* the memo was what was left.

A chunk is finalized before it executes, so the table never needed to be lazy.
`compute_needs_env_sync` — which already walks the constant pool to decide
`reads_topic` — now interns every string constant in one pass into a plain
`Vec<Option<Symbol>>`, and `const_sym` is an indexed load out of it. That is
the bar ADR-0066 records: a shorter pointer chase retires no fewer
instructions, only one load does.

`#[inline]` is part of the fix, not a flourish — see the measurement below.
The `intern_const_sym` fallback is `#[cold] #[inline(never)]` so inlining
copies only the load into each dispatch site.

`None` marks a non-string constant; an empty table marks a chunk that never
finalized (the hand-built `CompiledCode::new()` chunks in `runtime/`, whose
constant pools are empty or trivial). Both fall through to interning directly —
what `const_sym` did before ADR-0037 — so nothing regresses below the pre-memo
behaviour.

`add_constant` extends the table past finalize. Sizing it once at finalize
would otherwise have left a gap the lazy predecessor did not have: that one
sized its table on first *use*, i.e. at runtime, so it covered a chunk patched
with a new constant afterwards, where a finalize-sized table would let that slot
fall off the end and re-intern on every access.

## Measurement

callgrind, `--profile profiling`, base `69bedb4`. Retired instructions are
deterministic, so these are exact rather than sampled, and the A/B is
same-benchmark across three separately-built binaries.

| | `bench-fib` total | vs base | `const_sym` | `bench-startup` |
| --- | ---: | ---: | ---: | ---: |
| base (`69bedb4`) | 1,291,994,518 | — | 55.9M (4.33%) | 8,737,119 |
| eager table | 1,264,035,861 | −2.16% | 28.0M (2.21%) | 8,744,395 |
| + `#[inline]` | **1,248,131,156** | **−3.40%** | **10.5M (0.84%)** | **8,737,168** |

Two notes on the numbers:

- **The ticket recorded `const_sym` at 3.55%; it was 4.33%.** The
  `slice/index.rs` and `core/sync/atomic.rs` lines attributed to `const_sym`
  were not counted in the ticket's table.
- **The startup trade-off the ticket worried about did not materialise.** The
  eager variant interns every string constant of every chunk up front,
  including ones never called, and #7579 records `bench-startup` as flat over
  the regression window. It is flat here too: 49 instructions above base,
  0.0006%. The middle row shows why — the eager pass costs ~7.3k instructions
  at startup, and inlining `const_sym` gives them back.

The middle row is also why the `#[inline]` commit exists. With the table eager
but the function out of line, `bench-fib` paid a call, a prologue and an
epilogue 2.54M times: 11 retired instructions per call for a read that is
about six.

The later `add_constant` commit does not move these numbers: it adds a branch to
constant-pool insertion (compile time) and none to `const_sym`, whose read stays
the single indexed load the table records. `Vec` and `Box<[T]>` both index
through an inline pointer+len.

I did not take the ticket's other suggested direction (resolving into the
`OpCode` payload at finalize). It touches the compiler, `OpCode`'s 48-byte size
guard and the JIT `call_func` shim, and this gets the read to one load without
any of that.

## Left on the table

`const_sym` is now ~0.84% of `bench-fib`, and about four fifths of that is
call-site redundancy rather than cost per read: `exec_call_func_op_inner` calls
`const_sym(name_idx)` four times per dispatch where one hoisted local would do.
That is worth roughly 0.7%, but it is a call-site cleanup rather than a
`const_sym` fix, so I left it for whoever next picks up #7579 rather than widen
this PR. It is written up in the news entry.

## Testing

Run from scratch on the final head:

- `make lint` — green in all four configurations.
- `make test` — PASS, 3963 files / 42,149 tests.
- `make roast` — 1436 files / 218,939 tests. Failures are the three documented
  container artefacts, matching `docs/agent-environments.md` down to the counts:
  `6.c/S32-io/file-tests.t` (Failed 4) and `S16-filehandles/filetest.t`
  (Failed 25), both because this container runs as `uid 0` and root bypasses the
  permission bits these tests `chmod`; and `S32-io/IO-Socket-Async.t` (exit 124,
  planned 40 ran 17) on the sandboxed network. `id -u` confirms 0.

Unit tests pin every state a `const_sym` caller can land in: an unfinalized
chunk with an empty table, an index inside the table, a constant appended after
finalize, and a non-string slot staying `None` — all agreeing with interning
the constant directly.

CI's first run went red on `roast/S17-procasync/kill.t` test 7 while `test` and
`gc-stress` passed the same file on the same commit. That is a pre-existing
load-dependent flake, not this diff: it passes **32/32** in isolation under the
jit-stress environment with four CPU burners, and fails only as one of 1436
files under `prove -r -j4`. It is now filed with its evidence as #7892, and the
re-run on the current head is green including `jit-stress`.

Closes #7739

https://claude.ai/code/session_01N9eneyfyR623qcZ5JuhUeY

### DIFF
--- a/news/2026-09/const-sym-eager-table.md
+++ b/news/2026-09/const-sym-eager-table.md
@@ -1,0 +1,71 @@
+# `const_sym` reads an eager table instead of probing two OnceLocks
+
+`CompiledCode::const_sym` answers "what `Symbol` is the string constant at this
+pool index" — the question every `CallFunc`, `CallMethod` and variable-access
+opcode asks about its name. ADR-0037 Slice 1 made it a memo because the call
+path had been re-interning routine names per call, and `Symbol::intern` is a
+thread-local string-keyed hash: five of them per call had been ~26% of
+`bench-fib`. The memo worked.
+
+But the memo was reached lazily, through an outer `OnceLock` holding the side
+table and a second `OnceLock` per slot holding the symbol. Once ADR-0037 and
+ADR-0066 had cleared the interns and hash probes either side of it, the cost of
+*reaching* the memo was what was left, and it was not small: 55.9M of
+`bench-fib`'s 1,292.0M retired instructions, 4.33%.
+
+That is the same lesson ADR-0066 records from the other direction — a
+per-callsite inline cache reached through `OnceLock` → side table → index array
+→ slot retired no fewer instructions than the two hash probes it replaced,
+because five dependent loads cost what two hash probes cost. A shorter pointer
+chase is not the fix; one load is.
+
+## What changed
+
+A chunk is finalized before it executes, so the table never needed to be lazy.
+`compute_needs_env_sync` — which already walks the constant pool to decide
+`reads_topic` — now interns every string constant in one pass and stores the
+result in a plain `Box<[Option<Symbol>]>`. `const_sym` is an indexed load out of
+that, marked `#[inline]` so each dispatch site gets the load itself rather than
+a call; the intern fallback is `#[cold] #[inline(never)]` so the hash probe
+stays behind a call.
+
+`None` marks a non-string constant and an empty table marks a chunk that never
+finalized — the hand-built `CompiledCode::new()` chunks in `runtime/`, whose
+constant pools are empty or trivial. Both fall through to interning directly,
+which is exactly what `const_sym` did before ADR-0037, so nothing regresses to
+worse than the pre-memo behaviour.
+
+## Measurement
+
+callgrind, `--profile profiling`, at the merge of `69bedb4`. Retired
+instructions are deterministic, so these are exact rather than sampled.
+
+| | `bench-fib` total | vs base | `const_sym` | `bench-startup` |
+| --- | ---: | ---: | ---: | ---: |
+| base (`69bedb4`) | 1,291,994,518 | — | 55.9M (4.33%) | 8,737,119 |
+| eager table | 1,264,035,861 | −2.16% | 28.0M (2.21%) | 8,744,395 |
+| + `#[inline]` | **1,248,131,156** | **−3.40%** | **10.5M (0.84%)** | **8,737,168** |
+
+The eager variant's risk was startup: it interns every string constant of every
+chunk up front, including ones never called, and #7579 records `bench-startup`
+as flat over the regression window. It is flat here too — 49 instructions above
+base, 0.0006%. The intermediate row shows where that came from: the eager pass
+itself costs ~7.3k instructions at startup, and inlining `const_sym` gives them
+back.
+
+The middle row is also why `#[inline]` is part of the fix rather than a
+flourish. With the table eager but the function out of line, `bench-fib` paid a
+call, a prologue and an epilogue 2.54M times — 11 retired instructions per call
+for a read that is about six.
+
+## Left on the table
+
+`const_sym` is now ~0.84% of `bench-fib`, and roughly four fifths of that is
+redundancy rather than cost per read: `exec_call_func_op_inner` calls
+`const_sym(name_idx)` four times per dispatch (the `nqp::` flag check, the
+unit-scoped-routine check, the light-cache key, and the argument to
+`call_compiled_function_positional_light_at`) where one hoisted local would do.
+That is worth about 0.7% and is a call-site cleanup, not a `const_sym` fix, so
+it is left for whoever next picks up #7579.
+
+Closes #7739.
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -2809,6 +2809,44 @@ mod const_pool_dedup {
     }
 
     #[test]
+    fn const_sym_agrees_with_intern_whether_or_not_the_chunk_finalized() {
+        let mut code = CompiledCode::new();
+        let a = code.add_constant(Value::str("fib".to_string()));
+        let b = code.add_constant(Value::str("$n".to_string()));
+
+        // Before finalize the table is empty, so `const_sym` falls back to
+        // interning directly -- a hand-built chunk must still resolve.
+        assert!(code.const_syms.is_empty());
+        assert_eq!(code.const_sym(a), Symbol::intern("fib"));
+        assert_eq!(code.const_sym(b), Symbol::intern("$n"));
+
+        // After finalize the same answers come out of the eager table.
+        code.compute_needs_env_sync();
+        assert_eq!(code.const_syms.len(), code.constants.len());
+        assert_eq!(code.const_sym(a), Symbol::intern("fib"));
+        assert_eq!(code.const_sym(b), Symbol::intern("$n"));
+
+        // A constant appended after finalize is past the table's end and takes
+        // the fallback again.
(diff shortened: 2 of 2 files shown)


tokuhirom:
## `jit-stress` failure on 8f2820a is not this PR's

`jit-stress` failed on one assertion: `roast/S17-procasync/kill.t` test 7, the
`doesn't-hang` case that spawns four `Proc::Async` subprocesses, kills each with
a different signal spelling, and requires `All done!` within a 10-second budget.
The `Result: FAIL` line is that single test; every other entry in the summary is
a `TODO passed`. (The SIGSEGV in the crash report is the deliberate
`strdup(0)` NativeCall probe — the job's own allowlist accepts it and says so.)

Evidence that it is non-deterministic rather than a regression, per the triage
protocol in CLAUDE.md:

- **`test` and `gc-stress` both passed on this exact commit**, and both run the
  full roast suite including this file. Only `jit-stress` went red.
- **`make roast` passed it locally** on the release build before I opened the PR.
- **`scripts/flake-repro.sh -n 12 -l 4 -e MUTSU_JIT=on -e MUTSU_JIT_THRESHOLD=2
  roast/S17-procasync/kill.t` → `pass=12 fail=0`** on this branch's release
  binary, i.e. twelve runs under the jit-stress environment with four CPU
  burners emulating the runner's contention.

The shape also rules the diff out mechanically: this PR replaces a lazily-built
`Symbol` side table with an eagerly-built one. It changes neither signal
delivery, subprocess scheduling, nor anything else `kill.t` measures.

I have not added it to `flaky-tests.txt` — one CI failure plus twelve local
passes is not the evidence standard `docs/flaky-test-policy.md` sets for a
quarantine entry, and the quarantine list is not mine to grow on this PR.

Rather than spend the re-run on it, I am pushing a follow-up commit (which
re-runs CI anyway). Reviewing my own diff for this turned up a real gap worth
closing: `const_syms` was sized at finalize and never extended, so a chunk
patched with a new constant afterwards had that slot fall off the end of the
table and re-intern on every access. The lazy predecessor sized its table at
first use — i.e. at runtime, after any such patch — so it covered those slots.
`add_constant` now keeps the table index-aligned past finalize, restoring that.

---

tokuhirom:
Update: `kill.t` test 7 then failed in a **local** full `make roast` too, on a
run where the previous one had passed — so the non-determinism is now
reproduced on two machines rather than inferred from one CI job. Filed as
#7892, with the full evidence table and the two readings (contention noise vs.
a real `Proc::Async` reaping defect) that still need telling apart before it
could be quarantined.

The discriminator turns out not to be the JIT or CPU contention: in isolation
the file passes **32/32** under `MUTSU_JIT=on MUTSU_JIT_THRESHOLD=2` with four
CPU burners (12 runs before the last commit, 20 after it). It only fails as one
of 1436 files under `prove -r -j4`. Two of four full-suite runs failed; every
isolated run passed.

Pushed `bb61105` with the `add_constant` alignment fix described above. Gates on
that head, all re-run from scratch:

- `make lint` — green in all four configurations.
- `make test` — PASS, 3963 files / 42,149 tests.
- `make roast` — the three documented `uid 0` / sandboxed-network container
  artefacts, plus this one `kill.t` flake.

The `#7739` measurement is unaffected by that commit: it adds a branch to
`add_constant` (compile time) and none to `const_sym`, whose read stays the
single indexed load the benchmark table records.

---

Issue #7892, opened 2026-09-11
roast/S17-procasync/kill.t test 7 fails intermittently under full-suite load, and is not quarantined


tokuhirom:
Found while ruling out a CI failure on #7889 (which turned out to be unrelated to that PR's diff).

`roast/S17-procasync/kill.t` test 7 fails intermittently. It is **not** in
`flaky-tests.txt`, so there is no retry and a single unlucky roll blocks a merge
outright. Either it should be fixed or it needs an entry backed by the evidence
`docs/flaky-test-policy.md` requires — this ticket is to decide which.

## The test

Test 7 of 8 is the `doesn't-hang` case at `roast/S17-procasync/kill.t:47`,
`'.kill kills when multi-procs kill in multi-promises'`:

```raku
doesn't-hang ｢
        await ^4 .map: -> $n {
            start {
                with Proc::Async.new: $*EXECUTABLE, "-e", "sleep" -> $p {
                    start {
                        await $p.ready;
                        $n == 3 ?? $p.kill !! $p.kill: (SIGTERM, 'TERM', SIGTERM.value)[$n]
                    }
                    await $p.start
                }
            }
        }
        print 'All done!'
    ｣,   :out('All done!'), :err(''), :10wait,
```

Four `Proc::Async` subprocesses, each spawning a whole second mutsu, killed
through four different signal spellings, with a hard 10-second budget.

## Evidence gathered 2026-09-11

Both failures are the same subtest; everything else in the file passes.

| Run | Result |
| --- | --- |
| CI `jit-stress`, run 34548282326 (head 8f2820a) | **FAIL** test 7 |
| CI `test`, same commit | pass |
| CI `gc-stress`, same commit | pass |
| Local `make roast` (release), run 1 | pass |
| Local `make roast` (release), run 2 | **FAIL** test 7 |
| `flake-repro.sh -n 12 -l 4 -e MUTSU_JIT=on -e MUTSU_JIT_THRESHOLD=2` | 12/12 pass |
| `flake-repro.sh -n 20 -l 4 -e MUTSU_JIT=on -e MUTSU_JIT_THRESHOLD=2` | 20/20 pass |

So: 32/32 in isolation even with four CPU burners, but two failures out of four
full-suite runs. The discriminator is not the JIT and not CPU burners — it is
running as one of 1436 files under `prove -r -j4`, where hundreds of other
subprocesses compete. That matches the "`S17-*` concurrency tests may fail under
[...]

PR #7900, opened 2026-09-11
chore(vendor): re-vendor roast at 85a87909 (2026-09-07)

tokuhirom:
`roast/` had been pinned at `b2cbe8a4` (2026-06-12) for three months. It is now re-vendored at upstream master `85a87909` (2026-09-07) via `scripts/update-vendor.sh roast`, which also rewrote the `roast` row of `vendor.lock`. Seventeen commits, 20 touched test files (19 of them whitelisted) and one new file.

## Why now: upstream de-flaked three quarantined tests

Upstream raised the `.roll` sample size in the Set/Bag/Mix tests from 100 to 100000 and scaled the assertion bounds with it. The Binomial(100, 1/3) assertions that made `S02-types/{bag,baghash,mixhash}.t` **statistically flaky by design** now sit roughly 100 standard deviations from the mean and cannot fail by chance.

All three are removed from `flaky-tests.txt` (3/3 green locally, ~1-2s each on release), leaving `integration/advent2014-day05.t` as the only quarantined test. Their long prose in `CLAUDE.md` moves from the "known flaky" list to the "de-flaked" list, so a failure there is a real bug again.

Other de-flaking in the same batch, all of which mutsu passes unchanged: `S17-channel/basic.t` awaits the Supply's `done` before reading what the tap saw, `S17-promise/then.t` awaits the independent `.then` it had been reading without awaiting, `S29-os/system.t` runs a silent child instead of `$*EXECUTABLE -v` in its 200-iteration loops, and `S32-io/IO-Socket-Async.t` holds a real port for its EADDRINUSE probe instead of guessing one.

A second group (`my-6c.t`, `my-6e.t`, `arrays.t`, `S12-enums/basic.t`, `short-circuit.t`, `subst.t`) was rewritten so the runtime rather than a constant-folding compiler does the rejecting. mutsu passes all of those unchanged too.

## Whitelist follow-up — a net gain, not just a cost

Per `docs/vendoring.md`, every whitelisted file upstream touched was re-run. Thirteen of the nineteen still passed unchanged. Six files did not (five whitelisted, plus the brand-new `sprintf-a.t`), and each became an issue carrying its failing subtests, a repro, and the measured rakudo comparison.

**Two of the six were implemented on `main` within hours of being filed, so this PR now merges `main` and re-measures against them:**

| Issue | Landed as | File | Before | After |
|---|---|---|---|---|
| #7902 | `8fdafe81` C99 hexadecimal float literals | `S02-literals/numeric.t` | parse error, 0/89 | **PASS 89/89** |
| #7903 | `4f970ce7` `sprintf` `%a` / `%A` | `S32-str/sprintf-a.t` (new) | 0/586 | **PASS 586/586** |

Both are on the whitelist. The re-vendor's net effect is therefore **1436 → 1433 files but +586 subtests**, and mutsu is now *ahead of* Rakudo v2026.07 on both hexfloat files — it still fails them.

The four that stay off the whitelist are recorded in `TODO_roast/BLOCKERS.md` under "Files dewhitelisted by the 2026-09-11 roast re-vendor", each cross-referenced to its issue:

| File | mutsu | raku v2026.07 | Issue |
|---|---|---|---|
| `S05-capture/caps.t` | 55/56 (notok 47) | 43/56 | #7904 |
| `S05-metasyntax/charset.t` | 89/90 (notok 54) | 57/90 | #7905 |
| `S05-metasyntax/regex.t` | 66/68 (notok 59, 67) | 58/68 | #7906 |
| `S05-modifier/ignorecase.t` | 110/115 (notok 29/31/35/39/40) | 103/115 | #7907 |

**These are not mutsu regressions.** They are spec tests written against still-open rakudo bugs (rakudo#4105 capture state after backtracking, #5588 block-quantifier limits under backtracking, #2962 character-class fold/mark semantics, #4512 lookaheads over compound character classes), so the local oracle fails them too — by a wider margin than mutsu in every one. Each misses by one or two subtests, so whitelist return is close.

`S32-io/IO-Socket-Async.t` stays whitelisted: it times out at "planned 40 ran 17" in the remote agent container, which is the documented network-sandbox failure (`docs/agent-environments.md`), reached at the IPv6 section long before the rewritten EADDRINUSE probe.

## Verification (re-run on the merged tree)

- `make test` **PASS** — 3988 files, 42456 tests.
- `make roast` — `Files=1433`, and the only failures are the three container-only ones documented in `docs/agent-environments.md`: `6.c/S32-io/file-tests.t` tests 6-8/10 and `S16-filehandles/filetest.t` tests 57-64/69-72/77-80/101,103,105,107,109,111,117,121,125 (both from running as `uid 0`, confirmed with `id -u`), and `S32-io/IO-Socket-Async.t` exit 124 (network sandbox). Subtest sets match the documented ones exactly.
- This branch changes no Rust source of its own, so clippy/rustdoc output matches `main`.

## Note on the CI history of this PR

The earlier red `jit-stress` on `4db31e4` was `roast/S17-procasync/kill.t` test 7, which was failing intermittently on `main` too (runs 34556487764 `jit-stress`, 34554535164 `gc-stress`). It has since been root-caused — it is not a timeout but a lost-update race in mutable native-method dispatch, filed as **#7923** with a repro; **#7892** tracks the test. Nothing was quarantined.

https://claude.ai/code/session_01Pmx4WKX5S4aRuLvTyHSMEC


tokuhirom:
CI status on `4db31e4`: `test`, `gc-stress`, `lint-configs`, `wasm-e2e`, `changes`, `label` all **green**. `jit-stress` failed, and the failure is **not this PR's**.

The entire roast run in that job is clean except one subtest:

```
roast/S17-procasync/kill.t   (Wstat: 256 (exited 1) Tests: 8 Failed: 1)
  Failed test:  7
Files=1431, Tests=218654, 189 wallclock secs
```

(The job's other `##[error]` line, "1 mutsu process(es) died of a fatal signal", resolves in the same log to `-> known deliberate crash (argv matches the allowlist), not treated as a failure` — that step passed.)

Why this is not this PR's failure:

- **The diff contains no Rust.** It is `roast/` + `vendor.lock` + `roast-whitelist.txt` + `flaky-tests.txt` + three docs files. `lint-configs` passing on an unchanged `src/` is consistent with that.
- **`S17-procasync/kill.t` is not one of the 20 files the re-vendor touched.** It is byte-identical to what is on `main`.
- **This exact failure is already filed as #7892**, opened two hours before this PR, with the same job-spread signature: `jit-stress` fails test 7 while `test` and `gc-stress` pass on the identical commit. That issue records 32/32 passes in isolation (including under `MUTSU_JIT=on` with four CPU burners) against two failures in four full-suite runs, so the discriminator is full-suite load, not this diff.

No fix exists to port. #7892 deliberately stops short of quarantining the file: `docs/flaky-test-policy.md` requires the non-determinism to be understood and shown not to be a mutsu defect, and the open question there is exactly whether the 10-second `doesn't-hang` budget is being blown by contention or by a real `Proc::Async` reaping defect. Adding a `flaky-tests.txt` entry here to turn this PR green would be quarantining a test to get green, which the project rules forbid and which would bury that question — so I have not.

I have re-run the failed job once (the single re-run these rules allow). The new data point is recorded on #7892. If it fails again on the same subtest, that is evidence for #7892, not a reason to weaken this PR.


---

tokuhirom:
Update: the re-run failed on the identical subtest (`kill.t` test 7, everything else in the 1431-file roast run clean), and chasing it produced the decisive evidence — **`main` is red with the same failure**, so this is the "the check is red on the base branch too" case rather than anything in this diff:

| Run | head | Job | Result |
| --- | --- | --- | --- |
| [34556487764](https://github.com/tokuhirom/mutsu/actions/runs/34556487764) | `a822d9d` (main, merge of #7897) | `jit-stress` | **failure**, `kill.t` test 7, `Files=1436` |
| [34554535164](https://github.com/tokuhirom/mutsu/actions/runs/34554535164) | `51308ec` (main) | `gc-stress` | **failure** at the same roast step |

`Files=1436` on `a822d9d` is the old roast pin, so `main` fails this with the *pre-re-vendor* test files. Two consequences:

- It is not this PR's, by the base-branch criterion directly.
- It also kills the one way this PR could plausibly have contributed. Upstream raised the `.roll` sample in three files from 100 to 100000, so I measured whether that starved the test's 10-second budget: `bag.t` 0.20s → 3.66s, `baghash.t` 0.21s → 1.55s, `mixhash.t` 0.20s → 1.59s (release, `MUTSU_JIT=on MUTSU_JIT_THRESHOLD=2`) — about 6s added to a ~575s-CPU run, ~1%, and `main` fails without any of it. Hypothesis discarded.

My one permitted re-run is spent and there is no fix to port: #7892 is the tracking issue and its mechanism is still unknown, so a `flaky-tests.txt` entry would be quarantining a possibly-real `Proc::Async` reaping defect, which `docs/flaky-test-policy.md` forbids. I have posted the new evidence there instead — including that this is not JIT-specific (it failed in `gc-stress` on one main commit and `jit-stress` on another), which redirects that investigation.

This PR is green on `test`, `gc-stress`, `lint-configs`, `wasm-e2e`, `changes` and `label`. It stays watched until it is green or the blocker is resolved.


---

Issue #7879, opened 2026-09-11
A sigilless parameter of an ANONYMOUS sub is not captured by a nested Proxy STORE closure


tokuhirom:
## Symptom

A sigilless parameter (`\obj`) of an **anonymous** `sub ... is rw { ... }` that is
bound to a scalar is not visible inside a `Proxy` `STORE` closure created in that
sub's body: the name falls back to bareword-as-its-own-name, so `obj` reads back
as the string `"obj"`.

The same code with a **named** `sub f(\obj, \key) is rw { ... }` works, and so does
the `FETCH` side of the very same Proxy — only the `STORE` closure of an anonymous
sub loses the capture.

## Minimal repro

```raku
my $out = "";
my $f = sub (\obj, \key) is rw {
    Proxy.new(FETCH => { "f" }, STORE => -> $, $v { $out = "s:" ~ obj ~ ":" ~ key })
};
($f("OBJ","KEY")) = "V";
say $out;
```

| | output |
|---|---|
| `raku` (2026.07) | `s:OBJ:KEY` |
| `mutsu` (12d29a1 + the QuantHash-subclass branch) | `s:obj:key` |

Both sides agree when the sub is named:

```raku
sub f(\obj, \key) is rw {
    Proxy.new(FETCH => { "f" }, STORE => -> $, $v { $out = "s:" ~ obj ~ ":" ~ key })
}
(f("OBJ","KEY")) = "V";   # mutsu: s:OBJ:KEY  — correct
```

and when the parameters are `$`-sigiled:

```raku
my $f = sub ($obj, $key) is rw {
    Proxy.new(FETCH => { "f" }, STORE => -> $, $v { $out = "s:" ~ $obj ~ ":" ~ $key })
};
# mutsu: s:OBJ:KEY — correct
```

## Where it hurts

Found while making the `AccountableBagHash` distribution's suite pass. The
`nextcallee` base candidate for a container subclass's `AT-KEY`
(`src/runtime/container_element_proxy.rs`) is exactly this shape — a `Proxy`
built inside an anonymous `is rw` sub — and had to use `$`-sigiled parameters to
step around the bug. It should be able to use the sigilless spelling Rakudo's
own `Baggy::AT-KEY` uses.

## Suspected area

The closure free-var / capture analysis for sigilless names
(`crate::env::is_plain_user_lexical` gates several capture paths on a sigil), in
combination with the anonymous-sub compile path and the Proxy `STORE` invocation
[...]

PR #7926, opened 2026-09-11
fix: keep `is raw` on an anonymous sub

tokuhirom:
`is raw` on an **anonymous** sub was parsed and then thrown away:

```raku
my $store = 0;
my $f = sub () is raw {
    Proxy.new(FETCH => { $store }, STORE => -> $, $v { $store = $v })
};
($f()) = 7;
say $store;
```

| | output |
|---|---|
| `raku` 2026.07 | `7` |
| `mutsu` (before) | `Cannot modify an immutable value (sub is not rw)` |

The **named** spelling (`sub f() is raw { ... }`) worked, and so did `is rw` in
either spelling — which is what kept this hidden.

## Root cause

`SubTraits` had already parsed the trait; it had nowhere to go.
`Expr::AnonSub` and `Expr::AnonSubParams` — the two nodes every closure literal
lands on — carried only `is_rw`, so the anon-sub parser wrote
`is_rw: traits.is_rw` and dropped `traits.is_raw` on the floor. The pooled
`Stmt::SubDecl` the compiler then built hardcoded `is_raw: false`, and the
`SubData` the VM made from it therefore never reached the rw-capability oracle
(`is_rw || is_raw || a return-rw in the body`) that
`assign_callable_lvalue_with_values` consults before writing through a routine's
returned container.

Both AST nodes carry `is_raw` now, the parser records it, and the compiler
threads it into the pooled declaration and into `compile_routine_closure_body`'s
rw-tail decision, exactly as it already did for `is_rw`. The bulk of the diff is
the ~50 other construction sites, which are synthesized closures and pass
`false`; the three that destructure-and-rebuild an anonymous sub (the unary-wrap
helper, the invocant-alias binder, the proto-dispatch rewriter) carry the flag
through, as does the lift of one candidate out of an anonymous `multi sub`.

## Why now: the prelude was carrying two workarounds

`runtime/container_element_proxy.rs` is the native `AT-KEY` base candidate that
`nextcallee` hands a container subclass's own override (#7895). It is supposed to
be spelled the way Rakudo spells `Baggy::AT-KEY`, but it had to dodge this bug
*and* #7879 (sigilless captures in a `Proxy` STORE closure, since fixed by
#7911), so it used `$`-sigiled parameters and `is rw`. Both workarounds are gone:

```raku
sub (\obj, \key) is raw {
    Proxy.new(
        FETCH => { obj.__mutsu_container_at_key(key) },
        STORE => -> $, \value { obj.__mutsu_container_assign_key(key, value) }
    )
}
```

A prelude that diverges from the upstream spelling is a small private dialect of
exactly the kind BATTERIES.md §1 warns about — every divergence is a place where
mutsu's behaviour can drift from rakudo's without a test noticing. Keeping this
one honest cost one real compatibility fix.

## Tests

`t/routines/closure/anon-sub-raw-lvalue-return.t` (7 assertions, all verified to
pass under `raku` as well as mutsu): the named/anonymous × `is raw`/`is rw`
matrix, the sigilless-parameter capture through a `Proxy` STORE, and the
negative case that an anonymous sub with **neither** trait still refuses the
write (the fix must not make every anonymous sub rw-capable).

One spelling is deliberately *not* pinned: `sub is raw { ... }` with no
signature. mutsu accepts it, but rakudo parses `is` as the routine's name and
rejects the block, so there is no correct behaviour to assert.

`make test`, `make lint` and `make roast` are all green locally, apart from the
three container artifacts `docs/agent-environments.md` lists
(`6.c/S32-io/file-tests.t` and `S16-filehandles/filetest.t` run as uid 0;
`S32-io/IO-Socket-Async.t` needs network). `AccountableBagHash`'s suite is still
16/16 on the reworded prelude.

https://claude.ai/code/session_01R4XEqYhaYfKvEv2ZeCfGKe

---
_Generated by [Claude Code](https://claude.ai/code/session_01R4XEqYhaYfKvEv2ZeCfGKe)_


tokuhirom:
`gc-stress` went red on `roast/S17-procasync/kill.t` (`Failed 1/8 subtests`,
exit 1). This is not this PR's failure, and there is no fix to port — it is an
unquarantined non-deterministic test.

**Why it is not this PR's.** The diff adds an `is_raw` field to the two
anonymous-sub AST nodes and rewords the `container_element_proxy` prelude.
`kill.t` contains no `is raw`, no anonymous-sub lvalue return, no QuantHash, and
no `AT-KEY`/`nextcallee` — grepping it for any of those is zero hits — so
neither changed path is on its execution path at all. `test`, `lint-configs`
and `wasm-e2e` are all green on this same head.

**Measured, not assumed.** Under the exact `gc-stress` environment
(`MUTSU_GC=on MUTSU_GC_EVERY_CANDIDATE=1024 MUTSU_GC_VERIFY=1
MUTSU_ROAST_TIMEOUT_SCALE=2`) on a release build of this branch, the file failed
**1 of 61 runs** (8 + 12 + 40 in three batches) on an otherwise idle box. A CI
runner under load is exactly where a rate like that turns into an occasional red
job. `make roast` on this branch was green locally before the PR was opened.

**Not the first time.** The same file failed once on #7895's first CI run (in
the `test` job, subtest 7 — `.kill kills when multi-procs kill in multi-promises`)
and passed on the re-run there. That was a completely unrelated diff.

It is *not* in `flaky-tests.txt`, so neither runner retries it automatically.
Filed as #7927 with the measurement, rather than quarantining it from inside
this PR.

I have re-run the failed job — that is the one re-run the drive-to-green rules
allow. A second failure will be treated as real.

---

tokuhirom:
Correction to the comment above: the `kill.t` flake is filed as **#7929**, not
#7927 (I wrote the number before the issue was created). Everything else in that
comment stands.

---

Issue #7937, opened 2026-09-11
`.^isa` on a concrete builtin value answers False, and `.^mro` drops a parametrized name's base row


tokuhirom:
Two `.^`-MOP answers that do not consult the builtin-type catalog / E1 classifier, found while writing regression cover for [#7766](https://github.com/tokuhirom/mutsu/issues/7766) (they are NOT caused by that work — both reproduce identically on `main`, measured at `7c1e6346`). Filed together because they look like one family: the `.^`-MOP surface deriving ancestry from something other than the chain `receiver_class`/`Registry::class_mro_readonly` already compute correctly.

## 1. `.^isa` is False for a *concrete builtin value*

```
$ mutsu -e 'say "x".^isa(Cool); say True.^isa(Int); say 1.5.Rat.^isa(Cool)'
0
0
0

$ raku -e 'say "x".^isa(Cool); say True.^isa(Int); say 1.5.Rat.^isa(Cool)'
1
1
1
```

It is specifically the *concrete value* receiver. A type object and a user class are both right already:

```
$ mutsu -e 'class Foo {}; class Bar is Foo {}; say Bar.^isa(Foo); say Int.^isa(Cool)'
1
1          # both agree with raku
```

And the non-MOP `.isa` on the same values is right, so the ancestry is known — only this entry point misses it:

```
$ mutsu -e 'say True.isa(Int)'
True       # agrees with raku
```

`Bool`'s chain is already correct everywhere else (`mutsu -e 'say Bool.^mro.map(*.^name)'` gives `(Bool Int Cool Any Mu)`, same as raku), which is what makes this look like a wiring gap in the `^isa` handler rather than a missing ancestry.

## 2. `.^mro` for a parametrized name stops at the name

```
$ mutsu -e 'say Array[Int].^mro.map(*.^name).join(" ")'
Array[Int] Any Mu

$ raku -e 'say Array[Int].^mro.map(*.^name).join(" ")'
Array[Int] Array List Cool Any Mu
```

Both `receiver_class::catalog_chain_for_name` and `Registry::class_mro`/`class_mro_readonly` already implement exactly this splice — strip the `[...]`, splice the base type's catalog chain (ADR-0019 E2b twelfth slice, and the comments in both functions say so). `classhow_mro_names` (`src/runtime/methods_classhow_mro.rs`), which is what `.^mro` reads, does not, so the same name gets two different ancestries depending on who asks.

## Where to look

[...]

Issue #7766, opened 2026-09-09
perf: the last ~35 `Symbol::intern` calls per Test assertion — the binder's baked param syms, dispatch names, and one missing pre-interned key


tokuhirom:
Successor to [#7736](https://github.com/tokuhirom/mutsu/issues/7736), which cut the per-assertion `Symbol::intern` budget from **72.0 to 35.4** (PR [#7762](https://github.com/tokuhirom/mutsu/pull/7762), merge `dafffe7`; write-up `news/2026-09/per-assertion-symbol-intern-hoist.md`). That PR did slices 1, 3 and 4 and *half* of slice 2. This is the named residue, measured on the same benchmark, so nobody has to re-derive it.

## Measured, 2026-09-09, at `dafffe7`

Release, callgrind, `tmp/bench-ok-2k.raku` = `use Test; plan 2000; for ^2000 { ok 1, "x" }` under `MUTSU_REAL_TEST=1`, **warm precompilation cache**. Whole run 462.67 M Ir; `Symbol::intern` 94,807 calls / 13.18 M Ir inclusive (2.85%).

Remaining callers, per assertion (from `callgrind_annotate --tree=caller`, cross-checked against a per-string histogram taken as the steady-state slope between a 101- and a 201-assertion run):

| caller | interns/assertion | what it interns |
| --- | --- | --- |
| `bind_function_args_values_inner` | ~7 | each parameter name, once per call |
| `multi_arg_type_keys` + the compiled-method cache | ~5 | receiver class names (`Str`, `Int`, `Bool`, `IO::Handle`) |
| `exec_set_local_op_inner` | ~2 | `__mutsu_scalar_bind_no_container::<name>` |
| `call_compiled_function_named_inner` | ~2 | the routine name (`ok`, `proclaim`) and its package |
| `find_compiled_function_inner` / `resolve_function_multi_cached_keyed` / `user_method_overloads` | ~4 | the callee name, re-hashed once per resolution layer |

## The three units of work

**1. Finish slice 2: the binder should read the baked parameter symbols.**
`CompiledFunction::precompute_param_name_syms` already interns every `param_defs[i].name` at registration time, and `#7762` still does not read it — it interns the name once per parameter per call (a `OnceCell` per parameter, so at most once, but not zero). Threading `param_name_syms: &[Symbol]` into `bind_function_args_values_with_argspec` and indexing it with `param_defs.iter().enumerate()` takes this to zero.

[...]

PR #7940, opened 2026-09-11
perf(dispatch): intern the builtin-type catalog's ancestry chains once, and borrow them

tokuhirom:
`Str.^mro` is a constant. So is `Int`'s, `Bool`'s and every other row of the raku-adjudicated `builtins/builtin_type_catalog` — the table ADR-0019's E1 classifier and `Registry::class_mro_readonly` both read to answer "what is this receiver's ancestry". mutsu rebuilt each of those constants **on every method dispatch that reached it**: an 88-row linear scan of the catalog with a string compare per row to find the row, a `Symbol::intern` per ancestor name to turn it into a chain, and a fresh `Vec`/`Arc` allocation to hold the result.

This is the residue [#7766](https://github.com/tokuhirom/mutsu/issues/7766)'s previous session flagged as *"16.02 interns/assertion charged to `<Vec<T> as SpecFromIterNested>::from_iter`, by far the largest single row in the caller tree… probably worth attributing before starting any of the four splits above."* Attributed: the owner is `receiver_class::catalog_chain_for_name`, at 8,004 calls × 4 ancestors ≈ 32,000 of the row's 33,015 interns.

## What changed

**1. The catalog interns each row's `mro` once per process**, in both shapes its two consumers want, beside the row itself:

- `builtin_type_mro_syms(name) -> Option<Arc<[Symbol]>>` for the registry, whose builtin path hands the chain straight back as an `Arc<[Symbol]>` — now a refcount bump.
- `builtin_type_mro_ids(name) -> Option<&'static [TypeId]>` for the E1 classifier, which splices chains together and so wants to borrow. Minting the `TypeId`s in the catalog keeps that type's invariant intact — it is one of the two places allowed to produce one.

The same table doubles as the catalog's name index, so `builtin_type_info` is an `FxHashMap` probe instead of the linear scan every one of those lookups sat on top of.

**2. The classifier's chain is borrowed, not copied.** Interning once still left a copy per dispatch, because `dispatch_mro` handed out an owned `Vec<TypeId>` — and every consumer only ever *reads* it (`.first()`, `.iter()`, `&chain` into `resolve_sequence`). The chain type is now `Chain = Cow<'static, [TypeId]>`: borrowed straight from the catalog for a plain builtin receiver, owned only where something new is genuinely built (an Enum's `[EnumType, …base]`, a role mixin's role prefix, a user class's registry MRO with a catalog tail spliced on). No call site needed changing; the deref coercions carried it.

## Measured

Release, callgrind, `tmp/bench-ok-2k.raku` = `use Test; plan 2000; for ^2000 { ok 1, "x" }` under `MUTSU_REAL_TEST=1`, warm precompilation cache, second run after each rebuild (the first is not comparable — the ticket's warning is real).

| | before | after |
| --- | --- | --- |
| `Symbol::intern` calls | 98,096 | 66,805 |
| `Symbol::intern` (inclusive Ir) | 17.84 M (3.75%) | 12.89 M (2.73%) |
| `receiver_class::catalog_chain_for_name` (inclusive Ir) | 6.84 M (1.44%) | 1.31 M (0.28%) |
| `receiver_class::dispatch_mro` (inclusive Ir) | 7.49 M (1.57%) | 1.96 M (0.41%) |
| whole run (Ir) | 475.94 M | 472.88 M (−0.64%) |

−31,291 interns, i.e. **−15.6 per assertion** — the largest single cut to that budget so far (#7871 took −11.9). Roughly half the whole-run saving is the interning and the linear scan, the other half the per-dispatch `Vec` the `Cow` removes. `dispatch_mro` is not specific to this benchmark: `type_matches_value` calls it on every typed parameter bind and every smartmatch.

These are local callgrind numbers for review, not document numbers — nothing here is quoted into PERFORMANCE.md or PLAN.md.

## Regression cover

`t/oo/class/builtin-catalog-mro.t` pins what this class of change can get wrong: a shared, pre-built chain that drops, reorders or cross-wires an ancestor is a **wrong ancestry**, not a slowdown — it changes which class answers a method. It checks each catalog row's `.^mro` against raku's, that a repeat read is unchanged, that two user classes deriving from *different* builtins each get their own tail, and that a method augmented onto `Cool` is reachable from `Str`, `Int` and (through `Int`) `Bool`. `interned_views_agree_with_their_row` does the same at the unit level for every row, both interned views at once, plus the miss case.

Two pre-existing `.^`-MOP divergences turned up while writing that file and are filed as [#7937](https://github.com/tokuhirom/mutsu/issues/7937) — `.^isa` on a concrete builtin value, and `.^mro` dropping a parametrized name's base row. **Neither is caused by this PR**; both reproduce identically on `main`. The test uses `.isa` and omits the `Array[Int]` row, with a comment saying which issue to follow to swap them back.

## Verification

- `cargo fmt --all`, `make lint` — green (all four configurations).
- `make test` — green: 3992 files, 42505 tests.
- `make roast` — green except exactly the three failures `docs/agent-environments.md` documents as container artefacts, with the same failing test numbers: `6.c/S32-io/file-tests.t` (`Failed: 4`, tests 6-8/10) and `S16-filehandles/filetest.t` (`Failed: 25`) from running as `uid 0`, and `S32-io/IO-Socket-Async.t` (exit 124) from the sandboxed network.

## Scope

This is one finding taken to completion, not a slice of #7766's four-way split. The `_sym` variant chains for `call_compiled_function_named`, `user_method_overloads`/`has_user_method`, `multi_arg_type_keys` and the resolution layers are untouched; the issue stays open for them and is re-measured in a comment.

Refs #7766.

https://claude.ai/code/session_01Tk6gdMcBJZAkK7iaKTzPLr

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tk6gdMcBJZAkK7iaKTzPLr)_

### DIFF
--- a/news/2026-09/builtin-type-catalog-chains-interned-once.md
+++ b/news/2026-09/builtin-type-catalog-chains-interned-once.md
@@ -1,0 +1,78 @@
+# The builtin-type catalog's ancestry chains are built once, not per dispatch
+
+`Str.^mro` is a constant. So is `Int`'s, `Bool`'s and every other row of
+`builtins/builtin_type_catalog.rs` — the raku-adjudicated table ADR-0019's E1
+classifier and the registry both read to answer "what is this receiver's
+ancestry". mutsu rebuilt each of those constants from scratch on **every method
+dispatch that reached it**: an 88-row linear scan of the catalog with a string
+compare per row to find the row, a `Symbol::intern` per ancestor name to turn it
+into a chain, and a fresh `Vec`/`Arc` allocation to hold the result.
+
+Measured on `use Test; plan 2000; for ^2000 { ok 1, "x" }` under callgrind (the
+benchmark [#7766](https://github.com/tokuhirom/mutsu/issues/7766) is written
+against), `receiver_class::catalog_chain_for_name` alone was **the single
+largest `Symbol::intern` caller in the program** — 16.5 of the ~49 interns per
+assertion, and 1.44% of the run's retired instructions for an answer that could
+not change.
+
+## What changed
+
+The catalog now interns each row's `mro` **once per process**, in both shapes
+its two consumers want, beside the row itself:
+
+- `builtin_type_mro_syms(name) -> Option<Arc<[Symbol]>>` for
+  `Registry::class_mro`/`class_mro_readonly`, whose builtin path hands the chain
+  straight back as an `Arc<[Symbol]>` — now a refcount bump.
+- `builtin_type_mro_ids(name) -> Option<&'static [TypeId]>` for the E1
+  classifier, which splices chains together and so wants to borrow.
+
+The same table doubles as the catalog's name index, so `builtin_type_info` is an
+`FxHashMap` probe instead of the linear scan every one of those lookups sat on
+top of.
+
+Interning the chain once still left a copy of it per dispatch, because
+`dispatch_mro` handed out an owned `Vec<TypeId>`. Every consumer only ever
+*reads* the chain — `.first()`, `.iter()`, `&chain` into `resolve_sequence` — so
+the classifier's chain type is now `Chain = Cow<'static, [TypeId]>`: borrowed
+straight from the catalog for a plain builtin receiver, owned only for the
+composed cases that genuinely build something new (an Enum's `[EnumType,
+…base]`, a role mixin's role prefix, a user class's registry MRO with a catalog
+tail spliced on). No call site needed changing; the deref coercions carried it.
+
+## Measured
+
+Release, callgrind, warm precompilation cache, 2000 assertions, second run after
+the rebuild (the first is not comparable — see the ticket):
+
+| | before | after |
+| --- | --- | --- |
+| `Symbol::intern` calls | 98,096 | 66,805 |
+| `Symbol::intern` (inclusive Ir) | 17.84 M (3.75%) | 12.89 M (2.73%) |
+| `receiver_class::catalog_chain_for_name` (inclusive Ir) | 6.84 M (1.44%) | 1.31 M (0.28%) |
+| `receiver_class::dispatch_mro` (inclusive Ir) | 7.49 M (1.57%) | 1.96 M (0.41%) |
+| whole run (Ir) | 475.94 M | 472.88 M (−0.64%) |
+
+−31,291 interns, i.e. **−15.6 per assertion**, the largest single cut to that
+budget so far (#7871 took −11.9). `dispatch_mro` — which `type_matches_value`
+calls on every typed parameter bind and every smartmatch, not only in this
+benchmark — costs a quarter of what it did. Roughly half the whole-run saving is
+the interning and the linear scan; the other half is the per-dispatch `Vec` the
+`Cow` removes.
+
+## Regression cover
+
+`t/oo/class/builtin-catalog-mro.t` pins the thing this class of change can get
+wrong: a shared, pre-built chain that drops, reorders or cross-wires an ancestor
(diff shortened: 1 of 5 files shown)


tokuhirom:
## `wasm-e2e` is blocking this PR, and the failure is not this PR's

**Every other check is green** on `13b1b31`: `test`, `lint-configs`, `gc-stress`, `jit-stress`, `changes`, `label` all succeeded; `miri` is skipped by design. `mergeable_state` is `blocked` on `wasm-e2e` alone.

`wasm-e2e` is not *failing* — it is being **cancelled by its own `timeout-minutes: 15`** while `npx playwright install --with-deps chromium` waits on apt. The run is on `run_attempt: 4` and every attempt whose log I read died in that step, never reaching `Run E2E tests`:

```
# attempt 1 (job 103173845517)
06:54:36  Installing dependencies...            <- npx playwright install --with-deps chromium
06:59:31  Ign:2 http://archive.ubuntu.com/ubuntu noble InRelease
07:01:01  ##[error]The operation was canceled.  <- 15:12 after job start

# attempt 3 (job 103184973650)
07:44:46  Hit:2 http://archive.ubuntu.com/ubuntu noble InRelease
07:45:05  Get:3 http://archive.ubuntu.com/ubuntu noble-updates InRelease [126 kB]
07:48:03  ##[error]The operation was canceled.  <- 15:12 after job start
```

Attempt 4 (job 103192006584) is in the same place: `Install Playwright` began 08:10:07 and is still there.

**Why it is not this PR's.** The diff is `src/builtins/builtin_type_catalog.rs`, `src/runtime/receiver_class.rs`, `src/runtime/registry.rs`, one `.t` and one `news/` file. It touches no `site/`, no wasm code, and `.github/workflows/ci.yml` is byte-identical to `main` (`git diff HEAD origin/main -- .github/workflows/ci.yml` is empty). The failing step runs before any mutsu code is exercised, and the same job passed on `main` at 05:52 (run 34567623251) — the stall began somewhere in the 06:00–06:45 window and has persisted since.

**No fix exists to port.** The repo already carries a mitigation for exactly this — the `Use canonical apt mirror` step, whose own comment says `--with-deps` "hits the same azure.archive.ubuntu.com mirror hang as the native jobs". It applied correctly on every attempt (the logs show `http://archive.ubuntu.com/ubuntu/`); the canonical mirror is now the one stalling, so the existing workaround no longer covers this case.

**I have spent my one re-run** (the `rerun_failed_jobs` at 07:16) and am not re-running again — the failure is reproducing, not flaking.

### Proposed patch — deliberately NOT pushed here

This would widen a perf PR into the CI pipeline definition, and `ci.yml` is the one `.github/**` path excluded from the docs-only allowlist precisely because an edit to it must not ride along untested. It belongs in its own PR:

```diff
       - name: Install Playwright
         run: |
           npm install playwright
-          npx playwright install --with-deps chromium
+          # `--with-deps` is the only apt consumer in this job and it is what
+          # hangs when the Ubuntu mirror stalls; ubuntu-latest already ships
+          # the shared libraries Chromium needs.
+          npx playwright install chromium
```

Dropping `--with-deps` removes the job's only apt dependency. If the system libraries do turn out to be needed on a future runner image, the alternative is to keep `--with-deps` and raise `timeout-minutes`, but that only widens the window — `Build npm package` alone consistently takes ~8.5 of the 15 minutes, so this job has under 7 minutes of headroom for install plus tests even on a healthy runner, which is worth revisiting either way.

Happy to open that as a separate PR if you want it. Meanwhile this PR needs either that fix on `main`, the mirror recovering, or an admin merge.

---

Issue #7939, opened 2026-09-11
A listop imported from a module followed by `<...>` parses as `<` (blocks `String::Utils`' `t/01-basic.rakutest` at line 33)


tokuhirom:
Found while working [#7931](https://github.com/tokuhirom/mutsu/issues/7931) (the `CURSOR_MORE` cursor-protocol gap). This is what now stops `String::Utils`' `t/01-basic.rakutest` — earlier in the file than #7931, at line 33 — so that record still cannot move.

A sub called as a listop with a `<...>` word-quote argument parses fine when the sub is **declared in the same file**, and fails to parse when the very same sub is **imported from a module**: the parser does not know the name is a routine, so `root <abcd` is taken as the infix `<`.

## Repro

Declared locally — both agree:

```raku
sub root(*@s) { @s.elems }
say (root <abcd abce abde>);
```

| | output |
|---|---|
| `raku` | `3` |
| `mutsu` | `3` |

Imported — `mutsu` cannot parse it:

```raku
use String::Utils;          # exports `root`
say (root <abcd abce abde>);
```

| | output |
|---|---|
| `raku` | `ab` |
| `mutsu` | <pre>===SORRY!=== Error while compiling tmp/rootgap2.raku<br>Confused. expected statement: expected expression after infix operator or '.' or digits or generic radix literal or unicode numeric literal or ...<br>at tmp/rootgap2.raku:2<br>------>say (root &lt;abcd abce abde&gt;);<br>       ^</pre> |

(`String::Utils` 0.0.40, run with `-I lib` against the vendored dist tarball named in `ecosystem/dists/S/String--Utils~aed281d9.json`.)

## Why it is deep

The disambiguation `term <a b c>` vs `expr < expr` needs the parser to know, *at parse time*, that the preceding identifier names a routine. mutsu's parser tracks locally-declared subs, but a name that only exists because a `use`d module exported it is not in that set when the statement is parsed — module loading happens later in the pipeline. Fixing it means either making the parser consult the importer's symbol table (i.e. loading the module's export list during parse, which is what rakudo's `BEGIN`-time `use` does), or finding a parse-only heuristic for `identifier <word word ...>` that cannot misfire on a genuine chained comparison.

Both are architectural calls rather than patches, which is why this is filed rather than fixed inline.

## Where else it bites

[...]

Issue #7883, opened 2026-09-11
The regex cursor protocol (`Match.^lookup("!cursor_init")`, `$!pos` / `$!from` on a cursor) is not available to user code


tokuhirom:
Rakudo's `Match` exposes the NQP-level cursor protocol that its own regex engine runs on, and ecosystem code reaches for it to drive a regex by hand — matching at a known position and reading the cursor's `$!from` / `$!pos` instead of going through `~~` and a `Match` object. mutsu has no `!cursor_init`, so the whole idiom dies.

## Repro

```raku
my $cursor-init = Match.^lookup("!cursor_init");
my $cursor := /foo/($cursor-init(Match, "xfoox", :0c));
say $cursor.pos;
```

| | output |
|---|---|
| `raku` | `4` |
| `mutsu` | `No such method 'CALL-ME' for invocant of type 'Nil'` (the `^lookup` returns Nil) |

## Where it bites

`String::Utils`'s `replace` is written entirely in it
(`ecosystem/dists/S/String--Utils.json`, `lib/String/Utils.rakumod:563`):

```raku
my constant $cursor-init = Match.^lookup("!cursor_init");
my sub replace(str $haystack, Regex:D $needle, str $replacement) {
    my $cursor := $needle($cursor-init(Match,$haystack,:0c));
    my int $pos = nqp::getattr_i($cursor,Match,'$!pos');
    $pos >= 0
      ?? nqp::substr($haystack,0,nqp::getattr_i($cursor,Match,'$!from'))
           ~ $replacement
           ~ nqp::substr($haystack,$pos)
      !! $haystack
}
```

so every `replace`/`replace-all` assertion in that distribution's `t/01-basic.rakutest` dies rather
than failing (the subtest at line 160 aborts). `nqp::getattr_i($cursor, Match, '$!pos')` itself
works now — the missing piece is purely the cursor object that `!cursor_init` would produce and
the ability to invoke a `Regex` on it.

## Why it is deep

This is not one method. It means exposing mutsu's regex engine through rakudo's cursor interface:

- `Match.^lookup` must find the private `!cursor_init` method, i.e. `Match` needs the private
  methods in its MOP surface, not just its public ones;
- a *cursor* has to exist as a first-class value distinct from a completed `Match` — carrying
  `$!orig`, `$!from`, `$!pos`, and `:c` (the start position) — and mutsu's regex engine would have
  to be able to start from one and leave its end position in it;
[...]

PR #7946, opened 2026-09-11
feat(regex): CURSOR_MORE, the walk half of the NQP cursor protocol

### LINKED ISSUE #7931: `Match.^lookup` does not expose the NQP cursor protocol (`!cursor_init`, `CURSOR_MORE`), so String::Utils' `replace-all` dies
Split out of [#7881](https://github.com/tokuhirom/mutsu/issues/7881), which fixed the *parse* gap that made `String::Utils`' `t/01-basic.rakutest` unrunnable. The file now parses and reaches 117 of a 124-assertion `raku` baseline; this issue is the first thing that stops it going further.

`String::Utils` drives the regex engine through rakudo's internal cursor protocol rather than through `.match`:

```raku
my constant $cursor-init = Match.^lookup("!cursor_init");
my constant $global      = Match.^lookup("CURSOR_MORE");

my sub replace-all(str $haystack, Regex:D $needle, str $replacement) {
    my $cursor := $needle($cursor-init(Match, $haystack, :0c));
    my int $pos = nqp::getattr_i($cursor, Match, '$!pos');
    ...
}
```

## Repro

```raku
my constant $cursor-init = Match.^lookup("!cursor_init");
say $cursor-init.^name;
my constant $global = Match.^lookup("CURSOR_MORE");
say $global.^name;
```

| | output |
|---|---|
| `raku` | `NQPRoutine` / `Method+{is-implementation-detail}` |
| `mutsu` | `Method` / `Mu` |

`CURSOR_MORE` resolves to nothing at all, so the call site above invokes a type object and dies with:

```
No such method 'CALL-ME' for invocant of type 'Nil'
  in sub replace-all at lib/String/Utils.rakumod line 581
```

`!cursor_init` does return *a* `Method`, but there is no evidence it implements the protocol the caller then uses (`nqp::getattr_i($cursor, Match, '$!pos')` / `'$!from')`), so a fix has to cover both names together rather than just adding the missing lookup entry.

## Why it is deep

This is not one missing method: it is the boundary between mutsu's own regex engine (`src/runtime/regex.rs`, span-based captures per [ADR-0016](https://github.com/tokuhirom/mutsu/blob/main/docs/adr/0016-span-based-captures-and-lazy-match.md)) and rakudo's NQP-level `Cursor` representation. Making `Match.^lookup("!cursor_init")` hand back something a caller can drive means deciding how much of the `Cursor` shape — `$!pos`, `$!from`, `CURSOR_MORE`'s resumption semantics — mutsu wants to expose as a stable internal surface, and whether `nqp::getattr_i` on a `Match` reads real attributes or synthesised ones. That is an architectural call, not a patch.

## Where else it bites

The idiom is lizmat's house style for "walk every match without re-scanning", so the same protocol is likely to be needed by the other `String::*` / `Array::*` distributions in `ecosystem/`. Worth measuring the blast radius before designing the surface.

Also still open in the same file (independent of this one): assertions 29, 33 and 38 (`is SHA1 ok`, `does sha1 produce a valid SHA1`, `did trigrams work`) fail with wrong values rather than dying.

tokuhirom:
`Match.^lookup("CURSOR_MORE")` resolved to nothing, so the second half of the NQP cursor protocol was unavailable to ecosystem code. `String::Utils`' `replace-all` is written entirely in it and died with `No such method 'CALL-ME' for invocant of type 'Nil'`. #7883 landed the first half (`!cursor_init` plus calling a `Regex` on a cursor); this completes it.

## What the protocol asks for

rakudo's `Cursor.CURSOR_MORE` re-invokes the cursor's own `$!regexsub` on a fresh un-started cursor placed just past the last match, so a caller can walk every match without re-scanning the prefix. Two details of that were measured against rakudo 2026.07 and are load-bearing:

- the resumption position is `$!pos`, **bumped by one when the last match was zero-width** (`$!from == $!pos`). Without the bump, `/x*/` finds the same empty match forever and `replace-all` never terminates.
- the fresh cursor is un-started (`$!from == -1`), so the re-invocation *scans* forward rather than anchoring, and a run that finds nothing more yields the ordinary failed cursor (`$!pos == -3`) the caller's loop condition tests for.

All traces now agree with the oracle exactly:

| regex / haystack | `[$!from,$!pos]` per step | before |
| --- | --- | --- |
| `/o/` on `"foo boo"` | `[1,2] [2,3] [5,6] [6,7] [7,-3]` | `[1,2]` forever |
| `/\d+/` on `"a12b345"` | `[1,3] [4,7] [7,-3]` | `[1,3]` forever |
| `/x*/` on `"axb"` | `[0,0] [1,2] [2,2] [3,3] [4,-3]` | `[0,0]` forever |

## How mutsu remembers the regex

mutsu's `Match` has no `$!regexsub` slot, so it grows one: the internal attribute `CURSOR_REGEXSUB_ATTR` holds **the callable that produced the cursor**, stamped on the result of every cursor-protocol regex call. Storing the callable rather than a pattern string means `CURSOR_MORE` re-invokes it through the same `Regex.CALL-ME` path the original `$needle($cursor)` call took, so an `rx:i//`'s adverbs — and any later fix to how they are honoured — are inherited instead of reconstructed.

The stamp preserves the `Match` instance id (a cursor is handed straight to user code), and the lazy `Match` node (ADR-0016 P5) answers `None` for the attribute without forcing, alongside the other post-hoc keys, so a probe for it on an ordinary match costs nothing.

## The surface this deliberately does not claim

Both are reported with a specific message rather than guessed at:

- **an ordinary `"abc".match(/b/)`.** In rakudo every `Match` *is* a spent `Cursor` and carries its `$!regexsub`, so `CURSOR_MORE` works on one. Here the stamp would have to go on every match the engine produces, and because it rebuilds the `Match` eagerly that would cost ADR-0016 P5's laziness on the hottest path in the interpreter — for a surface no known consumer reaches for. Making it free instead means threading the invoked regex down to the engine entry points, the same objection the `cursor_class` stamp already records, and worth its own issue if a consumer ever needs it.
- **a grammar token method called on a cursor.** Its result feeds the custom-HOW subrule side channel by instance identity, so stamping the regexsub on it would perturb a path unrelated to this protocol.

A cursor that never ran, and one whose match failed, are also refused — rakudo dies with a null `$!regexsub` on both of those too.

Unchanged: `.^lookup`'s return value is a plain mutsu `Method`, where rakudo answers an `NQPRoutine` for `!cursor_init` and a `Method+{is-implementation-detail}` for `CURSOR_MORE`. Both are callable, which is all the idiom uses; mutsu has no NQP-level routine type to report.

## Result

The real, vendored `String::Utils` now runs its `replace-all` verbatim under mutsu with output identical to rakudo's (`XbarX`, `a-b-c-`, `<><>`).

Its own `t/01-basic.rakutest` is still blocked *earlier* than this, at a separate parse gap on line 33 (an imported listop followed by a `<...>` word quote), filed as #7939 — so that ecosystem record does not move yet.

## Tests

New `t/regex/match/match-cursor-more.t`, 24 assertions, **every expectation first run green against `raku` 2026.07**, next to `match-cursor-protocol.t` from #7883. Covers the walk traces above, the zero-width termination, `:i` surviving the resume, the resumed cursor's `.from`/`.to`/`.Str`/`.Bool`/`.^name`, character-not-byte positions, the two refusals, and the verbatim `String::Utils::replace-all` body.

## Verification

- `cargo fmt --all` — clean
- `make lint` — green (default clippy, `jit` off, wasm32 lib, rustdoc)
- `make check-t-layout` — green
- `make test` — **3994 files / 42524 tests, all pass**
- `make roast` — 1433 files / 219306 tests; the only failures are the three documented container-only exceptions, matching their recorded signatures exactly (`roast/6.c/S32-io/file-tests.t` `Failed: 4` and `roast/S16-filehandles/filetest.t` `Failed: 25`, both because the container runs as `uid 0`; `roast/S32-io/IO-Socket-Async.t` exit 124 "planned 40 ran 17" from the sandboxed network — see `docs/agent-environments.md`)

Closes #7931

https://claude.ai/code/session_01PijqPLBuF8QM4aW9N5BULV

---
_Generated by [Claude Code](https://claude.ai/code/session_01PijqPLBuF8QM4aW9N5BULV)_

### DIFF
--- a/news/2026-09/cursor-more-nqp-cursor-protocol.md
+++ b/news/2026-09/cursor-more-nqp-cursor-protocol.md
@@ -1,0 +1,106 @@
+# `CURSOR_MORE`: walking every match through the NQP cursor protocol
+
+`Match.^lookup("CURSOR_MORE")` resolved to nothing, so the second half of the
+NQP cursor protocol was unavailable to ecosystem code. `String::Utils`'
+`replace-all` is written entirely in it and died with
+`No such method 'CALL-ME' for invocant of type 'Nil'`
+([#7931](https://github.com/tokuhirom/mutsu/issues/7931)). The first half —
+`!cursor_init` plus calling a `Regex` on a cursor — landed in
+[#7883](https://github.com/tokuhirom/mutsu/issues/7883); this completes it.
+
+## What the protocol asks for
+
+lizmat's house style for "walk every match without re-scanning the prefix" is
+to drive the regex engine by hand rather than through `~~`:
+
+```raku
+my constant $cursor-init = Match.^lookup("!cursor_init");
+my constant $global      = Match.^lookup("CURSOR_MORE");
+
+my $cursor := $needle($cursor-init(Match, $haystack, :0c));
+my int $pos = nqp::getattr_i($cursor, Match, '$!pos');
+while $pos >= 0 {
+    ...                                    # $!from .. $!pos is one match
+    $cursor := $global($cursor);
+    $pos = nqp::getattr_i($cursor, Match, '$!pos');
+}
+```
+
+rakudo's `Cursor.CURSOR_MORE` re-invokes the cursor's own `$!regexsub` on a
+fresh un-started cursor placed just past the last match. Two details of that
+were measured against rakudo 2026.07 and are load-bearing:
+
+- the resumption position is `$!pos`, **bumped by one when the last match was
+  zero-width** (`$!from == $!pos`). Without the bump, `/x*/` finds the same
+  empty match forever and `replace-all` never terminates.
+- the fresh cursor is un-started (`$!from == -1`), so the re-invocation *scans*
+  forward rather than anchoring, and a run that finds nothing more yields the
+  ordinary failed cursor (`$!pos == -3`) the caller's loop condition tests for.
+
+Both traces now agree with the oracle exactly, including the zero-width walk
+that is the reason the bump exists:
+
+| regex / haystack | `[$!from,$!pos]` per step |
+| --- | --- |
+| `/o/` on `"foo boo"` | `[1,2] [2,3] [5,6] [6,7] [7,-3]` |
+| `/\d+/` on `"a12b345"` | `[1,3] [4,7] [7,-3]` |
+| `/x*/` on `"axb"` | `[0,0] [1,2] [2,2] [3,3] [4,-3]` |
+
+## How mutsu remembers the regex
+
+mutsu's `Match` has no `$!regexsub` slot, so it grew one: the internal
+attribute `CURSOR_REGEXSUB_ATTR`, holding **the callable that produced the
+cursor**, stamped on the result of every cursor-protocol regex call. Storing
+the callable rather than a pattern string means `CURSOR_MORE` re-invokes it
+through the same `Regex.CALL-ME` path the original `$needle($cursor)` call
+took, so an `rx:i//`'s adverbs — and any later fix to how they are honoured —
+are inherited instead of reconstructed. The stamp preserves the `Match`
+instance id, because a cursor is handed straight to user code.
+
+The lazy `Match` node (ADR-0016 P5) answers `None` for the attribute without
+forcing, alongside the other post-hoc keys, so a probe for it on an ordinary
+match costs nothing.
+
+## The surface this deliberately does not claim
+
+`CURSOR_MORE` resumes a cursor produced by a `Regex` called on a cursor. Two
+narrower surfaces are left out, and it says so rather than guessing:
+
+- **an ordinary `"abc".match(/b/)`.** In rakudo every `Match` *is* a spent
+  `Cursor` and carries its `$!regexsub`, so `CURSOR_MORE` works on one. Here
+  the stamp would have to go on every match the engine produces, and because it
+  rebuilds the `Match` eagerly that would cost ADR-0016 P5's laziness on the
+  hottest path in the interpreter — for a surface no known consumer reaches
+  for. Making it free instead means threading the invoked regex down to the
+  engine entry points, which is the same objection the `cursor_class` stamp
+  already records, and worth its own issue if a consumer ever needs it.
+- **a grammar token method called on a cursor.** Its result feeds the
+  custom-HOW subrule side channel by instance identity, so stamping the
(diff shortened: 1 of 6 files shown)


tokuhirom:
## `wasm-e2e` is red on Ubuntu's apt mirror, not on this diff

**Failing check:** `wasm-e2e` (both attempts). Every other job is green on this head: `test`, `lint-configs`, `gc-stress`, `miri`, `changes`, `label`.

**Why it is not this PR's.** It dies in `npx playwright install --with-deps chromium`, before anything in this repo is built or run, with Ubuntu's archive mirror mid-sync:

```
Err:11 mirror+file:/etc/apt/apt-mirrors.txt noble-updates/main amd64 Packages
  File has unexpected size (1261512 != 1261524). Mirror sync in progress? [IP: 91.189.91.82 80]
E: Failed to fetch .../noble-updates/main/binary-amd64/Packages
E: Some index files failed to download. They have been ignored, or old ones used instead.
Failed to install browsers
Error: Installation process exited with code: 100
```

It reproduced identically on the one re-run. The first attempt was the same apt stall presenting as the job's 15-minute `timeout-minutes` (`Ign:2 http://archive.ubuntu.com/ubuntu noble InRelease`, then cancellation at exactly 15:13); that run-level cancellation is also what killed `jit-stress` mid-`prove` at 07:37:40 — collateral, not a `jit-stress` failure. On the re-run, `apt` reached the mirror and reported the sync explicitly.

`archive.ubuntu.com` is a service this diff does not touch: the change is Rust under `src/`, one `.t`, and one `news/` file — nothing under `site/`, no wasm, no workflow. And CI run 18927 on `6d13b171` — this PR's own base commit — was green 40 minutes earlier, so the mirror window opened after that.

**No fix to port.** A mirror sync clears on its own. The only in-repo change that would harden this is editing `ci.yml`'s `playwright install` step (dropping `--with-deps`, or retrying `apt`), and that is a change to a required CI gate and to the file that defines these five jobs — not something to slip into a feature PR unilaterally. If it is worth doing, it belongs in its own PR.

**What it needs.** One more `wasm-e2e` re-run once the mirror has finished syncing; it will also clear on any later push to this branch. I have spent my single re-run budget, so I am not re-running it again. Auto-merge is on and will land the PR by itself the moment that check is green.

Keeping the PR watched until it is merged or closed.

---

Issue #7943, opened 2026-09-11
Concurrent native mutable methods on one instance clobber each other: the attribute map is committed wholesale


tokuhirom:
## The mechanism

Every native *mutable* instance method follows a snapshot-and-commit convention:

```rust
// src/runtime/native_methods/mod.rs:276
pub(super) fn call_native_instance_method_mut(
    &mut self,
    class_name: &str,
    attributes: AttrMap,       // <- an owned CLONE of the instance's map
    method: &str,
    args: Vec<Value>,
) -> Result<(Value, AttrMap), RuntimeError>
```

and the caller publishes the returned map afterwards:

```rust
// src/runtime/methods_mut_dispatch.rs:3005
match self.call_native_instance_method_mut(&class_name.resolve(), attributes.to_map(), method, args.clone()) {
    Ok((result, updated)) => {
        self.env.insert(
            target_var.to_string(),
            Value::write_back_sharing(&attributes, class_name, updated, target_id),
        );
```

`write_back_sharing` is `attrs.commit_attrs(map.into())` — a **whole-map replacement**, not a merge.

Two consequences when two threads call methods on the same instance:

1. **A window where a write is invisible.** Everything a method writes is invisible to every other thread until it returns *and* the caller commits. A method that wakes another thread partway through (resolving a promise, sending on a channel, spawning) has published nothing yet.
2. **Lost updates.** A method that started from an older snapshot commits that older map over a newer one when it returns, silently discarding the other thread's writes — including writes it never touched.

This is the instance-attribute twin of the `env` clobber fixed in #4167 (`t/lock.t`: "a parent-thread stale env sync clobbered wholesale").

## Confirmed instance

#7929 / PR (this branch) is (1): `Proc::Async.start` sets `started` and `pid` into its local copy, resolves `.ready` — waking a thread that calls `.kill` — and only then returns for the caller to commit. `.kill` read `started` out of the instance, saw the pre-spawn map, and threw `X::Proc::Async::MustBeStarted`. Measured 1 failure in 80 runs of `roast/S17-procasync/kill.t`; the `.ready`-racing-`.start` variant of the same bug hung about 3 runs in 4.

That was fixed *locally*, by moving the state onto a `SharedPromise` stored in the map — a promise is shared by reference, so it is the same object in every snapshot and sidesteps the convention entirely. That is a sound fix for one class, not for the convention.

## Why this is `todo:deep`

[...]

Issue #7942, opened 2026-09-11
Bare `sleep` (no parens, no argument) parses as a BareWord and never sleeps


tokuhirom:
## Symptom

`sleep` written with neither parentheses nor an argument returns immediately instead of sleeping forever.

```
$ timeout 3 raku  -e 'sleep' ; echo "rc=$?"
rc=124                      # still sleeping when the 3s timeout fired

$ timeout 3 mutsu -e 'sleep' ; echo "rc=$?"
rc=0                        # returned at once
```

`sleep()` — same routine, empty parens — is correct:

```
$ timeout 3 mutsu -e 'sleep()' ; echo "rc=$?"
rc=124
```

## Root cause

The parser does not resolve the bare name to the routine at all:

```
$ mutsu --dump-ast -e 'sleep'
[
    SetLine(1),
    Expr(
        BareWord("sleep"),
    ),
]
```

The builtin itself is fine — `Interpreter::builtin_sleep` (`src/runtime/builtins_system_async.rs:48`) already treats a missing argument as `Inf`:

```rust
let secs = Self::seconds_from_value(args.first().cloned());
match secs {
    None | Some(f64::INFINITY) => { /* sleep indefinitely */ }
```

So this is purely the argument-less bare-name call form never reaching it.

## Spec

`raku-doc/doc/Type/independent-routines.rakudoc`:

```
=head2 sub sleep

    sub sleep($seconds = Inf --> Nil)
```

The default is `Inf`, so a bare `sleep` blocks forever. That is exactly what rakudo does.

## Why it matters beyond `sleep`

`roast/S17-procasync/kill.t` spawns its child as `$*EXECUTABLE, "-e", "sleep"` — a process that is *supposed* to sit there until the test kills it. Under mutsu that child exits immediately, so several of that file's subtests are signalling processes that have already gone. The file currently passes, but it is not testing what it reads as.

More generally the bug is presumably not specific to `sleep`: any listop-style core routine whose entire argument list is optional will have the same shape. Worth checking the rest (`note`, `say`, `print`, `time`-likes, ...) as part of the fix rather than special-casing this one name — a `BareWord` fallthrough that silently evaluates to nothing is the real defect.

## Affected files

- the parser's bare-identifier path (`src/parser/primary/ident/`) and whatever decides a bare name is a term rather than a zero-argument call
- `src/runtime/builtins_system_async.rs` only as the (already correct) callee

## Notes

[...]

Issue #7609, opened 2026-09-08
Unexplained one-off SIGSEGV in `roast/S17-procasync/stress.t` (CI run 30590633128)


tokuhirom:
Split out of [#7543](https://github.com/tokuhirom/mutsu/issues/7543) on
2026-09-08, when measurement ruled the ADR-0068 cross-thread container-write
class out as its cause. The crash is real and still unexplained; it just does
not belong on that campaign's ledger.

## The event

CI run 30590633128 (2026-07-30) killed the interpreter while
`roast/S17-procasync/stress.t` was running the
[rakudo#3299](https://github.com/rakudo/rakudo/issues/3299) block — 1200
`Proc::Async` instances, each started inside its own `react`:

```raku
my $prog = 'cat';  my @target = '/dev/null';
for ^1200 {
  my $proc = Proc::Async.new($prog, |@target);
  react {
      whenever $proc.start { done }
      whenever signal(SIGTERM) {}
      whenever Promise.in(5) {}
  }
}
print 'pass'
```

The file is on `roast-whitelist.txt` and has been green ever since. It has never
reproduced — roughly 130 targeted runs across four configurations before
2026-09-08, plus the ~180 below.

## What has been ruled out (2026-09-08, ADR-0068 §12)

**It is not a cross-thread aliased container write.** ADR-0068 §1.2's `rust-gdb`
ignore-counter oracle, run on each block extracted to a standalone program:

| probe | `ContainerStructGuard` | `shared_array_elem_set` | unsynchronized aliased store | `gc_contents_mut<ArrayData>` |
|---|---|---|---|---|
| the rakudo#3299 block | 0 | 0 | 0 | **1** |
| block 1 (400 `Proc::Async`, `.tap` writing a captured `$output`, `@got.push`) | 0 | 0 | 0 | **1** |

The single `gc_contents_mut` call in each is not a thread doing anything — its
backtrace is `exec_set_local_op` → `stamp_descriptor_name` for the `my @target`
/ `my @got` declaration, on the main thread, before any `Proc::Async` exists.

Block 1 is worth flagging because it *looks* like ADR-0068 route 1 (a `.tap`
callback closing over `$output`, `@got` pushed from the main loop) and is not:
`$output` is a `Str`, not a container, and the `@got` push is unaliased.

**It is not reproducible at oversubscription either.** ADR-0068 §1.1's harness
[...]

Issue #7936, opened 2026-09-11
A paren-less zero-arg builtin call (`sleep;`, `my $x = exit;`) parses as a BareWord string instead of a call


tokuhirom:
Found while working #7924 — unrelated to that fix, but it is what makes that ticket's own repro racy under mutsu (`Proc::Async.new: $*EXECUTABLE, "-e", "sleep"` spawns a child that exits in 20ms instead of blocking, so the parent's `.kill` races the child's exit).

## Repro

```raku
my $t = now;
sleep;
say "elapsed=", (now - $t).round(0.001);
```

- raku (v2026.07): blocks forever (`timeout 10 raku …` → exit 124)
- mutsu: `elapsed=0`, exit 0

`sleep` is documented as `sub sleep($seconds = Inf --> Nil)` (`raku-doc/doc/Type/independent-routines.rakudoc:1192`), so a bare `sleep` sleeps indefinitely.

## Root cause

`builtin_sleep` is **not** the problem — it already handles the no-args/`Inf` case with an interruptible infinite loop (`src/runtime/builtins_system_async.rs:48-64`), and every other spelling works:

| form | mutsu |
| --- | --- |
| `sleep;` | returns immediately ❌ |
| `sleep();` | blocks ✅ |
| `sleep Inf;` | blocks ✅ |
| `sleep(Inf);` | blocks ✅ |
| `sleep(*);` | blocks ✅ |

The call never reaches the builtin. `--dump-ast` on `sleep;`:

```
Expr(
    BareWord(
        "sleep",
    ),
)
```

versus `sleep();`, which is a proper `Call { name: Symbol("sleep"), ... }`.

`sleep` *is* registered as a listop (`src/parser/primary/ident/predicates.rs:222`, `:274`), but the listop parse needs at least one argument to follow. With nothing after the identifier it falls through to the final catch-all at the end of `parse_identifier_call`, `src/parser/primary/ident/identifier_call.rs:2150`:

```rust
// Method-like: .new, .elems etc. is handled at expression level
Ok((rest, Expr::BareWord(name)))
```

So the result is the **string `"sleep"`**:

```raku
my $x = sleep;
say $x.^name, " ", $x.raku;   # mutsu: Str "sleep"   (raku: blocks forever)
```

[...]

PR #7950, opened 2026-09-11
fix: a Proc::Async that is ready is really ready

### LINKED ISSUE #7929: roast/S17-procasync/kill.t is non-deterministic (~1/60) and is not in flaky-tests.txt
## Symptom

`roast/S17-procasync/kill.t` fails intermittently with `Failed 1/8 subtests`
(exit 1). It is whitelisted, it is **not** listed in `flaky-tests.txt`, so
neither test runner retries it, and a single unlucky roll turns a CI job red on
an unrelated PR.

## Evidence that it is non-deterministic, not a regression

Measured on a release build (branch `claude/container-element-proxy-prelude-compat`,
head ff5b285f) under the exact `gc-stress` environment:

```
MUTSU_GC=on MUTSU_GC_EVERY_CANDIDATE=1024 MUTSU_GC_VERIFY=1 \
MUTSU_ROAST_TIMEOUT_SCALE=2 MUTSU_FUDGE=1 MUTSU_BIN=target/release/mutsu \
prove -e 'scripts/run-roast-test.sh' roast/S17-procasync/kill.t
```

**1 failure in 61 runs** on an otherwise idle box (batches of 8, 12 and 40; the
single failure was in the first batch). `make roast` on the same build was green.

## CI sightings

| run | job | PR | diff | outcome |
|---|---|---|---|---|
| 34552715239 | `test` | #7895 | QuantHash subclass backing store | subtest 7 (`.kill kills when multi-procs kill in multi-promises`) failed; **passed on re-run** |
| 34562755709 | `gc-stress` | #7926 | `is raw` on an anonymous sub | `Failed 1/8`; job red |

Two different PRs, two different jobs, neither diff anywhere near Proc::Async,
signals, promises or supplies.

## Why this is worth a ticket rather than a quarantine entry

`docs/flaky-test-policy.md` sets an evidence standard before anything goes into
`flaky-tests.txt`, and CLAUDE.md's triage protocol is explicit that most
historical "flaky" labels turned out to be deterministic bugs. A ~1.6% rate on
an idle box is consistent *either* with genuine timing non-determinism in a
multi-process signal test *or* with a real race that a loaded machine makes more
likely — and this file is precisely the shape (`Proc::Async`, `signal(...).act`,
several procs killed inside several promises) where those two are hard to tell
apart from the outside.

So the question this ticket asks is **which**, not "please quarantine it":

1. Reproduce under load with `scripts/flake-repro.sh` to get a rate that is not
   dominated by an idle box, and capture a *verbose* failing run — the
   `gc-stress` job does not do the failed-file re-run the `test` job does, so
   the CI log only ever says `Failed 1/8` without naming the subtest.
2. If subtest 7 is the repeat offender, look at whether the multi-proc /
   multi-promise kill has a lost-wakeup or an ordering assumption — the two
   S17 flakes that were previously root-caused (`S17-supply/batch.t`,
   `t/lock.t`) both turned out to be deterministic logic bugs, not load noise.
3. Only if it is genuinely timing-bound, add it to `flaky-tests.txt` with a
   review date, per the policy.

## Affected files

- `roast/S17-procasync/kill.t` (vendored, read-only)
- likely `src/runtime/builtins_system_proc.rs` / the `Proc::Async` + `signal`
  plumbing, and `src/runtime/react_whenever.rs` for the promise side
- `flaky-tests.txt` only as the last-resort outcome

tokuhirom:
`roast/S17-procasync/kill.t` had been failing about once in eighty runs, on two different CI jobs and two unrelated diffs. It was filed as a suspected flake ([#7929](https://github.com/tokuhirom/mutsu/issues/7929)) rather than quarantined, because `docs/flaky-test-policy.md` wants evidence first and because the two S17 flakes previously root-caused (`S17-supply/batch.t`, `t/lock.t`) both turned out to be deterministic logic bugs. This one does too.

## The failure

The `gc-stress` job never names the failing subtest, so the first job was to capture one. `scripts/flake-repro.sh -n 40 -l 4` under the gc-stress environment:

```
# Subtest: .kill kills when multi-procs kill in multi-promises
    1..3
    ok 1 - program did not hang
    ok 2 - STDOUT
    not ok 3 - STDERR
    # expected: ""
    #      got: "Unhandled exception in code scheduled on thread 7\n
    #             X::Proc::Async::MustBeStarted()\n  in block <unit> at -e line 5\n"
```

It does not hang. It dies of `MustBeStarted`, thrown by a `.kill` that ran *after* `await $p.ready` had already returned:

```raku
start {
    await $p.ready;                        # returns...
    $n == 3 ?? $p.kill !! $p.kill: (...)   # ...and this says "must be started"
}
await $p.start
```

## Root cause: the snapshot convention

A native *mutable* instance method is handed an owned **clone** of the instance's attribute map and returns a new one, which the caller commits afterwards (`call_native_instance_method_mut` → `Value::write_back_sharing` → `commit_attrs`). Nothing a method writes is visible to another thread until it has returned *and* the caller has published the map.

| `src/runtime/native_proc_async.rs`, the `"start"` arm | |
|---|---|
| `attrs.insert("started", TRUE)` | into the local copy |
| `attrs.insert("pid", pid)` | into the local copy |
| `ready.try_keep(pid)` | **the killer thread wakes here** |
| … | rest of `start`'s body |
| `Ok((ret, attrs))` | only now does the caller publish the map |

The woken thread's `"kill"` arm reads `attrs.get("started")`, sees the pre-spawn map, and throws. The window is the tail of `start` plus the return and commit — short, hence 1-in-80 rather than always.

The same convention broke the **other half** of the handshake much harder. `.ready` used to mint its promise on demand and hand it to `.start` through the map. A `.ready` racing `.start` built that promise from a pre-spawn snapshot; `.start`, already past its own resolve point holding an older snapshot, never saw it and never kept it — so `await $p.ready` hung forever. Measured at roughly **three runs in four**, far more often than the roast failure that led here.

## The fix

Build the promise in the constructor, and have `.start` keep it with the pid as soon as `spawn()` succeeds. A promise is shared by reference rather than copied with the map, so it is the same object in every snapshot however stale, and `keep` publishes status and value under one lock before waking anyone.

`kill`, `write`, `say`/`put`/`print` and `close-stdin` ask `proc_async_spawned_pid` instead of reading `started`/`pid` directly. It consults the latch **only when the caller's own snapshot has no `pid`**, so the check is monotone — an un-started `Proc::Async` still throws `MustBeStarted`, as `roast/S17-procasync/basic.t` requires (covered by assertion 6 below).

`Proc::Async.pid` is fixed for free: it used to hand back a freshly minted, never-kept promise when it found neither a `pid` nor a `ready_promise`.

## Measurements

`roast/S17-procasync/kill.t`, release build, gc-stress environment:

| | runs | failures |
|---|---|---|
| before, idle box | 140 | 1 |
| before, `flake-repro.sh -l 4` | 40 | 1 |
| **after, `flake-repro.sh -l 4`** | **60** | **0** |

Nothing added to `flaky-tests.txt`: the answer to "flaky or broken" was broken.

For the record, `1a19f33a` ("deliver `whenever <Promise>` off the waiter list") does not fix this — the rate was unchanged from the 1/61 first measured on a commit predating it. That change is real and valuable for #7609; it is not this bug.

## Tests

`t/concurrency/procasync-ready-latch.t`, 6 assertions, **all verified against rakudo**. It fails **6 times out of 6** against a binary without this change and passes **8/8** with it, in 0.4s. It covers the racing `.ready` (repeated, since which side wins is a genuine race), the `.kill`-after-`.ready` shape `kill.t` exercises, `.ready` fetched before `.start`, `.ready` asked for after the process exited, and the negative case that an un-started `Proc::Async` still refuses `.kill`.

`make lint`, `make test` (3991 files / 42475 assertions) and `make roast` are green locally, apart from the three container artifacts `docs/agent-environments.md` lists — `6.c/S32-io/file-tests.t` and `S16-filehandles/filetest.t` (uid 0) and `S32-io/IO-Socket-Async.t` (sandboxed network) — which match the documented failure shapes test-number for test-number.

## What this does not fix

Two findings filed rather than folded in:

- **[#7943](https://github.com/tokuhirom/mutsu/issues/7943)** (`todo:deep`) — the snapshot-and-commit convention itself. Because `commit_attrs` replaces the map wholesale, a method that started from an older snapshot commits it over a newer one and silently discards another thread's writes. That is the instance-attribute twin of the `env` clobber fixed in #4167, it touches every native mut handler, and it wants a design pass.
- **[#7942](https://github.com/tokuhirom/mutsu/issues/7942)** (`todo:ticket`) — a bare `sleep` (no parens, no argument) parses as `BareWord("sleep")` and never calls the routine, where rakudo sleeps forever (`sub sleep($seconds = Inf)`). `kill.t` spawns its children as `-e "sleep"`, so under mutsu they exit immediately and several of its subtests signal processes that have already gone. The file passes either way, but it is not testing what it reads as. Left out here because bare-name resolution has a far wider blast radius than this ticket.

Closes #7929

https://claude.ai/code/session_01R4XEqYhaYfKvEv2ZeCfGKe

---
_Generated by [Claude Code](https://claude.ai/code/session_01R4XEqYhaYfKvEv2ZeCfGKe)_

### DIFF
--- a/news/2026-09/a-proc-async-that-is-ready-is-really-ready.md
+++ b/news/2026-09/a-proc-async-that-is-ready-is-really-ready.md
@@ -1,0 +1,94 @@
+# A Proc::Async that is ready is really ready
+
+`roast/S17-procasync/kill.t` had been failing about once in eighty runs, on two
+different CI jobs and two unrelated diffs. It was filed as a suspected flake
+([#7929](https://github.com/tokuhirom/mutsu/issues/7929)) rather than quarantined,
+because `docs/flaky-test-policy.md` wants evidence first and because the two S17
+flakes that were previously root-caused — `S17-supply/batch.t` and `t/lock.t` —
+both turned out to be deterministic logic bugs. This one did too.
+
+The failing subtest does not hang. It dies:
+
+```
+    not ok 3 - STDERR
+    #      got: "Unhandled exception in code scheduled on thread 7\n
+    #             X::Proc::Async::MustBeStarted()\n  in block <unit> at -e line 5\n"
+```
+
+`MustBeStarted`, thrown by a `.kill` that ran *after* `await $p.ready` had
+already returned:
+
+```raku
+start {
+    await $p.ready;                        # returns...
+    $n == 3 ?? $p.kill !! $p.kill: (...)   # ...and this says "must be started"
+}
+await $p.start
+```
+
+## The snapshot convention
+
+A native *mutable* instance method is handed an owned **clone** of the
+instance's attribute map and returns a new one, which the caller commits
+afterwards (`call_native_instance_method_mut` → `Value::write_back_sharing` →
+`commit_attrs`). Nothing a method writes is visible to another thread until it
+has returned and the caller has published the map.
+
+`start` writes `started` and `pid` into its copy, spawns, and then resolves
+`.ready` — all before it returns. The thread that `.ready` wakes reads
+`started` out of the instance, where the commit has not landed, and throws.
+
+The same convention broke the other half of the handshake even harder. `.ready`
+used to *mint* its promise on demand and hand it to `.start` through the map. A
+`.ready` that raced `.start` built that promise from a pre-spawn snapshot;
+`.start`, already past its own resolve point and holding an older snapshot,
+never saw it and never kept it. `await $p.ready` then hung forever — measured at
+roughly three runs in four, far more often than the roast failure that led here.
+
+## The fix
+
+The promise is built by the constructor now, and `.start` keeps it with the pid
+as soon as `spawn()` succeeds. A promise is shared by reference rather than
+copied with the map, so it is the same object in every snapshot however stale,
+and `keep` publishes the status and the value under one lock before waking
+anyone. `kill`, `write`, `say`/`put`/`print` and `close-stdin` read "has it
+spawned, and as what pid" from that latch (`proc_async_spawned_pid`), falling
+back to it only when their own snapshot has no `pid` — so the un-started case
+still throws `MustBeStarted`, as `roast/S17-procasync/basic.t` requires. A third
+site is fixed for free: `Proc::Async.pid` used to hand back a freshly minted,
+never-kept promise when it found neither a pid nor a `ready_promise`.
+
+## Measurements
+
+`roast/S17-procasync/kill.t`, release build, under the `gc-stress` environment:
+
+| | runs | failures |
+|---|---|---|
+| before, idle box | 140 | 1 |
+| before, `flake-repro.sh -l 4` | 40 | 1 |
+| after, `flake-repro.sh -l 4` | 60 | 0 |
+
+Nothing was added to `flaky-tests.txt`: the answer to "flaky or broken" was
+broken.
+
+Pinned by `t/concurrency/procasync-ready-latch.t`, which fails 6 times out of 6
+against a binary without this change and passes 8/8 with it, in 0.4s. All six
+assertions were verified against rakudo.
+
+## What this does not fix
+
+The snapshot-and-commit convention itself is still there, and it has a second
+consequence this change only sidesteps: because `commit_attrs` replaces the map
+wholesale, a method that started from an older snapshot commits it over a newer
+one and silently discards another thread's writes. That is the instance-attribute
+twin of the `env` clobber fixed in #4167, it touches every native mut handler,
+and it is filed as [#7943](https://github.com/tokuhirom/mutsu/issues/7943) for a
+design pass rather than patched here.
+
+Root-causing this also turned up
+[#7942](https://github.com/tokuhirom/mutsu/issues/7942): a bare `sleep` — no
+parens, no argument — parses as `BareWord("sleep")` and never calls the routine,
+where rakudo sleeps forever (`sub sleep($seconds = Inf)`). `kill.t` spawns its
+children as `-e "sleep"`, so under mutsu they exit immediately and several of its
+subtests are signalling processes that have already gone. The file passes either
+way, but it is not testing what it reads as.
--- a/src/runtime/methods_object_native_ctors_io.rs
+++ b/src/runtime/methods_object_native_ctors_io.rs
@@ -310,6 +310,23 @@ impl Interpreter {
         let mut attrs = HashMap::new();
         attrs.insert("cmd".to_string(), Value::array(positional));
         attrs.insert("started".to_string(), Value::FALSE);
+        // The spawn latch, kept with the pid by `.start`. It doubles as the
+        // promise `.ready` hands back, and it is built HERE rather than on
+        // demand for a reason: every other attribute is observed through a
+        // *snapshot* of this map (a native mut method takes a copy and the
+        // caller commits the result once it returns), so anything created
+        // inside one method and handed to another through the map is lost when
+        // the two run concurrently. A `.ready` racing `.start` used to mint a
+        // fresh promise from a pre-spawn snapshot, which `.start` — already
+        // past its own keep — never resolved; and a `.kill` woken by `.ready`
+        // read `started`/`pid` from a map the commit had not reached yet.
+        // A promise is shared by reference, so one built at construction is
+        // the same object in every snapshot and both problems disappear.
+        // See `proc_async_spawned_pid`.
+        attrs.insert(
+            "ready_promise".to_string(),
+            Value::promise(SharedPromise::new()),
+        );
         attrs.insert("enc".to_string(), enc);
         attrs.insert(
             "stdout".to_string(),
--- a/src/runtime/native_proc_async.rs
+++ b/src/runtime/native_proc_async.rs
@@ -210,6 +210,42 @@ pub(in crate::runtime) fn proc_async_error(
     }
 }
 
+/// Whether this `Proc::Async` has actually spawned, and as what pid, read
+/// without depending on when the attribute map was snapshotted.
+///
+/// The post-`.start` methods (`kill`, `write`, `say`/`put`/`print`,
+/// `close-stdin`) may run on a thread that was woken by `.ready` while the
+/// `.start` call is still in flight on another one. Because a native mut method
+/// receives a *copy* of the attributes and the caller only commits the result
+/// after it returns, such a thread can hold a map in which `started` is still
+/// `False` and `pid` absent, and would wrongly throw
+/// `X::Proc::Async::MustBeStarted`.
+///
+/// `ready_promise` is built at construction and kept with the pid by `.start`
+/// as soon as `spawn()` succeeds. A promise is shared by reference, not copied
+/// with the map, so every snapshot — however stale — reaches the same latch.
+/// `keep` publishes the status and the value under one lock before waking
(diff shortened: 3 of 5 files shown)


tokuhirom:
`gc-stress` and `jit-stress` came back **cancelled**, not failed: both hit the job's 30-minute cap (`ci.yml` `timeout-minutes: 30`) part-way through the roast step. Every test step that actually ran passed in both.

This is not this PR's failure, and the evidence is in the step timings. Comparing `gc-stress` on this head against the same job on the previous PR (#7926, `ff5b285f`), which passed:

| step | #7926 `ff5b285f` | #7950 `bc30cdd1` | delta |
|---|---|---|---|
| Install prove + mold | **22s** | **2m54s** | **+2m32s** |
| Unit + integration tests (GC on) | 2m39s | 2m39s | 0 |
| TAP tests (`prove t/`, GC on) | 15m53s ✅ | 16m20s ✅ | +27s |
| Build (release, for roast) | 4m19s | 4m31s | +12s |
| Roast tests (`make roast`, GC on) | 3m30s ✅ | **cancelled, 3m30s in** | — |
| **job total** | **27m06s** | **30m13s (cap)** | |

`jit-stress` is the same story on the same run: `Install prove + mold` took **2m15s**, `prove t/` passed (18m42s), and roast was cancelled 3m01s in.

So an `apt` step that ran roughly 6-8x slower than usual ate the margin on a job that habitually finishes at 27 of its 30 minutes. `ci.yml`'s own comment on that timeout records this failure mode exactly:

> the job was hitting the limit and getting canceled even when every test passed (roast printing "Result: PASS" immediately before cancellation)

This PR's own contribution is not the cause: the added test, `t/concurrency/procasync-ready-latch.t`, was measured at **0.78s** under the gc-stress environment and **0.48s** under the jit-stress environment on a debug binary — against a 30-minute budget.

Using the one allowed re-run to confirm. No code change, because there is nothing here to fix in this diff.

Separately filed as [#7951](https://github.com/tokuhirom/mutsu/issues/7951): these two jobs sitting at ~90% of their timeout is a standing fragility that will keep costing re-runs, and it wants its own change rather than a widening of this one.

---

tokuhirom:
Correction to the comment above: the CI-budget issue is [#7956](https://github.com/tokuhirom/mutsu/issues/7956), not #7951 — I cited the number before filing it. Re-run of the two cancelled jobs is queued.

---

tokuhirom:
Re-run result: **`gc-stress` passed** (26m54s), **`jit-stress` was cancelled again** at the 30-minute cap.

`gc-stress` going green on an unchanged head is the confirmation that the earlier cancellation had nothing to do with this diff. And `jit-stress`'s second attempt makes the actual culprit unmistakable:

| step | jit-stress, attempt 2 |
|---|---|
| **Install prove + mold** | **16m05s** |
| Build (debug, for TAP tests) | 1m01s |
| TAP tests (`prove t/`, JIT on) | **cancelled 12m46s in** |
| Build (release), roast | never reached (skipped) |

Sixteen minutes to `apt-get install` two packages, against a normal of 22s. The tests never got a chance to run. Across the two attempts that one step has measured **22s → 2m15s → 2m54s → 16m05s** — the apt path is badly degraded right now, and it is the only thing that changed.

I have used the one allowed re-run, so I am not re-running a third time, and an empty commit to kick CI is not an option. Nothing in this diff can fix a 16-minute apt.

This makes [#7956](https://github.com/tokuhirom/mutsu/issues/7956) considerably more urgent than when I filed it an hour ago — a 16-minute `apt` step means a *45*-minute timeout would not have saved this job either, so option 1 (raise `timeout-minutes`) is not sufficient on its own; option 2 (stop paying for `apt` on every run — cache the packages, or fetch `mold` from its release tarball) is the one that actually addresses it. I have added the measurement there.

Handing this to @tokuhirom rather than editing `ci.yml` from inside this PR: changing the CI pipeline is a decision I should not make unilaterally, and it would widen a `Proc::Async` fix into CI infrastructure. Happy to do it as its own PR on request.

To be explicit about this PR's own state: every job that has actually executed its test steps on `bc30cdd1` is green — `test`, `lint-configs`, `wasm-e2e`, `miri`, `gc-stress`, `changes`, `label`. Only `jit-stress` is outstanding, and it has never reached a test on this head.

---

tokuhirom:
Re-running `jit-stress` once more, at @tokuhirom's explicit instruction — recording that here so the deviation from the one-re-run rule is on the record rather than looking like an agent quietly retrying until green.

The rationale is that the blocker established above is an apt degradation, which may well have passed: the step has been anywhere from 22s to 16m05s within the last hour.

If it is cancelled a third time, I will not re-run again — that would be waiting on infrastructure to heal by chance, and [#7956](https://github.com/tokuhirom/mutsu/issues/7956) option 2 is the fix.

---

tokuhirom:
Closing this: **the same bug was fixed on `main` while this PR sat in CI**, by `fd3f1df3` (PR #7951, from issue #7936). The fix there is the same design — build the `ready_promise` in the constructor so `.ready` and `.start` share one object whichever thread reaches which first, and take `started`/`pid` from it in `.kill`.

CI on `bc30cdd1` did finally go **8/8 green** after the third `jit-stress` attempt, and the PR then went `dirty` against the new `main`. Rather than resolve a conflict in favour of a duplicate, it is being closed.

## Verification that `main` really fixes it

`roast/S17-procasync/kill.t` on `8ecc2f3d`, release build, gc-stress environment, `flake-repro.sh -l 4`:

| | runs | failures |
|---|---|---|
| `f4e5632b` (before) | 40 | 1 |
| **`8ecc2f3d` (now)** | **40** | **0** |

So #7929 is genuinely resolved and I have closed it there rather than here.

## What was independently established here, in case it is useful

This PR and `fd3f1df3` were developed in parallel from the same base without either seeing the other — `f4e5632b` contained neither `fd3f1df3` nor `ff766e7c`. The two arrived at the same root cause and the same fix, which is at least good corroboration. Two things measured here are not recorded on the `main` side:

1. **The failure signature.** The `gc-stress` job never names the failing subtest, so it took a captured verbose run under load to establish that subtest 7 does not *hang* — it dies of `X::Proc::Async::MustBeStarted` from a `.kill` after `await $p.ready` returned. That is what pointed at the snapshot convention rather than at a lost wakeup.

2. **Rate for the `.ready` half.** The `.ready`-racing-`.start` deadlock reproduced in roughly **three runs out of four** (6/20 clean on the pre-fix binary), against 1-in-80 for the `.kill` half. Anyone re-testing this area should use the `.ready` shape, not `kill.t`.

The branch `claude/procasync-ready-before-started-race` is left on the remote so the diff stays readable.

## The one piece of this PR `main` does not have

This PR also routed `write`, `say`/`put`/`print` and `close-stdin` through the same latch. `fd3f1df3` deliberately left those:

> `.write`, `.close-stdin` and the other MustBeStarted gates have the same latent problem

I could **not** reproduce that residue on `8ecc2f3d`: a `start { await $p.ready; $p.say(...); $p.close-stdin }` racing `await $p.start` was clean **120/120 under `-l 4` load**, matching rakudo. So there is nothing to pin a regression test to, and shipping the extension without one would be speculative. It is recorded on [#7943](https://github.com/tokuhirom/mutsu/issues/7943) instead, which I have narrowed to just that remaining consequence.

---

PR #7951, opened 2026-09-11
fix(parser): a paren-less zero-arg builtin call is a call, not a bareword

### LINKED ISSUE #7936: A paren-less zero-arg builtin call (`sleep;`, `my $x = exit;`) parses as a BareWord string instead of a call

tokuhirom:
Closes #7936

`sleep;` returned instantly instead of sleeping forever, and `my $x = exit;` bound the *string* `"exit"` and let execution carry on.

## Root cause

Neither builtin was at fault — the call never reached them. `sleep` is registered as a listop, but the listop parse needs at least one argument to follow it. With nothing after the identifier the parse fell through to the last resort at the end of `parse_identifier_call`, which turns an unrecognised word into `Expr::BareWord(name)`.

That fallback had grown an ad-hoc allowlist of names rescued from it one at a time (`await`/`slip`/`slurp`, then `callframe`/`caller`, then `return`/`return-rw`, each added after a bug report). As the ticket says, the allowlist growing one name per bug report *is* the defect: every core routine whose parameters are all optional was affected in paren-less zero-arg position.

## Fix

The three hardcoded arms are replaced by one table, `is_zero_arg_callable_builtin()` in `src/parser/primary/ident/predicates.rs`.

Membership is **measured against Rakudo rather than guessed**: a name belongs there iff `raku -e 'my $x = NAME;'` compiles. Rakudo rejects a bare argument-requiring routine at compile time — either with `Unsupported use of bare "say"` (the Perl 5 unary carve-out) or with `Calling elems() will never work with signature of the proto ($, *%)` — so a name that compiles there is one whose signature really does accept zero arguments. Sweeping mutsu's builtin and listop name tables through that oracle produced the 38 entries; everything else (`say`, `ord`, `elems`, `floor`, `map`, `set`, `bag`, `sleep-until`, …) keeps falling through to the bareword/`X::Obsolete` path unchanged.

`take-rw` and `parse-names` are deliberately excluded even though Rakudo accepts them, because mutsu has no zero-arg routine behind either: compiling them to a call would trade a wrong value for a worse error (`Undeclared routine` at parse time, and `Unknown function`).

`callframe`/`caller`/`return`/`return-rw` keep their own arm, now merged into one: they must compile to a call even when what follows is not a statement terminator, which the table's gate does not allow.

| code | before | after (= raku) |
| --- | --- | --- |
| `sleep;` | returned immediately | sleeps indefinitely |
| `my $x = exit;` | `Str "exit"`, execution continued | exits |
| `my $x = get;` | `Str "get"` | `Any` (reads `$*IN`) |
| `my $x = prompt;` | `Str "prompt"` | `Any` |
| `my $x = join;` | `Str "join"` | `Str ""` |
| `my $x = sum;` | `Str "sum"` | `0` |

Beyond the four names in the report this also corrects `chdir`, `cross`, `flat`, `hash`, `item`, `join`, `lines`, `max`, `min`, `minmax`, `note`, `repeated`, `repl`, `roundrobin`, `run`, `sleep-timer`, `sort`, `squish`, `succeed`, `sum`, `take`, `undefine`, `unique`, `val`, `warn`, `words` and `zip` in the same position.

## Second commit: two Proc::Async deadlocks the old parse was hiding

`roast/S17-procasync/kill.t` is whitelisted and went red, because it uses this shape twice:

```raku
start { await $p.ready; $p.kill }
await $p.start
```

Neither half worked under mutsu. The file only stayed green because its children are spawned as `$*EXECUTABLE, "-e", "sleep"` — and the bareword `sleep` made them exit in 20ms whatever the parent did. Making the children block turned both bugs into a hard deadlock, so they are fixed here:

- **`.ready` handed out a promise nobody held.** It built a fresh `SharedPromise` and stored it into the instance attributes for `.start` to keep — but a handler stores through *its own copy* of those attributes, so with `.ready` on one thread and `.start` on another, `.start` saw no `ready_promise` and kept nothing. The promise is now built by the constructor, so both share one whichever thread reaches which first.
- **`.kill` reported a running process as not started.** `started` and `pid` are written through that same per-thread copy, so from another thread `.kill` raised `X::Proc::Async::MustBeStarted` against a demonstrably running process. The ready promise is shared by reference and `.start` keeps it *with the pid*, so `.kill` now takes the pid from it and treats a kept ready promise as proof the process started.

This is targeted at `Proc::Async`, not general — instance-attribute mutations still do not propagate across threads, and the news entry records that `.write`/`.close-stdin` have the same latent problem.

## Tests

- `t/lang/parsing/bare-zero-arg-builtin-call.t` (36 tests) — the zero-arg call really happens; `sleep;` really blocks, measured in a child process so that only a *broken* `sleep` can fail the assertion; `exit` really exits; argument-requiring routines still report `Unsupported use of bare …`; calls with arguments are untouched.
- `t/concurrency/thread-lock/proc-async-ready-cross-thread-kill.t` (4 tests) — `.kill` before `.start` is still `X::Proc::Async::MustBeStarted`, `.ready` resolves with the pid, and `start { await $p.ready; $p.kill }` ends a child blocked in a bare `sleep`.

## Verification

- `cargo fmt --all`, `make lint` (all four configurations) — green
- `make test` — 3995 files, 42540 tests, **PASS**
- `make roast` — 1433 files, 219306 tests; the only red files are `S16-filehandles/filetest.t` (`Failed: 25`, tests 57-64/69-72/77-80/101…125) and `S32-io/IO-Socket-Async.t` (exit 124, planned 40 ran 17), which are exactly the container-environment failures documented in `docs/agent-environments.md` (`id -u` is `0` here; the network is sandboxed). `S17-procasync/kill.t` is green.

https://claude.ai/code/session_018oPv2zBwaykmuKMvY14mPz

---
_Generated by [Claude Code](https://claude.ai/code/session_018oPv2zBwaykmuKMvY14mPz)_

### DIFF
--- a/news/2026-09/paren-less-zero-arg-builtin-call.md
+++ b/news/2026-09/paren-less-zero-arg-builtin-call.md
@@ -72,7 +72,10 @@ position.
 
 This also removes the raciness the finding was hit through: a child spawned as
 `Proc::Async.new: $*EXECUTABLE, "-e", "sleep"` now blocks, so a parent's `.kill`
-no longer races the child's own 20ms exit.
+no longer races the child's own 20ms exit. That in turn exposed two real
+`Proc::Async` deadlocks in `roast/S17-procasync/kill.t`, which the 20ms exit had
+been hiding; they are fixed alongside and written up in
+`proc-async-ready-kill-cross-thread.md`.
 
 Pinned by `t/lang/parsing/bare-zero-arg-builtin-call.t`, which checks all three
 directions — the zero-arg call really happens, `sleep;` really blocks (measured
--- a/news/2026-09/paren-less-zero-arg-builtin-call.md
+++ b/news/2026-09/paren-less-zero-arg-builtin-call.md
@@ -1,0 +1,81 @@
+# A paren-less zero-arg builtin call is a call, not a bareword
+
+`sleep;` returned instantly instead of sleeping forever, and `my $x = exit;` bound
+the *string* `"exit"` and let execution carry on:
+
+```raku
+sub f() { my $x = exit; say "STILL RUNNING, x=$x.raku()"; }
+f();
+say "AND HERE TOO";
+```
+
+Rakudo prints nothing and exits 0. mutsu printed both lines.
+
+## Root cause
+
+Neither `builtin_sleep` nor `builtin_exit` was at fault — the call never reached
+them. `sleep` is registered as a listop, but the listop parse needs at least one
+argument to follow it. With nothing after the identifier the parse fell through
+to the last resort at the end of `parse_identifier_call`, which turns an
+unrecognised word into `Expr::BareWord(name)` — the string `"sleep"`.
+
+That fallback had grown an ad-hoc allowlist of names rescued from it one at a
+time (`await`/`slip`/`slurp`, then `callframe`/`caller`, then
+`return`/`return-rw`, each added after a bug report — the last because
+Text::CSV's `$error and return;` guard kept executing the rest of the sub). The
+allowlist growing one name per bug report was the actual defect: every core
+routine whose parameters are *all* optional was affected in paren-less zero-arg
+position, and `sleep`, `exit`, `get` and `prompt` simply had not been reported
+yet.
+
+## Fix
+
+The three hardcoded arms are replaced by one table,
+`is_zero_arg_callable_builtin()` in `src/parser/primary/ident/predicates.rs`.
+Membership is measured against Rakudo rather than guessed: a name belongs there
+iff `raku -e 'my $x = NAME;'` compiles. Rakudo rejects a bare
+argument-requiring routine at compile time — either with
+`Unsupported use of bare "say"` (the Perl 5 unary carve-out) or with
+`Calling elems() will never work with signature of the proto ($, *%)` — so a
+name that compiles there is one whose signature really does accept zero
+arguments. Sweeping mutsu's builtin and listop name tables through that oracle
+produced the 38 entries now in the table; everything else (`say`, `ord`,
+`elems`, `floor`, `map`, `set`, `bag`, `sleep-until`, ...) keeps falling through
+to the bareword/`X::Obsolete` path unchanged.
+
+Two names Rakudo accepts are deliberately excluded because mutsu has no zero-arg
+routine behind them, so compiling them to a call would trade a wrong value for a
+worse error: `take-rw` (reachable only with an argument, so a zero-arg call is an
+"Undeclared routine" *parse* error) and `parse-names` (implemented only as a
+method bypass, so a zero-arg call is "Unknown function").
+
+`callframe`/`caller`/`return`/`return-rw` keep their own arm, merged into one:
+they must compile to a call even when what follows is not a statement
+terminator, which the table's gate does not allow.
+
+## Result
+
+| code | before | after (= raku) |
+| --- | --- | --- |
+| `sleep;` | returned immediately | sleeps indefinitely |
+| `my $x = exit;` | `Str "exit"`, execution continued | exits |
+| `my $x = get;` | `Str "get"` | `Any` (reads `$*IN`) |
+| `my $x = prompt;` | `Str "prompt"` | `Any` |
+| `my $x = join;` | `Str "join"` | `Str ""` |
+| `my $x = sum;` | `Str "sum"` | `0` |
+
+Beyond the four names in the report this also corrects `cross`, `flat`, `hash`,
+`item`, `join`, `lines`, `max`, `min`, `minmax`, `note`, `repeated`, `repl`,
+`roundrobin`, `run`, `sleep-timer`, `sort`, `squish`, `succeed`, `sum`, `take`,
+`undefine`, `unique`, `val`, `warn`, `words`, `zip` and `chdir` in the same
+position.
+
+This also removes the raciness the finding was hit through: a child spawned as
+`Proc::Async.new: $*EXECUTABLE, "-e", "sleep"` now blocks, so a parent's `.kill`
+no longer races the child's own 20ms exit.
+
+Pinned by `t/lang/parsing/bare-zero-arg-builtin-call.t`, which checks all three
+directions — the zero-arg call really happens, `sleep;` really blocks (measured
+in a child process, where only a *broken* `sleep` can fail the assertion),
+`exit` really exits, argument-requiring routines still report
+`Unsupported use of bare ...`, and calls with arguments are untouched.
--- a/news/2026-09/proc-async-ready-kill-cross-thread.md
+++ b/news/2026-09/proc-async-ready-kill-cross-thread.md
@@ -1,0 +1,55 @@
+# Proc::Async `.ready` and `.kill` work from a thread that did not call `.start`
+
+`roast/S17-procasync/kill.t` uses this shape twice:
+
+```raku
+start { await $p.ready; $p.kill }
+await $p.start
+```
+
+Under mutsu neither half worked, and the file only stayed green because its
+children were spawned as `$*EXECUTABLE, "-e", "sleep"` — and a bare `sleep` fell
+through the parser as a bareword, so the child exited in 20ms on its own,
+whatever the parent did. Fixing that parse (see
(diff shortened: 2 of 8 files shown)


tokuhirom:
CI status on `fd3f1df` — the substantive jobs are green and the two red ones are an `archive.ubuntu.com` outage, not this diff.

**Green on this head:** `test` (the release build + full `prove t/` + `make roast`), `gc-stress`, `lint-configs`, `changes`, `label`. `miri` skipped.

**Red, both in `apt` before a single line of mutsu runs:**

- `jit-stress` (attempt 1) died in `apt-get install -y perl mold` with
  `E: Failed to fetch .../mold_2.30.0%2bdfsg-1build1_amd64.deb  Connection failed [IP: 91.189.91.83 80]`, exit 100. No build, no test body; the job's own `tmp/jit-stress-t.log` artifact was never created.
- `wasm-e2e` (attempt 2) hit its 15-minute job timeout still inside `playwright install --with-deps chromium`'s apt stage, downloading fonts from the same mirror at roughly 60 KB/s (`fonts-wqy-zenhei`, 7.4 MB, over two minutes and unfinished). `##[error]The operation was canceled` with `npm exec playwright install` still an orphan process.

Neither job reaches any code this PR touches, and there is no fix to port — the failing dependency is an Ubuntu package mirror. I used the one sanctioned re-run (`rerun_failed_jobs` at 07:59 UTC); `jit-stress` got past `apt` on that attempt and is still running, `wasm-e2e` timed out in `apt` again. That re-run is now spent, so I am not re-running further and I am not touching the `wasm-e2e` timeout to work around it.

Auto-merge (merge method) is enabled, so the PR will land on its own once these two go green. I am keeping it watched until it merges or closes.

---

tokuhirom:
Correcting my previous comment on `jit-stress`, and pushing a small change.

**What actually happened on the re-run.** `jit-stress` attempt 2 got *past* `apt` and into `prove t/`, then hit its own `timeout-minutes: 30` — `prove` and its in-flight child were still listed as orphans at cleanup. So the second attempt was a job-budget timeout, not the `apt` hard failure I described. The underlying cause is the same degraded mirror: attempt 1 spent six minutes in `apt-get install -y perl mold` before failing outright, and that job has to fit `apt` + a debug build + a serial `prove -r t/` + a release build + roast into 30 minutes. `wasm-e2e` timed out in `apt` the same way. `test`, `gc-stress` and `lint-configs` are green on this head.

**One thing I checked and ruled out.** Making `sleep` genuinely infinite means a `Proc::Async` child that is never killed now blocks forever instead of exiting in 20ms, so I checked for leaked children. On this box there were seven stranded `mutsu -e sleep` processes — but they date from *before* the `Proc::Async` commit, when `.kill` from a non-starting thread raised `MustBeStarted` and never landed. On the current head, `roast/S17-procasync/kill.t` and both new tests leave **zero** stranded processes, and every signal (HUP/TERM/INT/KILL) terminates a blocked `mutsu` correctly. The stranded ones also sit at 0.0% CPU — the infinite `sleep` is a 100ms poll, not a spin — so they were not slowing anything either way.

**Pushed:** `330e8a4f` trims the bare-`sleep` test's child wait from 5s to 2s. That wait is a fixed cost in the *serial* `prove t/` of both 30-minute jobs, and 2s is still a hundredfold margin over the ~20ms a bareword `sleep` would take, so the assertion keeps its one-directional shape (a slow machine can only make it pass). I am not claiming three seconds fixes a thirty-minute timeout — the blocker is the mirror — but it is my own cost in the tightest jobs and worth giving back. The push also gets a fresh CI run, which the spent re-run could no longer provide.

---

PR #7957, opened 2026-09-11
fix: scope `sub EXPORT` per compunit so nested EXPORT modules can load

### LINKED ISSUE #7947: Two modules in one load chain that each declare `sub EXPORT` collide on the single `GLOBAL::EXPORT` key
Found by P5 triage of the first full-corpus ecosystem sweep (#7785). **10 distributions** fail to load on this, the largest single root cause in the `blocked_load` bucket.

## Repro

Four lines, two files. `Inner.rakumod`:

```raku
sub inner-fn() { "inner" }
sub EXPORT() { Map.new: ('&inner-fn' => &inner-fn) }
```

`Outer.rakumod` — a module that both `use`s Inner **and** has its own `sub EXPORT`:

```raku
use Inner;
sub EXPORT() { Map.new: ('&outer-fn' => &inner-fn) }
```

```
$ mutsu -I lib -e 'use Outer; say outer-fn();'
Redeclaration of routine 'EXPORT'. Did you mean to declare a multi-sub?
  in block <unit> at lib/Inner.rakumod line 1

$ raku -I lib -e 'use Outer; say outer-fn();'
inner
```

Controls, all measured:

| shape | mutsu | raku |
|---|---|---|
| one module with `sub EXPORT` | ok | ok |
| **two modules in a chain, each with `sub EXPORT`** | **X::Redeclaration** | ok |
| two `sub EXPORT` modules `use`d side by side from the script | ok | ok |
| only the inner module has `sub EXPORT` | ok | ok |
| `my sub EXPORT` in the outer module | **X::Redeclaration** | ok |
| three deep, two of them with `sub EXPORT` | **X::Redeclaration** | ok |
| two same-named *ordinary* private subs across a nested load | ok | ok |

So it is specific to `EXPORT`, it does not need the two declarations to be in the same file or the same scope kind, and it is not the general nested-load symbol handling — that works.

## Root cause

`sub EXPORT` is not scoped to its compunit. `src/runtime/runtime_module_export_sub.rs:64` says so outright:

```rust
// The module body runs under GLOBAL, so `sub EXPORT` registers as
// `GLOBAL::EXPORT`. Only participate when it is actually present.
let Some(def) = self.resolve_function("EXPORT") else {
```

Every module's hook therefore competes for **one** registry key. Sub declarations are hoisted, so entering Outer's body registers `GLOBAL::EXPORT` before `use Inner` runs; Inner's hoisted declaration then hits the redeclaration check in `src/runtime/registration_sub.rs:1251` (`has_single && !existing_is_stub && !shadows_outer_eval_single`) and the load dies.

Side-by-side `use` works because the first module's hook is consumed and recorded in `module_export_defs` before the second module loads. Nesting is what makes the two live at once.

In Raku `EXPORT` is per-compunit: each file may have exactly one, and two files' hooks never see each other. That is the invariant to restore.

## Affected files

- `src/runtime/runtime_module_export_sub.rs` — `apply_module_export`, the bare-name `resolve_function("EXPORT")` lookup
- `src/runtime/registration_sub.rs` — the redeclaration check that fires (around line 1251)
- `docs/adr/0087-*` — the `sub EXPORT` approximation's design; a per-compunit key is within its scope but worth a line there
- `src/parser/stmt/simple/module_exports/export_hook.rs` — the parse-time scan already excludes `EXPORT` from a module's collected routine set, so it is not implicated, but it is the other half of the mechanism

Keying the hook by compunit (the module name is already on `module_load_stack`) rather than under `GLOBAL` is the obvious shape. **Two paths must not regress**, both documented in comments right there: the inherited/Slangify `&EXPORT` handoff via `pending_inner_export_subs`, and the env discipline that keeps a bareword package (NativeLibs' `Map.new('NativeCall' => NativeCall, ...)`) bound while the hook runs (#7806). If that turns out to need a decision rather than a rename, re-triage this to `todo:deep` — the repro stands either way.

## Why it matters

Ten distributions in the ledger fail to load, every one of them on this:

`Data::Generators`, `Identity::Utils`, `List::AllUtils`, `List::SomeUtils`, `PURL`, `SBOM::CycloneDX`, `span`, `Test::Async`, `User::Timezone`, `UserTimezone`

Three were opened and confirmed to have the shape by source inspection:

- `span` — `use snip:ver<0.0.2+>:auth<zef:lizmat>;` then `sub EXPORT() { Map.new: ('&span' => &snip) }`
- `Identity::Utils` — `use String::Utils` on line 1, `my sub EXPORT(*@names)` on line 469
- `List::SomeUtils` — `use List::MoreUtils` on line 4, `sub EXPORT(*@args, *%_)` on line 15

This is the dominant lizmat idiom: a small distribution that re-exports a dependency under a different name, where both ends compute their exports in `sub EXPORT`. `Test::Async` is the one that hurts most beyond the count — it is a test framework, so every distribution that tests with it inherits the failure.

Ledger records carry the per-distribution detail (`ecosystem/dists/<S>/<name>~<digest>.json`, `status: blocked_load`); method in [docs/ecosystem-parity.md](https://github.com/tokuhirom/mutsu/blob/main/docs/ecosystem-parity.md).

Refs #7785

tokuhirom:
A module that both `use`d another `sub EXPORT` module **and** declared its own hook failed to load at all:

```
$ mutsu -I lib -e 'use Outer; say outer-fn();'
Redeclaration of routine 'EXPORT'. Did you mean to declare a multi-sub?
  in block <unit> at lib/Inner.rakumod line 1
```

## Root cause

Raku scopes `sub EXPORT` to its compunit: each file may declare exactly one, and two files' hooks never see each other. mutsu did not hold that invariant on the run-time side. A module body runs under `GLOBAL`, so every module's `sub EXPORT` registered under the single `GLOBAL::EXPORT` key — and sub declarations are hoisted, so entering the outer module's body registered `GLOBAL::EXPORT` before its `use` of the inner module ran. The inner module's own hoisted declaration then hit the redeclaration check in `registration_sub.rs` and killed the load.

Side-by-side `use` worked because the first hook is consumed and recorded in `module_export_defs` before the second module loads. Nesting is what made the two live at once, and it needed neither the same file nor the same scope kind: a `my sub EXPORT` in the outer module, or a three-deep chain, collided the same way.

## The fix

`load_module` now hides whatever `EXPORT` routine an enclosing compunit registered for the duration of a nested load (`Interpreter::hide_export_routines`) and restores it once `apply_module_export` has called and dropped that module's own hook — on both the success and the error path, so an aborted load cannot leave the enclosing hook hidden.

The mechanism is the one `hide_toplevel_global_routines` already uses for ordinary package-less top-level routines, but it deliberately **cannot** be folded into it, and that is the whole subtlety: that restore runs *before* `apply_module_export`, so including `EXPORT` in it put an outer module's stale hook back over the inner module's fresh one before it was ever read — which is exactly why `is_toplevel_global_routine_key` excludes `EXPORT` today. The new pair brackets `apply_module_export` instead.

Neither of the two paths the issue flags as must-not-regress moves: the inherited `&EXPORT` handoff (the Slangify pattern) is keyed by module in `pending_inner_export_subs`, and the env discipline that keeps a bareword package bound while the hook runs (#7806) is untouched. `t/modules/import-export/sub-export.t`, `module-inner-export-sub.t`, `module-export-sub-sees-module-scope-not-caller.t` and all 408 files under `t/modules` + `t/nativecall` pass.

## Measured against `raku`

Every row of the issue's control table, re-measured after the fix:

| shape | mutsu before | mutsu after | raku |
|---|---|---|---|
| one module with `sub EXPORT` | ok | ok | ok |
| two modules in a chain, each with `sub EXPORT` | **X::Redeclaration** | ok | ok |
| two `sub EXPORT` modules `use`d side by side | ok | ok | ok |
| only the inner module has `sub EXPORT` | ok | ok | ok |
| `my sub EXPORT` in the outer module | **X::Redeclaration** | ok | ok |
| three deep, a hook at every level | **X::Redeclaration** | ok | ok |

## Why it matters

Ten distributions in the ecosystem ledger were `blocked_load` on this one root cause — the largest single entry in that bucket: `Data::Generators`, `Identity::Utils`, `List::AllUtils`, `List::SomeUtils`, `PURL`, `SBOM::CycloneDX`, `span`, `Test::Async`, `User::Timezone`, `UserTimezone`. `Test::Async` reaches furthest: it is a test framework, so every distribution that tests with it inherited the failure. The ledger records are left for the next sweep to regenerate.

## Tests

New pin `t/modules/import-export/nested-export-sub-scoping.t` (6 assertions) with fixtures under `t/lib/ChainExport*.rakumod`: the nested chain, the `my sub EXPORT` variant, three levels deep, an outer hook that consumes the end user's `use` arguments, and the side-by-side shape that already worked. All six were run under `raku` first and match.

ADR-0087 gains a Consequences note: the parse-time scan already worked per module file, and the run-time side now agrees on that unit.

## Verification

- `cargo fmt --all`, `make lint` (all four configurations) — clean
- `make test` — **PASS**, 3997 files / 42516 tests
- `make roast` — the three container-only failures and nothing else, by name and by subtest range, per `docs/agent-environments.md`: `6.c/S32-io/file-tests.t` (6-8, 10), `S16-filehandles/filetest.t` (57-64, 69-72, 77-80, 101/103/105/107/109/111/117/121/125), `S32-io/IO-Socket-Async.t` (exit 124, planned 40 ran 17). `id -u` is 0 here.

Closes #7947

https://claude.ai/code/session_01Pv3MLDnwMiNLsDSpqaFizx

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pv3MLDnwMiNLsDSpqaFizx)_

### DIFF
--- a/docs/adr/0087-runtime-export-hook-parse-time-approximation.md
+++ b/docs/adr/0087-runtime-export-hook-parse-time-approximation.md
@@ -132,3 +132,14 @@ file — and every other call shape benefits too.
 - Pinned by `t/modules/import-export/runtime-export-listop-parse.t`, which
   covers both directions: the listop call must parse, and a unit-scope routine
   the hook deliberately withholds must still be unresolvable at run time.
+- The scan detects the hook **per module file**, which is the right unit: the
+  run-time hook is per-compunit too. That invariant was not held on the run-time
+  side at first — a module body runs under `GLOBAL`, so every module's
+  `sub EXPORT` registered under the one `GLOBAL::EXPORT` key, and two modules in
+  a single load chain that each declared one collided with `X::Redeclaration`
+  ([#7947](https://github.com/tokuhirom/mutsu/issues/7947)). `load_module` now
+  hides the enclosing compunit's hook for the duration of a nested load
+  (`Interpreter::hide_export_routines`, restored only *after*
+  `apply_module_export` has consumed the nested module's own), so the two sides
+  agree on the unit. Nothing in the approximation above changes: the scan
+  already read one file at a time.
--- a/news/2026-09/nested-sub-export-per-compunit.md
+++ b/news/2026-09/nested-sub-export-per-compunit.md
@@ -1,0 +1,64 @@
+# `sub EXPORT` is scoped per compunit again
+
+A module that both `use`s another `sub EXPORT` module **and** declares its own
+hook failed to load at all:
+
+```
+$ mutsu -I lib -e 'use Outer; say outer-fn();'
+Redeclaration of routine 'EXPORT'. Did you mean to declare a multi-sub?
+  in block <unit> at lib/Inner.rakumod line 1
+```
+
+Raku scopes `sub EXPORT` to its compunit: each file may declare exactly one, and
+two files' hooks never see each other. mutsu did not hold that invariant on the
+run-time side. A module body runs under `GLOBAL`, so every module's `sub EXPORT`
+registered under the single `GLOBAL::EXPORT` key, and sub declarations are
+hoisted — entering the outer module's body registered `GLOBAL::EXPORT` before
+its `use` of the inner module ran, so the inner module's own hoisted declaration
+hit the redeclaration check in `registration_sub.rs` and killed the load.
+
+Two hooks side by side worked, because the first is consumed and recorded in
+`module_export_defs` before the second module loads. Nesting is what made the
+two live at once, and it did not need them to share a file or a scope kind: a
+`my sub EXPORT` in the outer module, or a three-deep chain, collided the same
+way.
+
+## The fix
+
+`load_module` now hides whatever `EXPORT` routine an enclosing compunit
+registered for the duration of a nested load
+(`Interpreter::hide_export_routines`), so the nested module's own declaration
+lands on a clean key, and restores it once `apply_module_export` has called and
+dropped that module's hook.
+
+The mechanism is the one `hide_toplevel_global_routines` already uses for
+ordinary package-less top-level routines — but it deliberately could **not** be
+folded into it, and the reason is the whole subtlety. That restore runs *before*
+`apply_module_export`, so including `EXPORT` in it put an outer module's stale
+hook back over the inner module's fresh one before it was ever read, which is
+exactly why `is_toplevel_global_routine_key` excludes `EXPORT` today. The new
+pair brackets `apply_module_export` instead, on both the success and the
+error path, so an aborted load cannot leave the enclosing hook hidden.
+
+Neither of the two paths the old comments warn about moves: the inherited
+`&EXPORT` handoff (the Slangify pattern) is keyed by module in
+`pending_inner_export_subs`, and the env discipline that keeps a bareword
+package bound while the hook runs is untouched.
+
+## Why it mattered
+
+Ten distributions in the ecosystem ledger were `blocked_load` on this one root
+cause — the largest single entry in that bucket, found by triage of the first
+full-corpus sweep: `Data::Generators`, `Identity::Utils`, `List::AllUtils`,
+`List::SomeUtils`, `PURL`, `SBOM::CycloneDX`, `span`, `Test::Async`,
+`User::Timezone`, `UserTimezone`. It is the dominant lizmat idiom: a small
+distribution that re-exports a dependency under a different name, with both ends
+computing their exports in `sub EXPORT`. `Test::Async` is the one that reaches
+furthest — it is a test framework, so every distribution that tests with it
+inherited the failure.
+
+Pinned by `t/modules/import-export/nested-export-sub-scoping.t`, which
+covers the nested chain, the `my sub EXPORT` variant, three levels deep, an
+outer hook that consumes the end user's `use` arguments, and the side-by-side
+shape that already worked. All six assertions were measured against `raku`
+first.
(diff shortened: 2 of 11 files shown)


tokuhirom:
`jit-stress` came back `cancelled` on `bf7148f`, and it is not this PR's. It was killed by its own `timeout-minutes: 30` after the **`Install prove + mold` step took 10m19s** (08:09:25 → 08:19:44) instead of its usual ~20s — the azure apt mirror hang that `ci.yml`'s own "Use canonical apt mirror" comment documents (PRs #6684 #6691). That stall took a third of the job's budget.

The TAP run inside that job **did finish, and passed in full**. From the uploaded `jit-stress-logs` artifact:

```
All tests successful.
Files=3997, Tests=42516, 1102 wallclock secs (...)
Result: PASS
```

Zero `not ok` lines in 4000 lines of output. What never got to start was the job's *later* steps — the release build and `make roast` with JIT on — because only ~1 minute of the 30 was left by then.

Everything else on this head is green: `test` (release `cargo test` + `prove t/` + `make roast`), `gc-stress` (debug `prove t/`, serial), `lint-configs`, `wasm-e2e`. The diff touches `src/runtime/` module-load bookkeeping and adds one test file; nothing in it can lengthen an apt install.

Re-running the failed job now — this is the one re-run.

---

Issue #7856, opened 2026-09-10
`Symbol::flags()` costs 33 Ir per call for what should be an array load


tokuhirom:
Found while profiling the closure-capture path for #7557 (2026-09-10). Small, self-contained, and measured — but see "Interaction with ADR-0086" before picking it up, because the one caller that makes it visible today is a path ADR-0086 wants to shrink.

## Measurement

`Symbol::flags()` (`src/symbol.rs`) is the memoized flags word #7624 introduced so `capture_closure_env`'s filter could answer its per-key string questions (`PLAIN_USER_LEXICAL`, `ATTR_TWIGIL_ENV_KEY`, `CALLABLE_ID_META`, ...) with one lookup instead of re-scanning the symbol's bytes. It is a hit in essentially every call, and a hit still costs **33 Ir**:

```rust
pub(crate) fn flags(self) -> u16 {
    let idx = self.0 as usize;
    if let Some(f) = FLAG_CACHE.with(|c| c.borrow().get(idx).copied())
        && f & flags::COMPUTED != 0
    { return f; }
    ...
}
```

per call that is a thread-local access (with its lazy-init check), a `RefCell::borrow` (two writes to the borrow counter plus the guard's drop), a bounds-checked `Vec` index and a copy — for a value that is a pure function of a `u32`.

Callgrind, release, on `main` at 8bb9132:

| program | `Symbol::flags` calls | Ir | share |
|---|---|---|---|
| `sub make($n) { my $c = { $n + 1 } }` × 20000 | 620,000 | 20,467,946 | **1.74%** |
| `benchmarks/bench-ctor.raku` | 185,000 | 6,108,903 | **0.46%** |

In both, **100% of the calls come from `capture_closure_env`'s filter closure** (31 keys per creation in the first, 37 in the second).

## Why it should be cheaper

Interned ids are **global and append-only** — the `symbol::wk` module already relies on exactly that to cache well-known symbols in a `OnceLock` "valid for the life of the process and across threads". So the flags table does not have to be thread-local at all, and even as a thread-local it does not need `RefCell`: the value is write-once per id and a stale-but-`COMPUTED`-clear read is already handled by the recompute path. A plain array load is ~3 Ir.

Shapes worth costing, cheapest first:

- Drop the `RefCell` in favour of a `Cell`-based store (the borrow counter is the bulk of the 33 Ir), keeping the `Vec` growth behind the existing miss path.
[...]

Issue #7565, opened 2026-09-08
`use Test` taxes every hot loop in the file that loads it


tokuhirom:
Loading the vendored upstream `Test` module — which is what every `use Test`
does once PR #7523 lands (`todo/deep/vendor-real-test-module-flip.md`)
— makes unrelated hot loops in the *same file* several times slower. The module
is not running; merely having loaded it is enough.

The extreme case in the `t/` suite, measured on a 4-core box, debug build:

| | wall |
| --- | --- |
| `t/rebound-return-hot-loop.t`, native `Test` provider | 4.84 s |
| same file, vendored `Test` | **13.18 s** |

That file has 7 assertions and a 20 000-iteration loop, so essentially none of
the 8.35 s is assertion cost. It is the single largest per-file regression in
the whole suite — 2% of the switch's total `t/` cost sits in this one file,
against a median per-file delta of 94 ms.

## Two linear costs, both already named in earlier work

Reduced to a file with no `use Test` at all (debug, median of three):

| variant | wall |
| --- | --- |
| the loop alone | 4.48 s |
| `+ 60 unused mainline `our` variables` | 7.26 s |
| `+ 60 unused mainline subs` | 6.59 s |
| `+ a bare `EVAL "1"`` | 5.10 s |
| `+ use Test` (vendored) | 12.77 s |

1. **Cost linear in mainline env size.** `use Test` imports ~60 names into the
   caller's env. Sixty *unused* `our` variables reproduce most of the effect on
   their own, so nothing about `Test` specifically is at fault — any module with
   a wide export list taxes every loop iteration in its importer. This is the
   same linear-in-env-size cost the vendoring campaign's fourth optimization
   pass measured at ~0.98 ns per env entry per assertion and filed as
   `todo/perf/method-dispatch-flattens-the-env-on-every-call.md`; that ticket
   framed it as a *method dispatch* cost, and this shows it is broader.
2. **The reflective latch.** Loading a module that `EVAL`s anywhere (upstream
   `Test`'s `throws-like` does) latches
   `opcode::reflective_name_access_possible()`, which is process-global and
   monotonic. That disables `SetLocal`'s `skip_env_write`, so every store writes
   the env mirror as well as its slot for the rest of the process. A bare
   `EVAL "1"` alone costs +14%.

The two do not explain the whole 2.85x on their own (60 pads plus a latch
predicts ~8.6 s against a measured 12.8 s), so a third term is unaccounted for.
Finding it is the first job here.

[...]

PR #7964, opened 2026-09-11
perf(vm): closure capture keeps `__mutsu_type::` metadata only for a captured subject

tokuhirom:
Item 1 of the "what is left" list #7707 left on #7565.

## The problem

The filtered closure capture kept **every** `__mutsu_type::<name>` typed-lexical
metadata key it could see. It had to, under the rule it was written to: the filter
keeps a key that is a free variable of the closure **or** that is not a plain user
lexical, and a `__mutsu_*` metadata key is never a plain user lexical. So a closure
inherited the type-constraint metadata of every typed lexical in its creating
scope, including the ones its body cannot possibly name — 11 of the ~35 entries a
closure still captured after a bare `use Test`, scaling with the importing scope
exactly like the `__mutsu_callable_id::` markers #7707 removed.

## The fix

`__mutsu_type::<name>` is *shadow metadata*: nothing in the language can observe it
except through a read or a write of `<name>` itself, so it belongs in a capture on
exactly the same terms as its subject. That is the same reasoning
`CompiledCode::is_callee_local_sym` already applies to these keys on the
scoped-overlay return merge. The filter now unwraps the key and runs the rest of the
predicate on the subject.

Deciding it *that* way, rather than by the tempting shortcut "keep it only for a free
variable", is what keeps a **typed dynamic** correct — the hazard #7707 flagged.
`my Int $*dyn` is captured because its env key is a system name, never because it is
a free variable, so a free-variable test would have let `$*dyn = 'str'` through inside
any closure. Running the real predicate on the subject handles dynamics, `self`,
match captures and the rest for free, because those are precisely the names the
predicate already keeps.

`Symbol::type_meta_subject` memoizes the unwrapping in the chunked, lock-free table
established for `Symbol::flags` (#7856) — the mapping is a pure function of an
immutable string and symbol ids are append-only. A thread-local `RefCell<FxHashMap>`
was measured first and cost ~41 Ir per hit against one 32-bit load. Only ids that are
actually metadata keys reach the store, so a program with no typed lexical allocates
no chunk. `CompiledCode::is_callee_local_sym`, which did the same unwrapping by hand
per metadata key per named call, now shares it.

## Measured

Callgrind, warm runs, `MUTSU_JIT=off MUTSU_GC=off`, release, slope between 6 000 and
14 000 iterations of the ticket's loop — the protocol #7707 established. A repeated
baseline run reproduced to within 95 instructions on a ~480M-instruction run, so the
slopes are good to well under 1 Ir per iteration.

| | instructions per iteration |
| --- | --- |
| the loop with no `use Test` (the floor) | 77 738 → 77 917 |
| `+ use Test`, before | 106 616 |
| `+ use Test`, after | **100 333** |

Import tax **28 878 → 22 416 per iteration (-22%)**; the loop as a whole **-5.9%**.
The mechanism is visible directly in the profile: `Env::insert_sym` — the
per-closure-call capture merge — falls 48.0M → 38.3M (-20%).

The floor's +179 (+0.23%) is optimizer drift, not added work: an experiment that
removed the new branch from the untyped path *entirely* (a `const` generic on
`env_type_constraint_seen`, so the metadata arm is not even compiled for a program
with no typed lexical) still measured +180. Two further shapes of the same logic —
the predicate extracted to an `#[inline]` function, and the two "drop this key" flag
tests fused into one mask — measured between +0.1% and +0.7% *worse* on the
`use Test` path. Worth knowing before rearranging this filter for readability.

## Behaviour

One deliberate change beyond the capture width: the metadata of a name a closure
*shadows* with its own declaration is now dropped along with the shadowed name,
instead of being inherited by a frame that has no such variable. That is the same
cross-frame constraint leak `set_var_type_constraint_routine_scoped` was scoped to
prevent (`news/2026-09/type-constraint-global-side-table-retired.md`); the capture
was still carrying one case of it.

## Testing

- `t/routines/closure/closure-capture-type-constraint.t` — 12 new assertions over the
  shapes where the subject *is* captured and the constraint must still bite (typed
  scalar, typed dynamic, typed array element, typed hash value, a subject reached only
  through a nested closure, an escaping closure over a routine-local typed lexical, a
  typed `is copy` parameter), plus the two shapes where a closure-local redeclaration
  must *not* inherit an outer or creating-routine constraint. All twelve agree with
  `raku`.
- Two `#[test]`s pinning `Symbol::type_meta_subject`'s round trip and its
  cross-thread visibility.
- `make test`: PASS, 4000 files / 42558 tests.
- `make lint`: green in all four configurations.
- `make roast`: red only on `roast/6.c/S32-io/file-tests.t` (4) and
  `roast/S16-filehandles/filetest.t` (25) — the two `uid 0` `chmod` files
  [docs/agent-environments.md](https://github.com/tokuhirom/mutsu/blob/main/docs/agent-environments.md)
  records as container-only, with exactly the documented failure counts.

Refs #7565 (left open: the tax is reduced, not eliminated — items 2 and 3 of that
list remain).

https://claude.ai/code/session_017xhWSUqaSHmPW9or22tXfi

---
_Generated by [Claude Code](https://claude.ai/code/session_017xhWSUqaSHmPW9or22tXfi)_

### DIFF
--- a/news/2026-09/closure-capture-type-metadata-follows-its-subject.md
+++ b/news/2026-09/closure-capture-type-metadata-follows-its-subject.md
@@ -1,0 +1,97 @@
+# A closure's captured `__mutsu_type::` metadata now follows its subject
+
+The filtered closure capture (`Interpreter::capture_closure_env`) kept every
+`__mutsu_type::<name>` typed-lexical metadata key it could see, unconditionally.
+It had to, under the rule it was written to: the filter keeps a key when it is a
+free variable of the closure *or* when it is not a plain user lexical, and a
+`__mutsu_*` metadata key is never a plain user lexical. So a closure created
+anywhere in a scope inherited the type-constraint metadata of every typed lexical
+in that scope, including the ones its body cannot possibly name.
+
+That is the third of the three linear-in-import-size terms
+[#7565](https://github.com/tokuhirom/mutsu/issues/7565) decomposed. After the
+first two were fixed (#7656's END-phaser refresh, #7707's per-chunk reflective
+latch and `__mutsu_callable_id::` filter), 11 of the ~35 entries a closure still
+captured after a bare `use Test` were these keys, and they scale with the
+importing scope exactly like the routine-registration markers #7707 removed.
+
+## The rule
+
+`__mutsu_type::<name>` is *shadow metadata*: it records what `<name>` is
+constrained to, and nothing in the language can observe it except through a read
+or a write of `<name>` itself. So it belongs in a capture on exactly the same
+terms as its subject does — which is the same reasoning
+`CompiledCode::is_callee_local_sym` already applies to these keys on the
+scoped-overlay return merge. The filter now unwraps the key and decides it by
+running the rest of the predicate on the subject.
+
+Deciding it *that* way, rather than by the tempting shortcut "keep it only for a
+free variable", is what keeps a **typed dynamic** correct. `my Int $*dyn` is
+captured because its env key is a system name, never because it is a free
+variable; a free-variable test would have dropped its constraint and let
+`$*dyn = 'str'` through inside any closure. Running the real predicate on the
+subject handles dynamics, `self`, match captures and the rest for free, because
+those are precisely the names the predicate already keeps.
+
+A side effect worth noting: the metadata of a name the closure *shadows* with its
+own declaration is now dropped along with the shadowed name, instead of being
+inherited by a frame that has no such variable. That is the same cross-frame
+constraint leak `set_var_type_constraint_routine_scoped` was scoped to prevent
+(`news/2026-09/type-constraint-global-side-table-retired.md`); the capture was
+still carrying one case of it.
+
+## The unwrapping is memoized
+
+Reversing `__mutsu_type::<name>` to `<name>` meant resolving the symbol,
+re-scanning the prefix and re-interning the suffix — a string hash — every time.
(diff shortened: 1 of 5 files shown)


tokuhirom:
`jit-stress` came back **cancelled** on the first run (`e7bbf35`), with every test step green. That is the job's 30-minute `timeout-minutes` firing, not a test failure: the job started 09:17:57 and was killed at 09:48:14, with the serial TAP step alone taking 09:25:08 → 09:43:19 and the roast step cut off 36 seconds in (the log ends on `Terminate orphan process: (prove)`).

This is exactly the failure #7963 diagnosed and fixed — "gc-stress and jit-stress were habitually using ~90% of their 30-minute budget and reporting *cancelled* with every test step green, so a single slow apt step cancelled an innocent PR". That fix landed on `main` at 09:22, five minutes after this PR's run started, so this branch was still on the pre-fix `ci.yml` (base `38c1bbd`, which predates it).

Rebased onto `ec4c940` to pick it up rather than spending the re-run on the same 30-minute budget. No change to the diff — it is byte-for-byte the same commit, re-parented.

Re-validated on the new base before pushing: `make test` PASS (4006 files / 42642 tests), `make roast` red only on `roast/6.c/S32-io/file-tests.t` (4) and `roast/S16-filehandles/filetest.t` (25), the two `uid 0` `chmod` files [docs/agent-environments.md](https://github.com/tokuhirom/mutsu/blob/main/docs/agent-environments.md) records as container-only, with exactly the documented failure counts.

---

Issue #8096, opened 2026-09-12
A rule-local `:my $*VAR` declaration is established parse-wide and leaks past `.parse` into the caller's dynamic scope


tokuhirom:
## Summary

`:my $*VAR = …;` inside a grammar rule body declares a dynamic variable whose
scope, in rakudo, is the **dynamic scope of that rule's own match**. mutsu
instead evaluates every such declaration it can find anywhere in the grammar
once, into the interpreter's env, before the parse starts — so:

1. an action method sees the value even when the declaring rule never matched;
2. the binding survives `.parse` returning, and is visible to the caller.

## Repro

```raku
grammar H {
    proto token sigil { * }
    token sigil:sym<dollar> { :my $*LAST = 'dollar'; '$' }
    token sigil:sym<at>     { '@' }
    token TOP { <sigil> }
}
class B { method TOP($/) { make $*LAST // 'none' } }
say "dollar-parse: ", H.parse('$', :actions(B)).made;
say "at-parse:     ", H.parse('@', :actions(B)).made;
say "after-parse:  ", ($*LAST // 'unset');
```

rakudo 2026.07:

```
dollar-parse: none
at-parse:     none
after-parse:  unset
```

mutsu (`main`, 9a788ca10):

```
dollar-parse: dollar
at-parse:     dollar
after-parse:  dollar
```

The `at-parse` line is the clearest symptom: `sigil:sym<dollar>` never matched,
and its declaration is still in scope. `after-parse` shows the same binding
outliving the parse entirely.

(rakudo answers `none` for the first line too — by the time `TOP`'s action runs,
the inner token's dynamic scope has already been left. A test that wants to
observe the variable has to do it from an action method of the *declaring* rule,
or from a code block inside it.)

## Where it comes from

`Interpreter::establish_grammar_dynamic_vars`
(`src/runtime/methods_grammar.rs`) walks every rule pattern of the grammar's
package and its MRO ancestors, extracts every `:my $*/@*/%*NAME = INIT;`
declaration, evaluates each one into `self.env`, and returns the prior values so
the caller can restore them when the parse ends. Two things follow:

- **The union of all rules' declarations is in scope for the whole parse**, not
  each rule's own during its own match. The per-rule refinement that exists
  (`grammar_rule_dynvar_decls`, consumed by the reduce walk) gives a *declaring
  rule's match* its own binding, but does not stop the parse-wide slot from
  existing and being visible to everything else.
[...]

Issue #8095, opened 2026-09-12
The parser re-parses a module's source to collect its exports, ignoring the precompiled AST already on disk (209 M Ir per `use`)


tokuhirom:
## Summary

A `use Foo;` makes mutsu parse `Foo`'s source **twice**: once in the parser, to
learn what the module exports, and once at run time to actually load it. The
run-time half consults the on-disk precompilation cache and skips its parse on a
hit (`Interpreter::parse_module_source` -> `precomp::load_cached_unit`). The
parser half has no such cache and re-parses from source on every process.

It is the larger of the two.

## Measurement

Measured with callgrind (deterministic Ir), release build, precompilation cache
warm, on the container this repo's [#7576](https://github.com/tokuhirom/mutsu/issues/7576)
rounds 10-16 use:

| program | Ir |
| --- | --- |
| `say "ok";` | 9,044,380 |
| `use YAMLish; say "ok";` | 333,679,418 |

So loading `YAMLish` (1,116 lines, 26 KB) costs **324 M instructions**, split:

- `parser::stmt::simple::module_exports::find_and_scan_module` — **209,663,766
  (62.8% of the run)**, called once, from the *parser*;
- `runtime::run_modules::load_module_inner` — 104,789,630, the run-time load
  (which does hit the precomp cache and does not re-parse).

For scale, that 324 M is **17.6%** of the whole 60-row YAMLish parse benchmark
`#7576` has been shaving, and 63% of it is the redundant half.

## Root cause

`find_and_scan_module` (`src/parser/stmt/simple/module_exports.rs:377`) memoizes
in a thread-local `MODULE_SCAN_CACHE`, keyed by resolved path — so it is paid
once per process, not once per `use`. But the scan itself
(`scan_module_source`, same file) calls `crate::parser::parse_program_partial`
on the module's whole source, and nothing persists the result across processes.

Every `mutsu` invocation that uses a module therefore re-parses that module's
source for its export list, even when `~/.cache/mutsu` (or wherever
`precomp::cache_dir` points) already holds a validated AST for exactly that
file. `make test` pays it 3,000-odd times for `Test` alone.

## Why this is not a small fix

The obvious shape — persist `ModuleScanResult` to disk next to the precomp
entry, keyed by path + source content hash the way `precomp::save_cached_unit`
already is — has an invalidation question the precomp cache does not:

`ModuleScanResult` deliberately carries **transitive** names. `scan_module_source`
[...]

PR #8098, opened 2026-09-12
perf(parser,runtime): index the source newlines, and stop scratch interpreters seeding an env they never read

tokuhirom:
Round 16 of #7576. Instructions retired on the ticket's 60-row YAMLish document go **1,983,139,212 -> 1,841,091,693 (-7.16%)**. Attribution is callgrind Ir throughout (deterministic, load-independent), with the module precompilation cache warmed before each profiled run per round 14's note.

Round 15 handed off "the profile is flat: `LocalKey::with` 7.3% across a dozen callers, allocator ~19%, `memcpy` 3.9%, largest single mutsu function 1.85%", with one named item worth ~0.7%. A fresh `callgrind_annotate --tree=caller` found three larger things instead — none of them in the regex matcher, and none visible under their own name.

## (a) `current_line_number` rescanned the whole source prefix, per call (-3.07%)

`$?LINE`, every parse warning, every `X::Comp` location and every test-assertion line resolve a parser position to a line number by counting the newlines that precede it. It did so by reconstructing the prefix as a `&str` and calling `.matches('\n').count()` — **O(offset) per lookup**, and therefore quadratic in the source length. The parser asks once per statement *attempt*, backtracking included, so parsing the 26 KB `YAMLish.rakumod` made **347,485** of those lookups.

`set_original_source` builds the newline offsets once now (the parse already reads every byte) and each lookup is a `partition_point`. The heredoc `LeakedRegion` path — whose line numbering jumps at the terminator, so it counts newlines in `[jump_offset, offset)` rather than in a plain prefix — carries the same index, expressed as a difference of two `partition_point`s.

`CharSearcher::next_match` inclusive **63,315,136 (3.19%) -> 2,441,692 (0.13%)**, `memchr_aligned` **34,747,454 -> 1,592,214**.

This one is not specific to YAMLish: every mutsu parse pays it, and the cost grows with the square of the file.

## (b) The grammar dynamic-variable scan built one `String` per character (-1.03%)

`collect_dynamic_var_decls` walked *every character position* of each rule pattern and, at each one, materialized the entire remaining pattern as an owned `String` in order to `strip_prefix(":my ")` it. It walks the `:` positions with `str::find` and tests borrowed slices now.

Two things around it went with the same change: the scan ran **twice over every pattern** (once for the declaration list, once for the per-rule map that gives each match of a declaring rule its own binding), and the registry walk that collects the patterns spelled its prefix test `k.resolve()` — an owned `String` per `token_defs` key per `.parse` call, the shape round 11 (a) removed from the proto-variant scan.

`collect_dynamic_var_decls` inclusive **25,652,327 (1.29%) -> 0** (it no longer appears in the profile).

## (c) A scratch interpreter still seeded the process magicals (-2.69%)

Rounds 10 and 12 stopped a regex/grammar scratch interpreter from building the built-in registry, re-scanning the bundled-battery directory, and running `init_io_environment`. What survived is that `Interpreter::new` still built `$*PID`, `$*TZ`, `@*ARGS`, `$*INIT-INSTANT` and `$*SCHEDULER` on every construction — a `getpid`, a `localtime_r`, a `SystemTime::now`, a `make_instance` and five `String` keys — although **all ten scratch construction sites spell `Interpreter { env: <the caller's env>, .. }`** (directly, or through `make_regex_eval_env`, which starts from `self.env.clone()`), so none of it was ever read. `%*ENV`'s OS sweep was the only part already behind the guard. This document builds **1,609** scratch interpreters.

Two smaller process constants went the same way: the built-in encoding table (~30 `String`s per construction) is a shared `Arc` template, safe for the same copy-on-write reason as the built-in registry (`register_encoding` goes through `cow_table_mut`); and `precomp::enabled_by_default` read `MUTSU_PRECOMP` — an environment lock plus a `String` — once per construction, where it is a process-wide switch read once now.

`Interpreter::new` inclusive **93M (4.7%) -> 64.2M (3.49%)**; `make_instant_from_posix` **3,566,796 -> 0**; `std::env::_var` **3,252 calls -> 1,644**.

## Tests

Three new pins, all verified against rakudo 2026.07:

- `t/lang/quoting/line-number-after-heredoc.t` — `$?LINE` across the heredoc leaked-region line-jump arithmetic;
- `t/grammar/grammar-dynvar-decl-scan-with-adverbs.t` — the declaration scan with regex adverbs, `:sym<…>` proto variants and a multi-byte literal in its path;
- `t/regex/regex-code-block-process-magicals.t` — all six magicals still resolve inside an embedded regex code block, through the caller's env.

`cargo fmt --all`, `make lint`, `make test` (4071 files, 43319 tests) and `make roast` (1437 files, 219635 tests) were all run locally. roast's only failures are the three documented container-only ones — `roast/6.c/S32-io/file-tests.t` and `roast/S16-filehandles/filetest.t` (running as `uid 0`) and `roast/S32-io/IO-Socket-Async.t` (sandboxed network), each matching `docs/agent-environments.md` exactly, test numbers included.

## Method note

Round 12's lesson was "a profile's flat tail regenerates itself; re-run the caller attribution after every round rather than carrying the previous round's *what is left* forward as a conclusion." Round 16 sharpens it in one direction: **the tail regenerates from a different subsystem than the one you have been working on.** Rounds 10-15 each found their item in the regex engine or the interpreter constructor; two of round 16's three are in the *parser*, which no previous round had profiled at all, because the module parse hides behind `LocalKey::with`, `malloc` and `memchr` exactly the way every earlier round's item hid behind `malloc` and `memcpy`.

The corollary is procedural: a `--separate-callers=2` run is what turned "`next_match` is called from `LocalKey::with`" (useless — every thread-local access is) into "`next_match` is called from `LocalKey::with` called from `stmt_list_with_mode`". One extra callgrind run, worth reaching for whenever the immediate caller of a hot leaf is a generic wrapper.

## Found on the way, filed separately

- #8095 — the parser re-parses a module's source to collect its exports even though a precompiled AST for that file is already on the disk cache the *runtime* load consults. `use YAMLish` costs 324 M Ir, 209 M of it that redundant parse. Filed rather than fixed because caching the scan result has a transitive-invalidation question (a scan harvests names from the modules it recursively scans) that wants deciding first.
- #8096 — a rule-local `:my $*VAR` declaration is established parse-wide and leaks past `.parse` into the caller's dynamic scope (rakudo scopes it to the rule's own match).

The ticket stays open.

---

### DIFF
--- a/news/2026-09/parser-line-numbers-were-quadratic-in-the-source.md
+++ b/news/2026-09/parser-line-numbers-were-quadratic-in-the-source.md
@@ -1,0 +1,118 @@
+# The parser's line numbers were quadratic in the source, and a scratch interpreter still seeded an env it never read
+
+Round 16 of [#7576](https://github.com/tokuhirom/mutsu/issues/7576). Instructions
+retired on the ticket's 60-row YAMLish document go **1,983,139,212 ->
+1,841,091,693 (-7.16%)**. Attribution is callgrind Ir throughout (deterministic
+and load-independent), with the module precompilation cache warmed before each
+profiled run per round 14's note.
+
+Round 15 handed off "the profile is flat: `LocalKey::with` 7.3% across a dozen
+callers, allocator ~19%, `memcpy` 3.9%, largest single mutsu function 1.85%",
+with one named item (`resolve_parsed_token_candidates_in_pkg` interning its
+package name) worth ~0.7%. A fresh `callgrind_annotate --tree=caller` found
+three larger things instead, none of them in the regex matcher and none of them
+visible under their own name.
+
+## (a) `current_line_number` rescanned the whole source prefix, per call (-3.07%)
+
+`$?LINE` — and every parse warning, every `X::Comp` location, every
+`test-assertion-line` record — resolves a parser position to a line number by
+counting the newlines that precede it. It did so by reconstructing the prefix as
+a `&str` and calling `.matches('\n').count()`: **O(offset) per lookup**, and
+therefore quadratic in the source length. The parser asks once per statement
+*attempt*, backtracking included, so parsing the 26 KB `YAMLish.rakumod` made
+**347,485** of those lookups, scanning an average of several kilobytes each.
+
+`set_original_source` now builds the newline offsets once (the parse already
+reads every byte) and each lookup is a `partition_point`. The heredoc
+`LeakedRegion` path — whose line numbering jumps at the terminator, so it counts
+newlines in `[jump_offset, offset)` rather than in a plain prefix — carries the
+same index, expressed as a difference of two `partition_point`s. `next_match`
+inclusive **63,315,136 (3.19%) -> 2,441,692 (0.13%)**, `memchr_aligned`
+**34,747,454 -> 1,592,214**. Pinned by `t/lang/quoting/line-number-after-heredoc.t`
+(verified against rakudo 2026.07).
+
+This one is not specific to YAMLish: it is paid by every mutsu parse, and its
+cost grows with the square of the file.
+
+## (b) The grammar dynamic-variable scan built one `String` per character (-1.03%, and -2.6% of the program's allocations)
+
+`.parse` establishes a grammar's parse-wide `:my $*/@*/%*NAME` declarations by
+scanning every rule pattern for them. `collect_dynamic_var_decls` walked *every
+character position* of the pattern and, at each one, materialized the entire
+remaining pattern as an owned `String` in order to `strip_prefix(":my ")` it —
+quadratic again, for an answer that is "no declarations" for almost every rule.
+It now walks the `:` positions with `str::find` and tests borrowed slices.
+
+Two things around it went with the same change: the scan ran **twice over every
+pattern** (once to collect the declarations, once to build the per-rule map that
+gives each match of a declaring rule its own binding), and the registry walk that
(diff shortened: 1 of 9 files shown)


tokuhirom:
`gc-stress-tap` went red on `305765e0`, but not on anything this PR did — the TAP suite under GC stress passed in full and the job died in its artifact-upload post step.

From [the job log](https://github.com/tokuhirom/mutsu/actions/runs/34686108985/job/103533282703):

```
All tests successful.
Files=4081, Tests=43396, 398 wallclock secs
Result: PASS
No quarantine retries were needed in this job.
No GC verify violations across the TAP suite.
No crash reports: nothing died of a fatal signal in this job.
...
With the provided path, there will be 1 file uploaded   (tmp/gc-stress-t.log)
##[error]Failed to CreateArtifact: Unable to make request: ETIMEDOUT
```

The `ETIMEDOUT` is `actions/upload-artifact` failing to reach GitHub's artifact service, ~2m13s after it started the upload. It names a service the diff does not touch: this PR changes the parser's line-number index, a grammar declaration scan and what `Interpreter::new` seeds into a scratch interpreter's env — none of which is reachable from an artifact upload, and all of which the suite that just passed exercises.

Re-running the failed job once, per the one-re-run rule. A second failure of the same step would be real and I would treat it as such.

---

Issue #8134, opened 2026-09-12
`.sort` on an Instance whose display differs from its (apparent) sort key loses the display form


tokuhirom:
## Summary

Found while implementing #8070 (a `Version` subclass's native constructor). `@vs.sort`'s result
Seq, gisted directly, shows a *different* per-element string than gisting the same elements any
other way (a plain `for` loop, or the un-sorted `Array`) — the sort itself appears to replace each
element with something derived from a string comparison key, rather than keeping the original
value.

## Repro

```raku
class MyVer is Version { }
my @vs = MyVer.new("1.0.0"), MyVer.new("2.0.0");
say @vs;                                    # [v1.0.0 v2.0.0]  -- correct
say @vs.sort;                               # should be (v1.0.0 v2.0.0)
for @vs.sort -> $x { say $x.^name, " ", $x }  # MyVer3 v1.0.0 / MyVer3 v2.0.0 -- correct
```

| | rakudo | mutsu |
| --- | --- | --- |
| `@vs` | `[v1.0.0 v2.0.0]` | `[v1.0.0 v2.0.0]` |
| `@vs.sort` | `(v1.0.0 v2.0.0)` | **`(1.0.0 2.0.0)`** |
| `for @vs.sort -> $x { say $x.^name, " ", $x }` | `MyVer v1.0.0` / `MyVer v2.0.0` | `MyVer v1.0.0` / `MyVer v2.0.0` |

The ordering is correct in both cases, and iterating the sorted Seq element-by-element (a `for`
loop) shows the right `.^name` AND the right gisted form (`v1.0.0`) — only gisting the **Seq
itself** directly (`say @vs.sort`) loses the "v" prefix.

## Why this isn't `is Str`-subclass-blind

The same shape with `is Str` does NOT show a visible bug:

```raku
class MyStr is Str { }
my @xs = MyStr.new(:value("banana")), MyStr.new(:value("apple"));
say @xs;        # [banana apple]
say @xs.sort;   # (apple banana)  -- correct in both rakudo and mutsu
```

That is not evidence of correctness, though — a `Str` subclass's default `to_string_value()`
happens to equal its display form, so if `sort` is silently replacing each element with a string
*sort key* (rather than keeping the original element), the substitution is invisible for `Str`. For
`Version`, the sort key (`"1.0.0"`, no `v` prefix, matching `.Str`) differs from the display form
(`"v1.0.0"`, `.gist`/`.raku`), which is what exposes it.

[...]

PR #8140, opened 2026-09-12
fix(oo): a user class `is Version` inherits Version's native constructor

### LINKED ISSUE #8070: A user class `is Version` does not inherit Version's native positional `new` (2 distributions)
Split off from #8003 while investigating its "9 distributions" list. #8003 itself is fixed (a `subset`'s `where` predicate now runs in its declaration package, closing the `Net::Netmask` gap), but two of the nine listed members — `Version::Raku` and `Version::Nginx` — fail with the *same symptom* ("Default constructor ... only takes named arguments") for a **different, unrelated root cause**: neither declares any `subset`, and neither references a grammar.

## Repro

```raku
class MyVer is Version { }
say MyVer.new("1.2.3");
```

```
$ raku repro.raku
v1.2.3

$ mutsu repro.raku
Default constructor for 'MyVer' only takes named arguments
  in block <unit> at repro.raku line 3
```

Both `Version::Raku` and `Version::Nginx` are exactly this shape: `class Version::Raku ... is Version { ... }` / `class Version::Nginx ... is Version { }`, adding only extra methods, and relying on inheriting `Version`'s own positional-string constructor.

## Root cause

`src/runtime/methods_object_dispatch_new.rs`'s native-constructor dispatch matches on the **literal** `constructor_dispatch_name`:

```rust
"Version" => {
    let arg = args.first().cloned().unwrap_or(Value::NIL);
    return Ok(Self::version_from_value(arg));
}
```

For a subclass (`MyVer is Version`), `constructor_dispatch_name` is `"MyVer"`, not `"Version"`, so this arm never matches and dispatch falls through to the default (named-arguments-only) constructor. Compare with the `IO::CatHandle` / `IO::Path` handling a few lines above in the same function, which explicitly walks `self.class_mro(class_key)` to also catch a subclass:

```rust
let is_cathandle_like = base_class_name == "IO::CatHandle"
    || self.class_mro(class_key).iter().any(|name| name == "IO::CatHandle");
if is_cathandle_like && !self.has_user_method(class_key, "new") {
    return Ok(self.build_io_cathandle(*class_name, &args));
}
```

`Version` has no equivalent MRO-aware branch. Note `build_io_cathandle` is also passed `*class_name` (the actual, possibly-subclassed name) so the built instance is tagged as the subclass — `version_from_value` would need the same treatment (or Version's own internal representation may need checking for whether it can carry a subclass tag at all; not yet investigated).

## Why this is a separate ticket, not folded into #8003

- No `subset`, no `grammar`, no `where` clause anywhere in either affected module.
- The failure reproduces with a two-line repro that has nothing in common with #8003's mechanism (a subset predicate's declaration-scope resolution).
- Fixing it requires touching the native-constructor MRO-subclass dispatch shape (`methods_object_dispatch_new.rs`), a different code path than #8003's subset/type-matching fix.

## Affected

- `Version::Raku`
- `Version::Nginx`

## Reproducing

```sh
scripts/ecosystem-sweep.py --only Version::Raku --only Version::Nginx
```

Add a `t/` regression test for the minimized case above, not for either distribution.

tokuhirom:
## Summary

`class MyVer is Version { }` died with `Default constructor for 'MyVer' only takes named
arguments` on `MyVer.new("1.2.3")`, because `constructor_dispatch_name` matched Version's
constructor arm on the **literal** name `"Version"` — a subclass's own dispatch name (`"MyVer"`)
never matched, so dispatch fell through to the generic named-arguments-only constructor. This
blocked the `Version::Raku` and `Version::Nginx` distributions, both `class Version::* is Version
{ ... }` relying on inheriting the positional-string constructor.

## Repro

```raku
class MyVer is Version { }
say MyVer.new("1.2.3");
```
| | rakudo | mutsu (before) |
|---|---|---|
| output | `v1.2.3` | `Default constructor for 'MyVer' only takes named arguments` |

## Why not just an MRO check like `IO::CatHandle`/`IO::Path`

`Version`'s own representation (`ValueView::Version { parts, plus, minus, text }`) carries no
class-name tag to fall back on. Just relaxing the dispatch match to also catch a subclass via MRO
(the existing `is_cathandle_like`/`is_io_path_like` pattern) while still returning a bare
`ValueView::Version` would leave the built value reporting `.^name` as `"Version"`, and — more
importantly — leave a subclass's own added methods (`Version::Raku`'s actual reason for existing)
unreachable, since nothing about a bare `ValueView::Version` can dispatch through `MyVer`'s method
table.

## Fix

A genuine subclass (MRO includes `Version`, dispatch name isn't the literal `"Version"`, no
user-defined `new`) is built as an ordinary tagged `Instance` — the same `__mutsu_*_value`
native-payload convention `is Int`/`is Str` subclasses already use — carrying the built `Version`
under `__mutsu_version_value`. This gives the instance a real method table (so a subclass's own
methods resolve, and `~~ MyVer` / `~~ Version` both work via ordinary class-MRO smartmatch, no
Version-specific code needed) while `.Str`/`.raku`/`.gist` and anything else the subclass doesn't
implement fall through to the real Version's behavior, via the same three call sites already wired
for `is Int`/`is Str` (`display.rs`, `gist.rs`, `methods_instance_ops.rs`).

Plain `Version.new(...)` is untouched — the literal-name arm still returns a bare
`ValueView::Version`, exactly as before.

Closes #8070

## A separate finding, filed rather than chased here

Testing surfaced an unrelated pre-existing gap: `@vs.sort` on an array of `Version` subclass
instances loses the `v` prefix when the *sorted Seq itself* is gisted directly, even though the
same elements gist correctly individually, in a `for` loop, or as a plain unsorted `Array`. Points
at `sort`'s own internal handling, not at construction/display — filed as #8134.

## Testing

- `t/oo/construct/version-subclass-constructor.t` — new: the reported repro, `.^name`, both
  smartmatch directions, a subclass's own added method, plain `Version.new` unaffected, and a
  subclass's own `new(|c)` override still reaching the native constructor via `callsame`.
- `cargo fmt --all`, `make lint` — green (all four configurations).
- `make test` — green, `Files=4101, Tests=43627, Result: PASS`.
- `make roast` — the only failures are the documented container-only entries in
  `docs/agent-environments.md` (`6.c/S32-io/file-tests.t`, `S16-filehandles/filetest.t` — this
  container runs as `uid 0`); both pass in CI.

https://claude.ai/code/session_01DgfeutFHT5ZhakKNVXXoV7

---
_Generated by [Claude Code](https://claude.ai/code/session_01DgfeutFHT5ZhakKNVXXoV7)_

### DIFF
--- a/news/2026-09/version-subclass-native-constructor.md
+++ b/news/2026-09/version-subclass-native-constructor.md
@@ -1,0 +1,37 @@
+# A user class `is Version` now inherits Version's native positional constructor
+
+`class MyVer is Version { }` died with `Default constructor for 'MyVer' only takes named
+arguments` on `MyVer.new("1.2.3")`, because the native-constructor dispatch in
+`methods_object_dispatch_new.rs` matched Version's constructor arm on the **literal** dispatch
+name `"Version"` — a subclass's dispatch name is its own (`"MyVer"`), so the arm never matched and
+fell through to the generic named-arguments-only constructor. This blocked the `Version::Raku` and
+`Version::Nginx` distributions, both `class Version::* is Version { ... }` adding only extra
+methods and relying on inheriting the positional-string constructor.
+
+Unlike `IO::CatHandle`/`IO::Path` (whose MRO-aware subclass check a few lines above already
+handles this), `Version`'s own representation (`ValueView::Version { parts, plus, minus, text }`)
+carries no class-name tag to fall back on — building one the same way `Version.new` always has
+would leave every subclass instance reporting `.^name` as `"Version"`, and would leave a
+subclass's own added methods (`Version::Raku`'s actual reason for existing) unreachable, since
+nothing about a bare `ValueView::Version` value can dispatch through `MyVer`'s method table.
+
+A genuine subclass (MRO includes `Version`, dispatch name isn't the literal `"Version"`, no
+user-defined `new`) is now built as an ordinary tagged `Instance` — the same `__mutsu_*_value`
+native-payload convention `is Int`/`is Str` subclasses already use — carrying the built `Version`
+under `__mutsu_version_value`. This gives the instance a real method table (so `Version::Raku`'s
+own methods resolve, and `~~ MyVer` / `~~ Version` both work via the ordinary class-MRO
+smartmatch, no Version-specific code needed) while `.Str`, `.raku`, `.gist`, and any other method
+the subclass does not implement fall through to the real Version's behavior — the same three
+call sites (`display.rs`, `gist.rs`, `methods_instance_ops.rs`) already wired for `is Int`/`is Str`.
+
+Plain `Version.new(...)` (the overwhelmingly common case) is untouched: the literal-name arm still
+returns a bare `ValueView::Version`, exactly as before.
+
+One gap surfaced but left alone as out of scope, filed as #8134: `@vs.sort` on an array of
+`Version` subclass instances loses the "v" prefix when the *sorted Seq itself* is gisted directly
+(`say @vs.sort` prints `(1.0.0 2.0.0)` instead of `(v1.0.0 v2.0.0)`), even though the same elements
+gist correctly individually, in a `for` loop, or as a plain (unsorted) `Array`. This points at
+`sort`'s own internal handling rather than at construction or display, and is unrelated to the
+reported symptom.
+
+Fixes #8070.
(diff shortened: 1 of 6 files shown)


tokuhirom:
`test-suites` (and the `test` aggregator) failed on this PR's first CI run, but not because of this diff: `scripts/battery-testsuite.sh` couldn't even set up the `Template::Mustache` battery —

```
error: could not fetch 27f3e862d1060c60042956b852cf39dc183b56a5 from https://github.com/raku-community-modules/Template-Mustache.git
```

That's a `git fetch` failure against an external repo at CI runtime, unrelated to anything this PR
touches (constructor dispatch, display/gist formatting for native subclass instances) — every
other battery in the same run passed or reported its usual pre-existing failures. Re-ran the
failed jobs once to confirm it's transient before doing anything else.

---

Issue #8153, opened 2026-09-12
`.Str` on an instance falls back to the class's own `Numeric` method instead of `Mu.Str`


tokuhirom:
Found while fixing [#8115](https://github.com/tokuhirom/mutsu/issues/8115) (`Enumeration` as a composable core role), but entirely independent of it.

A class that defines a `Numeric` method and no `Str` method gets its `.Str` routed to `Numeric`. In rakudo `Str` falls back to `Mu.Str` — which renders `TypeName<id>` — and a `Numeric` method has no bearing on it.

## Minimal repro

```raku
class P { method Numeric { 7 } }
say P.new.Str;
```

mutsu:

```
7
```

raku:

```
P<3722464423584>
```

The same divergence shows through every stringifying context (`~$p`, `"$p"`, `say $p` via `gist`), and it fires for any class with a `Numeric` method, so it is not tied to any one type.

## Why it matters beyond the ugly default

`TypeName<id>` looks like something nobody depends on, but the fallback means a class cannot have a `Numeric` method *and* the default object stringification at the same time — the numeric coercion silently doubles as the string one. It surfaced on a class composing the core `Enumeration` role, whose `Numeric` returns `$.value`: rakudo gives `Tern<4665716083224>` for `.Str` there and mutsu gives `1`.

## Affected code

The instance `Str` fallback chain — see `src/runtime/methods_instance_ops.rs`'s fallback list (`["Dateish", "Real", "Numeric", "Cool", "Any"]` around line 3060) and the `Str` dispatch that consults it. The fix is presumably to stop treating a user `Numeric` method as a `Str` provider, while keeping the `Dateish` private-`!formatter` path and the genuine `Cool`/`Stringy` routes that existing tests pin (`t/oo/role/dateish-role-composition.t`).

## Test

A `t/` regression pinning `class P { method Numeric { 7 } }; P.new.Str` against the `TypeName<...>` shape, plus the `~`/interpolation forms.

tokuhirom:
Claiming: claude/todo-ticket-pr-8f98ok

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #7993, opened 2026-09-11
ecosystem parity: Invalid typename in parameter declaration blocks 21 distributions


tokuhirom:
Measured by the ecosystem parity ledger ([#7785](https://github.com/tokuhirom/mutsu/issues/7785) P5, [ADR-0085](https://github.com/tokuhirom/mutsu/blob/main/docs/adr/0085-ecosystem-testsuite-parity-measurement.md)). One root cause, grouped across the corpus by `scripts/ecosystem-tickets.py`.

`eco-cluster: 1332ae2d`  ·  family `invalid-typename`

## Impact

| | |
|---|---|
| distributions affected | **21** |
| of which `blocked_load` | 20 |
| axis | guts 3, native 6, pure 12 |

## The failure

```
Invalid typename 'CrossAlign' in parameter declaration.
```

What varies (count = distributions): `IndRef` ×6, `CrossAlign` ×3, `TAP::Entry::Handler` ×3, `Sizing` ×2, `Box`, `Column::List(Any)`, `EncodeBuffer`, `Enumeration`, `GType`, `NCName`, `NcplaneHandle`, `Net::BGP::Error:D`, `Event:D`, `License::Software::Year` — 14 distinct typenames.

The names point at several different resolution paths, which is why this is `todo:deep` rather than one fix — the first job is to decide how many gaps these 14 names actually are:

- **a type from another module in the dist** (`CrossAlign`, `Sizing` in `Grammar::Editor`/`Selkie`, `TAP::Entry::Handler` in `TAP`) — an imported/exported type not visible where a signature names it;
- **a `subset`** declared in the same or a sibling module (`IndRef` in the PDF family, `NCName` in `LibXML`, `Enumeration` in `Logic::Ternary`);
- **a definedness-constrained name** (`Net::BGP::Error:D`, `Event:D`) — the `:D` smiley on a user type in a parameter;
- **a coercion type** (`Column::List(Any)` in `SQL::Abstract`) — different machinery again.

## Suggested approach

1. Take one from each of the four groups above, minimise, and see whether they collapse (e.g. all four are "a type declared after / outside the compilation unit that names it") or are genuinely four gaps.
2. File the narrower tickets that come out of that, and keep or close this one accordingly.
3. `PDF` / `PDF::Class` / `PDF::Content` / `PDF::Font::Loader` / `FDF` / `Pod::To::PDF::Lite` are six of the 21 and all fail on `IndRef`, so that group alone is a third of the impact.

[...]

PR #8164, opened 2026-09-12
fix(oo): make `Enumeration` a composable core role, not just an enum constraint

### LINKED ISSUE #8115: `Enumeration` is not a composable core role, so `class Foo does Enumeration` fails
Found while re-grouping the `invalid-typename` ecosystem cluster ([#7993](https://github.com/tokuhirom/mutsu/issues/7993)) after [#8066](https://github.com/tokuhirom/mutsu/pull/8066) landed. `Logic::Ternary`'s entry in that cluster is not a typename-resolution gap at all — the name it reports, `Enumeration`, is a core role mutsu has as a *type-check constraint* but not as a *composable role*.

## The gap

Every other natively-modelled core role composes:

```
$ for r in Dateish Positional Associative Numeric Real Callable Stringy Iterable Enumeration; do
    mutsu -e "class C does $r { }; say 'ok'"
  done
Dateish        ok
Positional     ok
Associative    ok
Numeric        ok
Real           ok
Callable       ok
Stringy        ok
Iterable       ok
Enumeration    X::InvalidType: Invalid typename 'Enumeration'
```

`Enumeration` is in `runtime/utils/type_constraints.rs`, `types/type_matching.rs` and `value/types_isa.rs` as a constraint that enum values satisfy, but it is in neither list of composable core roles — `BUILTIN_ROLE_NAMES` (`src/runtime/types/type_registry.rs`) nor the known-parent list in `src/runtime/registration_class_decl.rs`. So `does Enumeration` never resolves.

## Minimal repro

```raku
class Tern does Enumeration {
    method new(Str:D $val) { self.bless: key => $val, value => 1 }
}
my $t = Tern.new('yes');
say $t.key, ' ', $t.value;
say $t ~~ Enumeration;
```

mutsu:

```
X::InvalidType: Invalid typename 'Enumeration'
  in block <unit> at tmp/enum-role3.raku line 1
```

raku:

```
yes 1
True
```

## Why it is more than adding the name to a list

Rakudo's `Enumeration` is a real role with state and a documented API, not a marker:

- it declares `has $.key` and `has $.value` — which is exactly what the repro above relies on, and what makes `class C does Enumeration { has $.key }` correctly fail in rakudo with *"Attribute '$!key' already exists in the class 'C', but a role also wishes to compose it"*;
- it supplies `key`, `value`, `kv`, `pair`, `enums`, `CALL-ME`, `pick`, `roll`, `pred`, `succ`, `Numeric`, `Int`, `Real` and `===` (`raku-doc/doc/Type/Enumeration.rakudoc`).

mutsu implements all of that natively for *enum values* (which are their own `Value` shape), so the work is to make the same behaviour available to an arbitrary user class that composes the role, rather than to write it a second time. Deciding how much of the API a composing class gets, and where it lives relative to the existing enum-value fast paths, is the substance of the ticket.

## Affected distributions

`Logic::Ternary` 0.0.4, whose `lib/Logic/Ternary.rakumod:34` is `class Logic::Ternary does Enumeration {` and which supplies its own `new` blessing `key`/`value` exactly as the repro does. One of the 21 in #7993.

## Reproducing against the real distribution

```sh
mutsu -I <Logic-Ternary-0.0.4>/lib -e 'use Logic::Ternary; say "ok"'
```

## Test

Add a `t/oo/role/` regression test for the minimised case, not for the distribution.

tokuhirom:
## The gap

`class Foo does Enumeration { ... }` died as `X::InvalidType: Invalid typename 'Enumeration'`, while every other natively-modelled core role (`Dateish`, `Positional`, `Associative`, `Numeric`, `Real`, `Callable`, `Stringy`, `Iterable`) composed fine. `Enumeration` existed in mutsu only as a *type-check constraint* that enum values satisfy, and in neither list of composable core role names.

Rakudo's `Enumeration` is a real role with state — `has $.key`, `has $.value` — and `raku-doc/doc/Type/Enumeration.rakudoc` documents composing it from an ordinary class as a supported way to build a constrained key/value pair (`class DNA does Enumeration`, whose `new` blesses `key`/`value` itself). `Logic::Ternary` 0.0.4 does exactly that, which is how the gap surfaced: its entry in the `invalid-typename` ecosystem cluster (#7993) was never a typename-resolution problem at all.

## The fix

The role is now supplied as **real Raku source** (`ENUMERATION_ROLE_PRELUDE` in `src/runtime/run.rs`), parsed once and spliced into any compunit that names it, the way the `Rational`, `IO::Socket` and `Metamodel::Naming` roles already are. A composing class therefore gets the role's attributes, its generated accessors, `$!key`/`$!value` visibility inside its own method bodies, and the attribute-collision diagnostic, all from the ordinary role-composition path rather than from a second native implementation of behaviour mutsu already has for enum values.

The method set is the part of `Enumeration.^methods` that actually **works** on a composing class in rakudo, measured there rather than transcribed from the role's source: `kv`, `pair`, `Numeric`, `Int`, `Real`, `gist`, `raku`. The rest (`enums`, `pred`, `succ`, `pick`, `roll`, `CALL-ME`) reach `self.^enum_values`, which a `ClassHOW` does not have — they die in rakudo too (`No such method 'enum_values' for invocant of type 'Perl6::Metamodel::ClassHOW'`), so supplying them would be a divergence, not a feature. `.Str` is absent for the same reason: rakudo's composing class falls back to `Mu.Str`.

Three smaller pieces go with it:

- the prelude is injected for **modules** as well as the main program, because the motivating distribution composes the role inside the module while the program that loads it (`use Logic::Ternary;`) never names `Enumeration` at all, so gating on the main program's source alone would never fire;
- `Enumeration` joins `BUILTIN_ROLE_NAMES`, which one consumer needs *before* any prelude is registered: the compiler qualifies a class header's `does` parent with the enclosing package unless the name is a known core type, so without the entry `unit module M; class C does Enumeration` compiled a parent named `M::Enumeration`;
- the two `constraint == "Enumeration"` fast paths (`type_matches_value` and the `~~` handler in `seq_helpers/smart_match.rs`) returned `false` for anything that was not an enum value or enum type object, pre-empting the ordinary composed-role check. They now assert enum membership and otherwise fall through, so `$instance ~~ Enumeration` is `True` for a class that has composed the role and still `False` for one that has not.

## Verification

The whole minimised repro, and the documented `DNA` example, now produce byte-identical output under mutsu and rakudo. The real distribution loads too:

```
$ mutsu -I Logic-Ternary-0.0.4/lib -e 'use Logic::Ternary; say "ok"'
ok
```

which moves its `ecosystem/` record off `blocked_load`. Its own test suite is still red on unrelated gaps (an exported `infix:<not3>` invisible to the importer, a stack overflow in the operator file, an `ACCEPTS` multi-resolution failure, an `EXPORT` sub re-adding a method to `Any`, and a `Bool`-into-class coercion) — those belong to the `ecosystem-dist-fix` workflow and the next sweep's re-measure, not to this ticket, which scoped the test to the minimised case explicitly. The ledger record is left for that sweep to rewrite rather than hand-edited.

Pinned by `t/oo/role/class-does-enumeration-role.t`, whose 18 assertions were all measured against rakudo first, including the two enum-value non-regressions.

One finding outside the slice was filed rather than folded in: #8153 — `.Str` on an instance falls back to the class's own `Numeric` method instead of `Mu.Str` (`class P { method Numeric { 7 } }; say P.new.Str` gives `7`, rakudo gives `P<...>`). It is a pre-existing, general behaviour that composing `Enumeration` merely makes newly reachable.

## Suites

- `cargo fmt --all`, `make lint` — clean (all four configurations).
- `make test` — PASS, 4112 files / 43769 tests.
- `make roast` — the three documented container-only failures and nothing else: `roast/6.c/S32-io/file-tests.t` (`Failed: 4`, tests 6-8 and 10), `roast/S16-filehandles/filetest.t` (`Failed: 25`, tests 57-64, 69-72, 77-80, 101/103/105/107/109/111/117/121/125), `roast/S32-io/IO-Socket-Async.t` (exit 124, planned 40 ran 17) — matching `docs/agent-environments.md` by name and by subtest range.

Closes #8115

---
_Generated by [Claude Code](https://claude.ai/code/session_01EBmiJH8aYB8BQhXFJZaTQ7)_

### DIFF
--- a/news/2026-09/enumeration-is-a-composable-core-role.md
+++ b/news/2026-09/enumeration-is-a-composable-core-role.md
@@ -1,0 +1,62 @@
+# `Enumeration` is a composable core role, not just an enum-value constraint
+
+`class Foo does Enumeration { ... }` died as `X::InvalidType: Invalid typename
+'Enumeration'`. Every other natively-modelled core role composed — `Dateish`,
+`Positional`, `Associative`, `Numeric`, `Real`, `Callable`, `Stringy`,
+`Iterable` all accept a `does` — because `Enumeration` existed in mutsu only as
+a *type-check constraint* that enum values satisfy
+(`runtime/utils/type_constraints.rs`, `types/type_matching.rs`,
+`value/types_isa.rs`) and in neither list of composable core role names.
+
+That is not a cosmetic gap. Rakudo's `Enumeration` is a real role with state —
+`has $.key`, `has $.value` — and `raku-doc/doc/Type/Enumeration.rakudoc`
+documents composing it from an ordinary class as a supported way to build a
+constrained key/value pair (`class DNA does Enumeration`, whose `new` blesses
+`key`/`value` itself). `Logic::Ternary` 0.0.4 does exactly that, which is how
+the gap surfaced: its entry in the `invalid-typename` ecosystem cluster
+([#7993](https://github.com/tokuhirom/mutsu/issues/7993)) was never a
+typename-resolution problem at all.
+
+## What it took
+
+The role is now supplied as **real Raku source** (`ENUMERATION_ROLE_PRELUDE` in
+`src/runtime/run.rs`), parsed once and spliced into any compunit that names it,
+the way the `Rational`, `IO::Socket` and `Metamodel::Naming` roles already are.
+A composing class therefore gets the role's attributes, its generated
+accessors, `$!key`/`$!value` visibility inside its own method bodies, and the
+attribute-collision diagnostic, all from the ordinary role-composition path
+rather than from a second native implementation of behaviour mutsu already has
+for enum values.
+
+The method set is the part of `Enumeration.^methods` that actually *works* on a
+composing class in rakudo, measured there rather than transcribed from the
+role's source: `kv`, `pair`, `Numeric`, `Int`, `Real`, `gist`, `raku`. The rest
+(`enums`, `pred`, `succ`, `pick`, `roll`, `CALL-ME`) reach
+`self.^enum_values`, which a `ClassHOW` does not have — they die in rakudo too,
+so supplying them would have been a divergence, not a feature. `.Str` is absent
+for the same reason: rakudo's composing class falls back to `Mu.Str`, even
+though a real enum value stringifies to its key.
+
+Three smaller pieces went with it:
+
+- the prelude is injected for **modules** as well as the main program, because
+  the distribution that motivated this composes the role inside the module
+  while the program that loads it (`use Logic::Ternary;`) never names
+  `Enumeration` at all;
+- `Enumeration` joined `BUILTIN_ROLE_NAMES`, which one consumer needs *before*
+  any prelude is registered: the compiler qualifies a class header's `does`
+  parent with the enclosing package unless the name is a known core type, so
+  without the entry `unit module M; class C does Enumeration` compiled a parent
+  named `M::Enumeration`;
+- the two `constraint == "Enumeration"` fast paths (`type_matches_value` and
+  the `~~` handler in `seq_helpers/smart_match.rs`) returned `false` for
+  anything that was not an enum value or enum type object, pre-empting the
+  ordinary composed-role check. They now assert enum membership and otherwise
+  fall through, so `$instance ~~ Enumeration` is `True` for a class that has
+  composed the role and still `False` for one that has not.
+
+Pinned by `t/oo/role/class-does-enumeration-role.t`, whose 18 assertions were
+all measured against rakudo first — including the documented `DNA` example and
+the two enum-value non-regressions.
+
+Closes [#8115](https://github.com/tokuhirom/mutsu/issues/8115).
--- a/src/runtime/run.rs
+++ b/src/runtime/run.rs
@@ -214,6 +214,40 @@ role GLOBAL::IO::Socket {
 }
 "#;
 
+/// Builtin `Enumeration` role. mutsu models the role's behaviour natively for
+/// *enum values* (they are their own `Value` shape, with native `key`/`value`/
+/// `kv`/`pair`/`Numeric` methods), but Rakudo's `Enumeration` is a real role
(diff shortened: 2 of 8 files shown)


tokuhirom:
`test-suites` went red on `roast/S02-types/set.t`, `Failed: 1` — and the failing assertion is **test 127, `Set.pick(*) returns elements in a random order`**, which is not this PR's:

```
roast/S02-types/set.t (Wstat: 256 (exited 1) Tests: 250 Failed: 1)
  Failed test:  127
```

That subtest is `isnt @a.join, 'abcdefgh'` on `set(<a b c d e f g h>).pick(*)` (`roast/S02-types/set.t:275`) — an upstream assertion that a random permutation of 8 distinct elements is not the sorted one, so it fails by construction once every `8! = 40320` runs. The job's own re-run of the file passed all 250. This diff touches role composition, the `Enumeration` constraint fast paths and a prelude injection; it goes nowhere near `Set`, `pick`, or the RNG.

Ruled out a biased shuffle rather than assuming, since a skewed `pick(*)` would raise that rate far above 1/40320 — 24000 draws of `set(<a b c d>).pick(*)` produce all 24 permutations with counts 936-1068 against an expected 1000 (±3σ ≈ ±92), i.e. uniform:

```
24 distinct permutations
min=936 max=1068 expected=1000
```

`roast/S02-types/set.t` is not in `flaky-tests.txt`, which is why the in-job retry did not absolve it. I am not adding it there from this PR: `docs/flaky-test-policy.md` sets a deliberately high bar for that shared ledger, the entry would want its own measurement and review date, and the neighbouring statistical files (`bag.t`, `baghash.t`, `mixhash.t`) were removed from it once upstream raised their sample counts — the same treatment this assertion wants, upstream, rather than a quarantine here.

Re-running the failed jobs once. A second failure on the same subtest would mean the rate is not 1/40320 and I would chase it as real.

Local runs before publication: `make lint` clean, `make test` PASS (4112 files / 43769 tests), `make roast` green apart from the three documented container-only files.

---

PR #8169, opened 2026-09-12
fix: dispatch the listop invocant colon as a method everywhere

### LINKED ISSUE #8141: the invocant-colon listop form (`die "x":`) drops the colon instead of dispatching a method, and `die` does not parse at all
Split out of the parse-gap cluster [#7988](https://github.com/tokuhirom/mutsu/issues/7988) while working the head of its survey table. It is the one head construct I deliberately did **not** fix there, because the small fix would have traded a loud wrong answer for a quiet one.

## The construct

Raku's listop invocant colon makes the first argument the invocant: `foo $x: @args` means `$x.foo(@args)`. With no arguments after the colon, `die "boom":` therefore means `"boom".die` — and since `Str` has no `die` method, rakudo raises `X::Method::NotFound`.

Math::FractionalPart writes exactly that (`lib/Math/FractionalPart.rakumod:129`):

```raku
die "FATAL: Unexpected number format '$x' with no embedded 'e'; please report this issue":
```

## What mutsu does

mutsu does not implement the invocant-colon semantics for **any** listop — it drops the colon and calls the listop normally. `die` is the only one that additionally fails to *parse*. Measured against rakudo 2026.07:

| construct | rakudo | mutsu |
|---|---|---|
| `say "hello":` | `hello` (`"hello".say`) | `hello` — same output, but by dropping the colon |
| `my @a = 3,1,2; say sort @a:` | `(1 2 3)` | `(1 2 3)` — likewise |
| `sub h() { return "r": }; say h()` | `r` | `r` — likewise |
| `sub g() { warn "w": }; g()` | `No such method 'warn' for string 'w'` | warns `w` — **wrong answer** |
| `sub f() { die "boom": }; f()` | `No such method 'die' for string 'boom'` | **parse error** |

The `say` / `sort` / `return` rows agree only by coincidence: dropping the colon and dispatching the method happen to produce the same result there. `warn` is where the divergence becomes visible, and `die` is where it becomes loud.

## Why this is one issue and not two

Fixing `die`'s parse alone — making it behave like mutsu's other listops, i.e. dropping the colon — would make Math::FractionalPart load, but it would silently `die "FATAL: …"` where rakudo raises `X::Method::NotFound`. That is a worse failure mode than the current parse error, and it would hide the real gap behind three more distributions that look fixed.

The real gap is the shared one: **a listop's invocant colon should dispatch a method on the first argument.** Fix that and `die` follows, `warn` stops giving a wrong answer, and `say` / `sort` / `return` become right for the right reason.

## Affected distributions

`Math::FractionalPart`, `Astro::Utils`, `DateTime::Julian` (the survey's count for this construct — all three reach it through `Math::FractionalPart`). The `warn` divergence is not blocking any measured distribution but is a live wrong answer wherever it occurs.

## Reproducing

```sh
raku  -e 'sub f() { die "boom": }; f()'   # No such method 'die' for string 'boom'
mutsu -e 'sub f() { die "boom": }; f()'   # ===SORRY!=== parse error

raku  -e 'sub g() { warn "w": }; g()'     # No such method 'warn' for string 'w'
mutsu -e 'sub g() { warn "w": }; g()'     # warns "w"

scripts/ecosystem-sweep.py --only Math::FractionalPart
```

A regression test belongs in `t/routines/call/`, covering all five listops above so the shared semantics is pinned rather than just `die`'s parse.

tokuhirom:
## Summary

Raku's listop invocant colon (`foo $x: @args` == `$x.foo(@args)`) already parsed and compiled correctly for a non-tail statement-level call, but several other compiler paths either dropped the invocant marker and called the listop as a plain function, or hit `unreachable!()`:

- `compile_tail_stmt_call_value` (shared by several tail-position call sites) panicked on `CallArg::Invocant` in its positional/named-pairs split.
- `compile_routine_body_stmts` (a named `sub`'s own tail statement) and `compile_closure_body_with_routine_flag`'s value-producing branch both silently converted the invocant into a plain positional argument, dropping the colon.

`die`/`fail` additionally had their own dedicated statement parser and expression-position arm, neither of which recognized a trailing invocant colon at all — `die "boom":` was a hard parse error instead of dispatching `.die`.

### Fix

- Added `Compiler::invocant_colon_method_call`, a shared helper that turns `(name, args)` into the equivalent `Expr::MethodCall` when `args` opens with an invocant, wired into the three tail-position compiler sites that previously mishandled it.
- Taught `die_stmt` and `die`/`fail`'s expression-position arm to try the same no-paren invocant-colon parse `say`/`print`/`put`/`note` already use (`parse_io_colon_invocant_stmt`, made `pub(super)` and documented as the general shape it always was, not something I/O-specific) before falling back to their own statement form.

`say $x:` / `sort $x:` / `return $x:` already matched rakudo by coincidence — `Str`/`Array` really do have `.say`/`.sort` methods, and `.return` is a real control-flow method mutsu already implements, so dropping the colon and dispatching the method happened to agree. `warn $x:` / `die $x:` did not, since `Str` has no such method: mutsu warned/died with the string instead of raising `X::Method::NotFound` like rakudo does.

This also lets `Math::FractionalPart`'s `die "FATAL: ...; please report this issue":` load at all (previously a hard parse error), which unblocks `Astro::Utils` and `DateTime::Julian` too (both reach the construct through `Math::FractionalPart`).

Closes #8141.

## Test plan

- [x] Added `t/routines/call/call-invocant-colon-listop.t`, pinning all five listops from the issue's table against expected behavior (verified against `raku` first).
- [x] `cargo fmt --all`
- [x] `make lint`
- [x] `make test` (4121 files, all green)
- [x] `make roast` — only the three documented container-environment failures (`roast/6.c/S32-io/file-tests.t`, `roast/S16-filehandles/filetest.t` — both `uid 0` chmod issues, and `roast/S32-io/IO-Socket-Async.t` — sandboxed network), matching `docs/agent-environments.md`'s known exceptions exactly.
- [x] Manually verified `Math::FractionalPart` now loads via `scripts/ecosystem-sweep.py --only Math::FractionalPart`.

https://claude.ai/code/session_01SsSEWrwCvPpagBfNjR5hFc

---
_Generated by [Claude Code](https://claude.ai/code/session_01SsSEWrwCvPpagBfNjR5hFc)_

### DIFF
--- a/news/2026-09/listop-invocant-colon-dispatches-a-method.md
+++ b/news/2026-09/listop-invocant-colon-dispatches-a-method.md
@@ -1,0 +1,51 @@
+# The listop invocant colon now dispatches a method everywhere
+
+Raku's listop invocant colon makes the first argument the invocant: `foo $x:
+@args` means `$x.foo(@args)`. mutsu's parser already built the right AST node
+for this (`CallArg::Invocant`, `Expr::MethodCall`) for the *non-tail*
+statement-level case, but several other places that compile a `Stmt::Call`
+either dropped the invocant marker and called the listop as a plain function,
+or hit `unreachable!()`:
+
+- `compile_tail_stmt_call_value` (shared by several tail-position call sites,
+  including a block/closure whose last statement is a call) panicked, because
+  its positional/named-pairs split has no arm for `CallArg::Invocant`.
+- `compile_routine_body_stmts` (a named `sub`'s own tail statement) and
+  `compile_closure_body_with_routine_flag`'s value-producing branch both
+  converted `CallArg::Invocant` into a plain positional argument, silently
+  dropping the colon.
+
+`die`/`fail` additionally had their own statement parser (`die_stmt`) and
+their own expression-position arm in `identifier_call.rs`, neither of which
+looked for a trailing `:` at all — `die "boom":` was a hard parse error
+(`fail`, `warn`, `say`, `sort`, `return` all *parsed*, since they go through
+the general listop-argument parser, but only the non-tail-statement case
+dispatched correctly).
+
+Fixed by adding one shared helper,
+`Compiler::invocant_colon_method_call`, that turns `(name, args)` into the
+equivalent `Expr::MethodCall` when `args` opens with an invocant, and wiring
+it into the three compiler call sites above; and by teaching `die_stmt` (and
+`die`/`fail`'s expression-position arm) to try the same no-paren
+invocant-colon parse `say`/`print`/`put`/`note` already use
+(`parse_io_colon_invocant_stmt`, renamed from an I/O-specific name since it is
+the general no-paren invocant-colon shape) before falling back to their own
+statement form.
+
+`say $x:`, `sort $x:` and `return $x:` already produced the right answer by
+coincidence — `Str`/`Array` really do have `.say`/`.sort` methods, and `.return`
+is a real control-flow method mutsu already implements, so dropping the colon
+and dispatching the method happened to agree. `warn $x:`/`die $x:` disagreed,
+because `Str` has no such method: mutsu warned/died with the string instead of
+raising `X::Method::NotFound` the way rakudo does. This let
+`Math::FractionalPart`'s
+`die "FATAL: ...; please report this issue":` load at all (previously a hard
+parse error), which also unblocks `Astro::Utils` and `DateTime::Julian`
+(both reach the construct through `Math::FractionalPart`).
+
+`t/routines/call/call-invocant-colon-listop.t` pins all five listops from the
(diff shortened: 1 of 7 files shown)


tokuhirom:
Closing as a duplicate: this issue (#8141) was already fixed and merged in #8161 (`818b5d9e`) while this PR was independently in flight — same root cause (the listop invocant colon being dropped instead of dispatching a method), same shared `Compiler::invocant_colon_method_call` helper approach, but #8161 also covers `fail`/`take`/`take-rw` and removes a separate bare-colon patch on `return` that this PR did not touch. No unique changes here worth carrying forward.

---

Issue #8087, opened 2026-09-12
`format!("__mutsu_...::{name}")`-built env keys: 286 sites, the same hotspot found in 4 separate perf campaigns, and 96 of them have a memoized replacement that already exists


tokuhirom:
Building a lookup key with `format!` and then probing `Env` with it has been the profiling finding
in at least **four separate perf campaigns**, each fixed locally, each time leaving the pattern
itself in place. It is back again as ~28% of the element-store cost in #8069. This issue is about
whether to keep treating it one call site at a time.

## The recurrence

| when | where | what the profile said |
| --- | --- | --- |
| 2026-08 | `news/2026-08/bench-ctor-map-compile-and-atomic-lane-probe.md` | the atomic-lane probe, per construction |
| 2026-09 | `news/2026-09/closure-call-hot-path-key-memoization.md` ([#7571](https://github.com/tokuhirom/mutsu/issues/7571)) | "rebuilding, on every single closure call, strings whose contents never change"; the lane key alone was ~1.5% of the RIPEMD run in `core::fmt` machinery |
| 2026-09 | `news/2026-09/typed-lexical-metadata-probes-take-symbols.md` | `__mutsu_type::<name>` probed up to **7 times per store**; ~7.2k instructions per vendored-`Test` assertion |
| 2026-09 | `news/2026-09/bench-ctor-name-rederivation.md` | `Symbol::intern` **732,232 calls for 5000 constructions — 146 name interns per object** |
| 2026-09 | #8069 | **22 interns and 8 allocations per `@a[$i] = $v`**, ~28% of a 14,053-instruction store |

Five findings, one mechanism.

## The state of the code

```
286  format!("__mutsu_…")  sites in src/
 36  distinct magic-key namespaces
```

The largest namespaces: `__mutsu_sigilless_alias::` (64 mentions), `__mutsu_sigilless_readonly::`
(43), `__mutsu_type::` (37), `__mutsu_atomic_arr::` (36), `__mutsu_callable_id::` (34),
`__mutsu_atomic_hash::` (19), `__mutsu_bound::` (14), `__mutsu_hash_key_type::` (13).

**The most damning part: the memoized replacements already exist.** `runtime/runtime_var_meta.rs`
and `runtime/shared_store.rs` between them carry seven `*_key_for_sym` / `atomic_lane_key` helpers —
`type_meta_key_for_sym`, `hash_key_meta_key_for_sym`, `callable_id_key_for_syms`,
[...]

Issue #8085, opened 2026-09-12
The bench series' own noise floor (median 16-33% per commit) is larger than its 1.5x regression alarm — add a deterministic instruction-count series


tokuhirom:
`bench.yml` records wall clock on a shared GitHub runner and warns when a benchmark's median
slows >1.5x vs the previous recorded run. Measured against the history the job has actually
produced, **that alarm sits below the series' own noise floor**, so it cannot do the job it was
added for.

## The noise, measured

For each series in `bench-history.tsv`, the median and p90 of `|change|` between **consecutive
recorded main commits** — i.e. the step a reader has to learn to ignore before a real regression is
visible. Last 40 recorded commits:

| series | median \|Δ\| (absolute s) | p90 \|Δ\| (absolute s) | median \|Δ\| (ratio) |
| --- | ---: | ---: | ---: |
| `bench-ctor` | 31.7% | 96.1% | 25.5% |
| `method-call` | 30.8% | 57.7% | 14.5% |
| `bench-mandelbrot` | 28.1% | 43.8% | 7.7% |
| `bench-fib` | 23.3% | 43.7% | 8.1% |
| `bench-hash` | 23.8% | 41.4% | 6.2% |
| `bench-array` | 16.4% | 32.8% | 8.3% |
| `bench-threads` | 15.2% | 40.9% | 7.6% |

Every one of the 48 series measures a median step of **16-33%** and a p90 of **29-96%**. The
mutsu/raku ratio column — the thing the workflow header says "normalizes runner speed" — is better
but still a 5-25% median step, and for `bench-ctor` it is 25%.

Consequences:

- **A 1.5x alarm is inside the p90 of most series.** It fires on noise, and a genuine 30%
  regression — which is a very large regression — never trips it at all.
- The `+X%` column in the job summary is, for most rows, unreadable.
- Practically, a slowdown is only noticed when someone profiles for an unrelated reason. That is
  how the `@a[$i] = $v` finding in #8069 stayed invisible.

Reproduce with the script in §"Reproducing" below.

## Callgrind instruction counts are ~1500x more sensitive, and the suite fits in 4 minutes

Three back-to-back callgrind runs of the same binary on the same benchmark:

```
bench-hash            269,436,183 / 269,406,388 / 269,437,654   spread 0.012%
bench-grammar-parse    50,299,105 /  50,328,170 /  50,282,570   spread 0.09%
```

[...]

Issue #8119, opened 2026-09-12
A role's `::?CLASS:U:` multi candidate is dropped when the role is composed


tokuhirom:
Found while fixing #8061: with `TAP` loading, `t/source-file.rakutest` dies here.

## Minimal repro

```raku
role R {
    proto method f(|) { * };
    multi method f(::?CLASS:U: Str:D $s) { "U-invocant: $s" }
}
class C does R {
    multi method f(::?CLASS:D: Str:D $s) { "D-invocant: $s" }
}
say C.f('x');
say C.new.f('y');
```

raku:

```
U-invocant: x
D-invocant: y
```

mutsu:

```
Cannot resolve caller f(C:U: Str:D); none of these signatures matches:
    (C:D $self:: Str:D $s, *%_)
```

The class's own `:D`-invocant candidate is the *only* one in the list: the role's `:U`-invocant
candidate did not survive composition. A `:D` candidate declared in the role body does survive, so
this is specific to the undefined-invocant smiley on a composed candidate, not to role multis in
general.

## Where it comes from

`TAP::SourceHandler` (`TAP` 0.3.15) is the shape:

```raku
role SourceHandler {
    proto method make-source(|) { * };
    multi method make-source(IO:D $path, IO:D() :$cwd = $path.CWD, *%args) {
        self.make-source($path.relative($cwd), :$cwd, |%args);
    }
    multi method make-source(::?CLASS:U: Str:D $name, *%args) {   # <- lost
        self.new.make-source($name, |%args);
    }
    ...
}
class SourceHandler::File does SourceHandler {
    multi method make-source(::?CLASS:D: Str:D $name, IO:D() :$cwd = $*CWD, *%) { ... }
}
```

`TAP::SourceHandler::File.make-source('foo.tap', :cwd(...))` is called on the *type object*, so the
role's `:U` candidate is the one that should take it (and then re-dispatch on `self.new`). mutsu
reports only the two candidates that did compose:

```
Cannot resolve caller make-source(TAP::SourceHandler::File:U: Str:D, :cwd(Str)); none of these signatures matches:
    (TAP::SourceHandler::File $:: IO:D $path, IO:D() :$cwd, Any *%args, *%_)
    (TAP::SourceHandler::File:D $self:: Str:D $name, IO:D() :$cwd, Any *%__ANON_HASH__, *%_)
```

## Affected distributions

[...]

PR #8176, opened 2026-09-12
perf(runtime): move every memoized-namespace `__mutsu_*` key site onto MetaNs

tokuhirom:
Stage 2 of #8087.

Stage 1 (#8104) built `MetaNs` — one memoizing constructor for `__mutsu_<ns>::<name>` env keys, with no public string form — and wired `scripts/check-magic-keys.sh` as a per-file ratchet over the **276** hand-built `format!` sites that remained. 102 of those were in namespaces whose memoized replacement already existed, which is exactly why the pattern kept coming back in profiles: five separate perf campaigns each fixed only the sites its own profile walked through.

This converts all 102. Every namespace that had a memoized helper is now at **zero** hand-built sites across `src/`:

| namespace | before | after |
| --- | ---: | ---: |
| `__mutsu_sigilless_alias::` | 24 | 0 |
| `__mutsu_sigilless_readonly::` | 20 | 0 |
| `__mutsu_atomic_arr::` | 17 | 0 |
| `__mutsu_callable_id::` | 12 | 0 |
| `__mutsu_atomic_hash::` | 9 | 0 |
| `__mutsu_bound_index::` | 6 | 0 |
| `__mutsu_shaped_array_dims::` | 4 | 0 |
| `__mutsu_type::` | 3 | 0 |
| `__mutsu_bound::` | 3 | 0 |
| `__mutsu_hash_key_type::` | 2 | 0 |
| `__mutsu_state_key::` | 2 | 0 |

## Not only "call the helper"

Calling the constructor removes the `format!`, but a site that then probes the env *by name* still hashes the 30-odd-byte key on every access. So each converted site holds the key as a pre-interned `Symbol` and probes with `get_sym` / `contains_key_sym` / `remove_sym` / `get_mut_sym` — neither the string nor its hash is rebuilt. `runtime::utils`'s `sigilless_alias_key` / `sigilless_readonly_key` return a `Symbol` for that reason, and `types::signature`'s duplicate pair of the same two functions is re-exported from there instead of being a second definition of the key shape.

Two memo tables that had grown up beside `MetaNs` fold into it:

- `shared_store::atomic_lane_key`'s own pair of thread-locals becomes the `AtomicArr` / `AtomicHash` namespaces on the one table, with `atomic_lane_str_key` as the `&'static str` form the by-name shared store needs.
- `callable_id_key_for_syms`'s `(package, name)` table becomes `MetaNs::key_pair` — the general two-part form; `CallableId` is the one namespace not keyed by a single name.

## The hazard, and how it is tested

`Env::insert_sym` deliberately does **not** run `note_env_key`, the latch that arms the "could a key of this family exist?" probes in `src/env.rs` — that latch is what lets a program with no sigilless binding skip the readonly probe on every assignment. A name-derived write landing on `insert_sym` therefore stores the metadata and leaves the reader permanently switched off: no error, just a constraint or an alias silently never found again.

Every converted write goes through `insert_sym_noting`. `t/vm/binding/var-metadata-key-namespaces.t` makes that testable rather than asserted — 23 cases driving each namespace end-to-end through behaviour that fails if the key is written but not read back (sigilless term refusing assignment, two-hop `:=` alias chain, bound element surviving a later plain element store, shaped-array dimensions, typed lexical and object hash refusing a wrong-typed store, a `state` cell per closure clone, same-named subs in two packages keeping separate state, a raw parameter writing back to its caller, concurrent pushes landing in a shared array and hash). All 23 pass under rakudo too. `meta_ns.rs`'s unit tests spell out the exact key every namespace produces, so a prefix typo is caught by `cargo test`.

## Measured

Deterministic instruction counts (`scripts/bench-det.sh`, #8085 — callgrind simulation, ~0.1% resolution), against a same-toolchain `origin/main` release binary:

| benchmark | JIT off | JIT on |
| --- | ---: | ---: |
| `bench-threads-serial` | **−3.71%** | **−4.15%** |
| `bench-threads` | **−2.92%** | **−3.22%** |
| `bench-ctor` | −0.30% | −0.30% |
| `bench-index-store` | −0.07% | −0.05% |
| `bench-hash` | −0.04% | −0.03% |
| `bench-grammar-parse` | +0.04% | −0.16% |

The concurrency rows are the atomic-lane keys: once any `start` block pushes to a shared array the lane probe is armed for the rest of the process, and every `@`/`%` access after that was building its lane key from scratch. `bench-grammar-parse` is flat at the noise floor, which is the expected shape — a program with no sigilless binding, shaped array or atomic lane has its probes latched off and never builds these keys at all. Nothing regressed.

## What is left

The ratchet is re-cut at **174**, in the namespaces with no memoized constructor yet (`__mutsu_atomic_name::`, `__mutsu_deleted_index::`, `__mutsu_bound_decont::`, `__mutsu_shared_dirty::`, `__mutsu_ro_index::`, `__mutsu_elem_share::`, …). Four of the old 276 were not sites at all but doc comments quoting the `format!` they described; those are reworded, so every number the ratchet reports is now a real construction.

Stages 3 and 4 stay open on #8087.

## Verification

- `make lint` — clean (all four configurations)
- `make test` — 4102 files, 43669 tests, all pass; `cargo test` 1128 pass
- `make roast` — 1437 files, 219635 tests; the only failures are the three documented container-only ones (`6.c/S32-io/file-tests.t`, `S16-filehandles/filetest.t` — `uid 0`; `S32-io/IO-Socket-Async.t` — sandboxed network), per `docs/agent-environments.md`
- `make check-magic-keys` — ok (174 sites, baseline 174)

Refs #8087

https://claude.ai/code/session_01QoRrJgwXv3pryDCnR3aYKS

---
_Generated by [Claude Code](https://claude.ai/code/session_01QoRrJgwXv3pryDCnR3aYKS)_


tokuhirom:
**`test-suites` was failing, and it was not this PR's.** Pushed the fix anyway, because `main` itself is red and nothing else was going to unblock it.

## The failure

`t/oo/role/class-does-enumeration-role.t`, exit 255, "planned 18 ran 15" — it aborts mid-file at test 16:

```
Ambiguous call to 'gist(DNA: )'; these signatures all match:
  (DNA $:: *%_)
  (DNA:D $self:: *%_)
```

Not a flake: 3/3 locally on this branch. And **not this PR's** — it reproduces identically at `b3d2ede7`, this PR's merge base, built with neither of this branch's commits. `b3d2ede7..origin/main` is empty, so that is `main`'s current head: `main` is red.

## Root cause: two correct PRs, merged three minutes apart

Same shape as the `local_attr_key` build break earlier today — neither CI run could see the other.

- **#8115** (`7a73d263`) made `Enumeration` a composable core role supplied by `ENUMERATION_ROLE_PRELUDE`, and added this test. Its derived methods (`kv`, `pair`, `Numeric`, `Int`, `Real`, `gist`, `raku`) were declared with **no invocant type**.
- **#8119** (`7bcad758`) fixed the opposite bug: the "a class multi replaces a matching role multi" dedup compared candidates through `method_positional_signature`, which strips the invocant, so a role's `:U:`-invocant candidate was dropped as a false duplicate of a class's `:D:`-invocant one. It now requires the invocant's own type constraint to match.

Composed, the role's constraint-free `gist` stops being deduped against the documented `class DNA does Enumeration`'s `multi method gist(::?CLASS:D:)`; both candidates survive composition, and mutsu reports ambiguity where rakudo picks the `:D` one as narrower.

## The fix carried here

The prelude is the side that was wrong. Rakudo's `Enumeration` declares **no `gist` at all** — `Enumeration.^methods(:local)` is `new enums pair CALL-ME COERCE pred succ clone` — and every instance method it does declare carries the smiley (`pair`, `pred`, `succ` are all `:(::?CLASS:D $:: *%_)`). mutsu's prelude supplies a deliberately larger set (#8115 measured which ones actually work on a composing class), but it was giving them a shape rakudo does not use.

All seven now take `::?CLASS:D:`. That is both the faithful spelling and what lets #8119's invocant-aware dedup do its job: a composing class's `:D:` override replaces the role's candidate instead of competing with it. It closes the whole class of collision, not only `gist` — a class overriding `raku` or `Numeric` the same documented way would have hit it too.

It no-ops the moment `main` carries an equivalent fix.

## Verified on the repaired tree

- `t/oo/role/class-does-enumeration-role.t` 18/18 **and** #8119's own pin `t/oo/role/role-multi-invocant-smiley-across-composition.t` — so neither fix undoes the other.
- `make test` — green, 4135 files / 44017 tests.
- `make lint` — green, all four configurations.
- `make roast` — the three container-only failures `docs/agent-environments.md` records (`6.c/S32-io/file-tests.t`, `S16-filehandles/filetest.t` under `uid 0`; `S32-io/IO-Socket-Async.t` sandboxed network), **plus** one chance firing of `roast/S02-types/sethash.t`'s self-described *"only a 1/40_320 chance of that test failing by chance alone"* `SetHash.pick(*)` ordering assertion. It passed 8/8 on re-run and cannot involve this change: that file never mentions `Enumeration`, so the prelude is not injected into it. Not quarantining it — one chance firing of a test that documents its own odds is not evidence of a defect.

---

tokuhirom:
Follow-up to my comment above: **`main` has since fixed this itself, so the carried fix is dropped.**

The upstream fix (`src/runtime/run.rs`, now on `main`) makes the prelude's `gist` a `multi method gist(::?CLASS:D:)`. Mine had given all seven derived methods a plain `::?CLASS:D:` invocant. Both resolve the ambiguity; theirs is on `main`, so I rebased onto it and `git rebase --skip`ed my commit rather than re-litigating the shape. My news entry went with it.

Branch force-pushed to `ae59b6a8`, rebased onto `fe54ea58`. Re-verified on that base, since `main` had moved by 24 commits:

- `make test` — green, 4136 files / 44033 tests.
- `make roast` — the failing set is exactly the three `docs/agent-environments.md` records as container-only (`6.c/S32-io/file-tests.t`, `S16-filehandles/filetest.t` under `uid 0`; `S32-io/IO-Socket-Async.t`, sandboxed network). The `S02-types/sethash.t` chance firing I mentioned did not recur.
- `make lint` — green, all four configurations.
- `make check-magic-keys` — ok (174 sites, baseline 174).
- Spot-checked the tests either side of the collision on the new base: `t/oo/role/class-does-enumeration-role.t`, #8119's pin `t/oo/role/role-multi-invocant-smiley-across-composition.t`, `main`'s new `t/collections/element-cell-gist-and-str.t`, and this PR's own `t/vm/binding/var-metadata-key-namespaces.t` — 63 assertions, all pass.

One observation I am **not** acting on here, since it is out of this PR's scope: the upstream fix covers `gist`. The prelude's other six derived methods (`kv`, `pair`, `Numeric`, `Int`, `Real`, `raku`) still declare no invocant type, so a composing class that overrides any of them with the documented `::?CLASS:D:` invocant — which is how `raku-doc`'s own example writes `gist` — would hit the same ambiguity through the same #8119 dedup path. Worth a ticket against the `Enumeration` prelude if someone wants it closed properly.

---

PR #8260, opened 2026-09-13
fix(vm): raise on deep recursion instead of aborting the process

### LINKED ISSUE #8232: Deep recursion aborts the process on a Rust stack overflow instead of raising a catchable error
Found while making upstream `JSON::Fast` run ([#8226](https://github.com/tokuhirom/mutsu/issues/8226)): its `t/01-parse.t` feeds the parser `Q«[{"":» x 10_000`, which recurses ~20,000 Raku routines deep. rakudo raises a catchable `X::AdHoc`; mutsu **aborts the process**, so the rest of the test file never runs.

## Repro

No module needed:

```raku
my $d = 0;
sub rec($n) { $d = $n; rec($n+1) unless $n >= 20000; }
END { note "depth reached: $d" }
rec(1);
```

- `raku` → `depth reached: 20000`
- `mutsu` → `thread 'mutsu-main' (…) has overflowed its stack` / `fatal runtime error: stack overflow, aborting` — and **no `END` phaser, no backtrace, no catchable exception**, because the process is killed by the guard page.

Bisected on a debug build: 5,000 frames is fine, 10,000 already aborts.

The JSON::Fast shape, for the record:

```console
$ mutsu -I <JSON-Fast-0.20.1>/lib -e 'use JSON::Fast; try { from-json(Q<[{"":> x 10_000) }; say "survived"'
thread 'mutsu-main' (…) has overflowed its stack
fatal runtime error: stack overflow, aborting

$ raku  -I <JSON-Fast-0.20.1>/lib -e 'use JSON::Fast; try { from-json(Q<[{"":> x 10_000) }; say "survived"'
survived
```

## Root cause

Every Raku-level call consumes Rust stack — the VM's `exec_one` dispatch frame, the call-op handler, signature binding, and for a re-entrant construct a whole `eval_block_value` frame. `src/main.rs:149` already asks for a 256 MB stack for the `mutsu-main` thread, so the ceiling is not a missing `stack_size`: it is ~30–50 KB of Rust stack per Raku call in a debug build. When it is exhausted the thread hits its guard page and the Rust runtime aborts — there is no unwinding, so nothing mutsu can `catch`, and `try`/`CATCH` cannot help.

The two `.stack_size(16 * 1024 * 1024)` spawns in `src/runtime/mod.rs:4622,4653` (user-visible threads) are a 16 MB ceiling, i.e. ~16× lower again, so a recursive routine inside a `start`/`Thread` dies far sooner than one on the main thread.

## Why it is large

The honest fix is a **recursion-depth guard that raises**, not a bigger stack — a bigger stack only moves the cliff, and it cannot be sized correctly because the per-call cost depends on the construct (a plain compiled call is cheap; a re-entrant `eval_block_value` is not). That means:

- picking where the counter lives and is checked (one chokepoint that every routine entry passes, not one per call op) and proving it has no measurable cost on the bench workloads;
- deciding the limit and whether it is configurable, and what rakudo's own behaviour is for comparison (rakudo has no fixed limit — it grows the call stack on the heap and fails with an ordinary exception when memory runs out, which mutsu cannot copy without moving frames off the Rust stack);
- the error: rakudo does not have a dedicated `X::Recursion`, so the exception type and message need deciding;
- auditing the spawned-thread stack sizes against whatever limit is chosen, so a thread does not die at a *different* depth than the main thread.

Reducing the per-call Rust stack cost is a worthwhile separate lever (it raises the real ceiling for everyone, not just the failure mode), but it is a perf/architecture question and does not remove the need for a guard.

## Impact

- `roast`: any spec test that checks deep recursion behaves, or that merely recurses deeply as a means to an end, takes the whole file down rather than failing one assertion.
- `JSON::Fast`'s `t/01-parse.t` — one of the two blockers keeping the real distribution from being vendored and the native provider from being deleted (#8226, ADR-0096 §D4/E2).
- Any user program whose input depth is attacker-controlled: a deeply-nested document is an unconditional process abort, which a `try` cannot contain.

tokuhirom:
```raku
my $d = 0;
sub rec($n) { $d = $n; rec($n+1) unless $n >= 20000 }
END { note "depth reached: $d" }
rec(1);
```

printed `fatal runtime error: stack overflow, aborting` and nothing else — no
`END` phaser, no backtrace, no exit status a script could act on, and nothing
for `try` or `CATCH` to contain. It now prints `depth reached: 5751` and raises
an ordinary catchable exception.

## Root cause

mutsu executes a Raku call as a **Rust** call: the `exec_one` dispatch frame,
the call-op handler, signature binding, and for a re-entrant construct a whole
`eval_block_value` frame all sit on the native stack, with the callee's body
loop running nested inside them. Raku recursion is native recursion, so running
out of it hits the thread's guard page and the Rust runtime aborts — there is no
unwinding, which is why no amount of defensive `try` helped. That matters most
for a program whose input depth is attacker-controlled, where one deeply nested
document is an unconditional process kill.

The 256 MB stack `main.rs` already asks for was never the fix: it moves the
cliff without removing it.

## The guard

Recorded as **ADR-0100** (new, `Accepted (implemented)`), which also writes down
what was rejected and why.

Each thread that runs user VM code notes, at start-up, the lowest stack address
it may still make a call from — its stack top, minus its size, plus a 16 MiB
reserve. The VM's seven call entry points (the ones that already carry a
`SafepointKind::Call` GC safepoint or push a call frame) compare against it and
raise `X::AdHoc` with `Too deep recursion (out of stack space)` rather than
proceed. Verified to cover sub, method, closure and block recursion, with JIT on
and off, and inside a `start {}` worker.

**The measurement is the stack pointer, not a frame count**, because the
per-call cost is not a constant — measured ~42 KiB for a JIT-on debug frame and
~156 KiB for an interpreted one, so a depth limit generous enough for one
program would be unsafe for another. The stack pointer needs no maintenance
because it *is* the state.

`X::AdHoc` because rakudo has no dedicated `X::Recursion` to match (it grows its
call stack on the heap and fails with an out-of-memory exception instead), and a
mutsu-only type would make the obvious generic `CATCH` miss it.

## Arming is a claim that has to be checkable

`init_thread_stack_floor(stack_size)` takes the caller's word for its stack
size, and a wrong word must leave the thread **unguarded** rather than arm it
into an already-exhausted state. Saturating the subtraction did the opposite,
which took `wasm-e2e` down on the first push of this branch: `usize` is 32 bits
on wasm32 and the stack pointer sits a megabyte or two into linear memory, so

```rust
approx_stack_pointer()                    // ~1_000_000
    .saturating_sub(stack_size)           // 256 MiB -> clamps to 0
    .saturating_add(STACK_RESERVE_BYTES)  // -> 16_777_216
```

placed the floor **above** the whole wasm stack and the very first check fired.
15 of 28 `site/concurrency.test.mjs` cases died on code as shallow as
`my $p = start { die "boom" }`.

Two fixes, both in the second commit:

1. `checked_sub` with a `floor < top` postcondition — an impossible claim now
   declines to arm, which is what this module already documented.
2. wasm does not arm at all. `thread_compat::spawn_thread` queues the closure
   on the single browser thread rather than creating one, so the
   `Some(USER_THREAD_STACK_SIZE)` it was handed described no stack that exists.

Pinned by `a_stack_size_bigger_than_the_address_below_us_leaves_the_thread_unguarded`,
confirmed to fail on the pre-fix arithmetic by reverting the hunk and re-running.
On a 64-bit host the bogus floor lands far *below* `sp` rather than above it, so
the `assert_eq!(floor, 0)` is what catches it there.

## Performance

Reading the stack pointer at every call boundary measured **~3–4% slower** on
`bench-fib` and `method-call`, because taking the address of a local forces a
stack slot and acts as an optimization barrier in the hottest functions mutsu
has. So the hot path is now a countdown on an `Interpreter` field and the real
read happens once per 32 calls; the 16 MiB reserve is sized to absorb a whole
interval of the fattest frames (~5 MiB) on top of the unwind and `END` phasers.

The residual is ~2–4% on the two call-only microbenchmarks and flat on
`bench-tak`. Two controls establish that it is **not** the countdown
arithmetic: a build with the guard's *body* replaced by `Ok(())` and the call
sites unchanged measured the same delta, and two independently built baselines
measured the same as each other. It is the per-call-site overhead itself, and
there is no cheaper arrangement of the same check. The authoritative numbers are
the bench CI's, not these locals. Paying it removes an uncatchable process
abort, which is the trade `CLAUDE.md`'s gain/risk definition asks for.

## Two corrections to the issue

- The two `stack_size(16 * 1024 * 1024)` spawns #8232 names in
  `src/runtime/mod.rs` are `#[test]` helpers, not user-visible threads.
  `start`/Promise/Supply workers already get the same 256 MB stack as
  `mutsu-main` (`builtins_system::USER_THREAD_STACK_SIZE`), so there was no 16×
  lower ceiling for threaded user code to audit — the guard fires at comparable
  depth everywhere by construction.
- The limit is deliberately **not** configurable: a recursion-depth setting
  would be a promise mutsu cannot keep, for the same non-constant-cost reason
  above. The stack size is the knob.

## Known gap

A thread mutsu does not own — an `Interpreter` driven from a `#[test]`, or the
wasm library build — is left unguarded, since there is no portable way to
discover its stack bound (`pthread_getattr_np` is glibc-only and `libc` is an
optional dependency here). That is today's behaviour, unchanged; the new public
`mutsu::arm_stack_guard(stack_size)` lets an embedder opt in.

## Test

`t/vm/deep-recursion-raises-instead-of-aborting.t`. The depth the guard fires at
is a property of the build (measured ~5,700 frames debug+JIT, ~1,500 debug
without), so nothing there asserts a depth — only that the failure is catchable,
that `try` contains it, and that execution continues afterwards. ~5.5s under
both the gc-stress and jit-stress environment settings.

Plus the four `vm_stack_guard` unit tests, including the wasm-shape one above.

## Rebase note

`main` landed its own ADR-0099 (`0099-regex-engine-performance-strategy.md`)
while this was open, so this one was renumbered **0099 → 0100** during the
rebase, file and every reference. `docs/adr/README.md` carries both rows.

## Verification

- `cargo fmt --all`, `make lint` (all four configurations) — clean, re-run after
  the rebase.
- `make test` — 4190 files, 44636 tests, all pass.
- `make roast` — the only failures are the three documented container-only ones
  (`6.c/S32-io/file-tests.t` `Failed: 4` and `S16-filehandles/filetest.t`
  `Failed: 25` run as `uid 0`, `S32-io/IO-Socket-Async.t` exit 124 sandboxed
  network), per `docs/agent-environments.md`.
- `wasm-pack` build + `node site/concurrency.test.mjs` locally: **28 passed, 0
  failed**, against CI's 13 passed / 15 failed on the first commit.

Closes #8232

https://claude.ai/code/session_01AMQwQoZeiJDBM5iK3hvPsK

### DIFF
--- a/docs/adr/0100-deep-recursion-raises-on-native-stack-headroom.md
+++ b/docs/adr/0100-deep-recursion-raises-on-native-stack-headroom.md
@@ -1,0 +1,168 @@
+# ADR-0100: Deep recursion raises a catchable error, guarded by native stack headroom
+
+- Status: Accepted (implemented)
+- Date: 2026-09-13
+- Issue: [#8232](https://github.com/tokuhirom/mutsu/issues/8232)
+
+## Context
+
+mutsu executes a Raku routine call as a **Rust** call: the VM's `exec_one`
+dispatch frame, the call-op handler, signature binding, and — for a re-entrant
+construct — a whole `eval_block_value` frame all sit on the native stack, and
+the callee's body loop runs nested inside them. Raku recursion is therefore
+native recursion.
+
+When the native stack runs out, the thread hits its guard page and the Rust
+runtime **aborts the process**. There is no unwinding, so:
+
+- `try` / `CATCH` cannot contain it,
+- `END` phasers do not run,
+- there is no backtrace and no exit code a script can act on,
+- and the rest of a test file after the offending statement never runs.
+
+```raku
+my $d = 0;
+sub rec($n) { $d = $n; rec($n+1) unless $n >= 20000 }
+END { note "depth reached: $d" }
+rec(1);
+```
+
+rakudo prints `depth reached: 20000`; mutsu printed
+`fatal runtime error: stack overflow, aborting` and nothing else.
+
+This is not only a diagnostics problem. Any program whose input nesting depth
+is attacker-controlled — a JSON or XML document, a recursive-descent parser
+over user text — turns a deeply nested input into an unconditional process
+abort that no amount of defensive `try` can contain. `JSON::Fast`'s
+`t/01-parse.t` feeds its parser `Q«[{"":» x 10_000` (≈20,000 Raku frames)
+precisely to check that this is survivable.
+
+rakudo has no fixed recursion limit: it grows the call stack on the **heap**
+and eventually fails with an ordinary, catchable out-of-memory exception.
+mutsu cannot copy that without moving call frames off the Rust stack, which is
+a VM redesign (see "Rejected alternatives").
+
+## Decision
+
+**Guard on remaining native stack headroom at the VM's call boundary, and
+raise an ordinary catchable Raku exception when it is exhausted.**
+
+1. **The measurement is the stack, not a frame count.** Each thread that runs
+   user VM code records, at start-up, the lowest address it may safely touch
+   (`stack top − stack size + reserve`). At a call boundary the VM takes the
+   address of a stack local and compares. That is one load, one subtract and
+   one branch, with no counter to keep coherent across the several call paths
+   and no state to unwind.
+
+2. **The chokepoints are the VM's call entry points.** The seven functions
+   that carry a `SafepointKind::Call` GC safepoint or push a call frame —
+   `call_compiled_function_fast`, the two light paths,
+   `call_compiled_function_named_inner`, `call_compiled_closure_in_unit`, and
+   the two `call_compiled_method*` entries. All seven already return
+   `Result<_, RuntimeError>`, and all seven are already established as points
+   that hold no container borrow, which is what a raise needs. Between them
+   they cover every shape of Raku recursion: sub, method, closure/block, and
+   both JIT-on and JIT-off dispatch.
+
+3. **The check is amortized over 32 calls.** Reading the stack pointer needs
+   the address of a local, which forces a stack slot and acts as an
+   optimization barrier in the hottest functions mutsu has. The hot path is
+   therefore a countdown on an `Interpreter` field, and the real read happens
+   once per `STACK_CHECK_INTERVAL` calls. The reserve absorbs the overshoot.
+
+4. **The reserve is 16 MiB of a 256 MiB stack (6%).** The guard fires while
+   there is still room for three things: an interval's worth of overshoot
+   (32 frames of the fattest kind measured — ~156 KiB each, an interpreted
+   non-JIT frame in a debug build — is ~5 MiB), the native recursion a
+   *single* Raku call can perform between two checks (a deep regex match,
+   dropping a deeply nested value), and then constructing the error,
+   unwinding, and running the `END` phasers the unwind reaches.
+
+5. **The error is `X::AdHoc`**, message
+   `Too deep recursion (out of stack space)`. rakudo has no dedicated
+   `X::Recursion` to match, and inventing a mutsu-only type would make the
+   obvious `CATCH { when X::Whatever { } }` miss it. `X::AdHoc` is what every
+   generic handler already catches, and the message is what a user sees.
+
+6. **Not configurable.** The knob that matters is the stack size itself, and
+   it is already uniform (below). A recursion-depth setting would be a promise
+   mutsu cannot keep, because the per-call stack cost is not a constant: a
+   plain compiled call is cheap and a re-entrant `eval_block_value` is not, so
+   the same "depth" buys different amounts of stack in different programs.
+
+7. **Every thread that runs user VM code already has the same 256 MiB stack**,
+   so the guard fires at comparable depth everywhere. `main.rs` spawns
+   `mutsu-main` with `stack_size(256 MiB)` and
+   `builtins_system::USER_THREAD_STACK_SIZE` (the `start` / Promise / Supply
+   worker stack) is the same 256 MiB constant. The two
+   `stack_size(16 * 1024 * 1024)` spawns in `src/runtime/mod.rs` are
+   `#[test]` helpers, not user-visible threads — #8232 read them as a 16×
+   lower ceiling for user code, and that part of the report is wrong. Service
+   threads (timer, socket pump, signal reader) keep the default stack because
+   they run no user VM code.
+
+## Consequences
+
+- Deep recursion now raises where it used to abort. `try`, `CATCH`, `END` and
+  a normal exit status all work, and a test file continues past the failure.
+- **It costs about 3% on a call-only microbenchmark.** Local A/B on this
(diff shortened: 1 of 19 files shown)


tokuhirom:
`wasm-e2e` is red on `28f1c40` — **15 of 28** `site/concurrency.test.mjs` cases fail, and it is this PR's, not noise. Trivial code dies:

```
FAIL: caught: boom
  code:     my $p = start { die "boom" }; try { await $p; CATCH { default { say "caught: " ~ .message } } }
  expected: "caught: boom"
  got:      "caught: Too deep recursion (out of stack space)\n"
```

## Root cause: the arming arithmetic saturates on wasm32

`usize` is **32 bits** on wasm32, and the stack pointer sits a megabyte or two into linear memory. `init_thread_stack_floor(256 MiB)` then computes

```rust
approx_stack_pointer()          // ~1_000_000
    .saturating_sub(stack_size) // 256 MiB  -> clamps to 0
    .saturating_add(STACK_RESERVE_BYTES) // -> 16_777_216
```

a floor of 16 MiB — **above** the entire wasm stack. `floor != 0 && sp < floor` is therefore true at the first check, so the guard refuses the first call every `start {}` block makes.

The PR's own "Known gap" section says the wasm library build is left unguarded, but `builtins_system.rs:82` arms it anyway: on wasm `thread_compat::spawn_thread` queues the closure on the single browser thread rather than creating one, while `stack_size` is still `Some(USER_THREAD_STACK_SIZE)`, so the arming call fires with a size that describes no stack that exists.

## Fix

Two changes, both pushed here:

1. **`init_thread_stack_floor` refuses to arm into an already-exhausted state.** `checked_sub` instead of `saturating_sub`, with a `floor < top` postcondition. A wrong `stack_size` now leaves the thread unguarded — which is what the module documents — instead of computing a floor above the stack pointer. This is the general defect: saturation turned "this claim is impossible" into "you are out of stack".
2. **wasm does not arm at all**, matching the module docs and `spawn_thread`'s own comment that the size is meaningless there.

## Verification

- New unit test `a_stack_size_bigger_than_the_address_below_us_leaves_the_thread_unguarded` pins the arithmetic. Confirmed it **fails on the pre-fix code** (reverted the hunk, re-ran, `FAILED`) and passes after — the `assert_eq!(floor, 0)` is what catches it on a 64-bit host, where the bogus floor is merely far below `sp` rather than above it.
- Built the npm package with `wasm-pack` and ran `node site/concurrency.test.mjs` locally: **28 passed, 0 failed** (CI's run on the parent commit: 13 passed, 15 failed).
- `cargo fmt --all` and `make lint` clean; `make test` and `make roast` re-running before this lands.

---

Issue #8233, opened 2026-09-13
`++$p` as an argument to an `int $p is rw` parameter is refused ("expects a writable container")


tokuhirom:
Found while making upstream `JSON::Fast` run ([#8226](https://github.com/tokuhirom/mutsu/issues/8226)). Its whitespace scanner hands a pre-incremented native-int parameter straight to another `is rw` parameter:

```raku
my sub nom-ws(str $text, int $pos is rw --> Nil) {
    ...
    nqp::if(
      nqp::iseq_i(nqp::ordat($text,$pos),47),  # /
      nom-comment($text,++$pos)                #  <-- here
    );
}
my sub nom-comment(str $text, int $pos is rw --> Nil) {
    unless $*ALLOW-JSONC {
        --$pos;  # un-eat the /
        return;
    }
    ...
}
```

`nom-comment`'s `--$pos` has to reach `nom-ws`'s `$pos` and from there `from-json`'s, or `from-json` concludes it consumed the whole string and accepts trailing garbage.

## Repro

```raku
sub inner(int $p is rw) { --$p }
sub outer(int $p is rw) { inner(++$p) }
my int $pos = 5;
outer($pos);
say $pos;
```

- `raku` → `5` (the `++` and the `--` cancel: the native reference is bound through)
- `mutsu` → dies:
  ```
  Parameter '$p' expects a writable container (variable) as an argument,
  but got '6' (Int) as a value without a container.
    in sub outer at … line 2
  ```

Passing the parameter *without* the increment is fine in both (`inner($p)` → `4`), so the binding machinery works; it is specifically the `++$p` / `--$p` argument shape that loses the container.

Note the narrow scope: rakudo accepts this only when the operand is itself a **native** `is rw` parameter. At mainline, `my int $pos = 5; inner(++$pos)` is an error in rakudo too (`Expected a modifiable native int argument for '$p'`), and a non-native `$p is rw` with `++$p` is an error in rakudo as well (`Cannot assign to a readonly variable or a value`). So the fix should not make every `++$x` argument container-binding — only the native-reference case rakudo allows.

The user-visible consequence in `JSON::Fast`:

```console
$ mutsu -I <JSON-Fast-0.20.1>/lib -e 'use JSON::Fast; say from-json(q<{"a":"b"}/>).raku'
{:a("b")}                      # trailing "/" silently accepted

[...]

Issue #8226, opened 2026-09-13
Vendor the real JSON::Fast and delete the native provider — the gap is 9 `nqp::` ops, not the recorded ~50


tokuhirom:
[#8183](https://github.com/tokuhirom/mutsu/issues/8183) retired the JSON `use`-time interception and closed with item 3 — *deciding `JSON::Fast`* — deliberately open, but **filed no successor**. ADR-0096 §D4 requires a written justification for every surviving rung-3 entry, §D4/E2 already records this one's rationale as **expired**, and §D4's own rule is "an entry whose stated rationale has expired is re-decided, not re-cited". This is that re-decision, with the measurement it needs.

Measured on `main` @ `a17009b5` against upstream **`JSON::Fast:ver<0.20.1>:auth<zef:timo>`** (1117 lines, Artistic-2.0, `depends: []`).

## The headline: the recorded blocker no longer exists

`docs/batteries/json-tiny.md`, `BATTERIES.md:288`, `src/runtime/json.rs:1-7` and `news/2026-09/json-use-time-interception-retired.md` all still say:

> the real distribution depends on ~50 `nqp::` ops mutsu does not implement

Run it and that is false. Upstream `JSON::Fast` **parses completely** and reaches *runtime*:

```console
$ mutsu -I <JSON-Fast-0.20.1>/lib -e 'use JSON::Fast; say to-json({a=>1, b=>[1,2,"x"]});'
Unsupported nqp:: op: nqp::bindpos
  in block <unit> at .../lib/JSON/Fast.pm6 line 489
```

Probing every one of the 51 `nqp::` ops the file uses, **9 are missing**:

| op | uses | lines in `Fast.pm6` |
|---|---|---|
| `nqp::bindpos` | 22 | 489-510 (one contiguous table init) |
| `nqp::shift_i` | 6 | 591, 604, 631, 635, 644, 645 |
| `nqp::pop_s` | 4 | 256, 278, 292, 310 |
| `nqp::push` | 4 | 846, 853, 989, 996 |
| `nqp::chr` | 3 | 615, 667, 674 |
| `nqp::p6scalarwithvalue` | 3 | 797, 846, 853 |
| `nqp::p6bindattrinvres` | 2 | 904, 910 |
| `nqp::hash` | 1 | 765 |
| `nqp::ifnull` | 1 | 763 |

[...]

Issue #8183, opened 2026-09-12
Retire the JSON `use`-time interception: a performance gap does not justify a substitution


tokuhirom:
`use JSON::Fast` / `use JSON::Tiny` answer `to-json` / `from-json` from a native
Rust implementation (`src/runtime/json.rs`, 759 lines, plus `src/vm/vm_native_json.rs`)
that is selected by matching the module *name*. This issue argues the mechanism
should be retired, independently of whether mutsu keeps a fast JSON path at all.

User decision (2026-09-12): **being slow is not a reason to introduce a mechanism
like this.** A measured gap justifies optimizing; it does not justify substituting a
semantically divergent implementation under the real module's name.

## What the mechanism actually does

It is not a provider filling a gap where nothing resolves. The vendored module is on
the path and its routines *do* resolve — the native path then deliberately returns
before `call_routine_def` to beat them:

- `src/runtime/calls.rs:189` — statement position: if the resolved def's package is
  `JSON::Fast`/`JSON::Tiny`, `try_native_json_function` wins over the resolved def.
- the expression path does the same in `vm/vm_call_func_ops.rs`.
- `src/parser/stmt/simple/module_exports.rs:201` — the parser carries the two module
  names in its export list.
- `src/runtime/runtime_module.rs:20` — `json_module_loaded()` gates the whole thing on
  `use`.

Three specific problems, in increasing order of seriousness:

1. **Module names are hardcoded into dispatch at three layers.** Any dispatch change
   has to know these strings exist in order not to break them silently.

2. **The exception type is chosen by the set of loaded module names.**
   `json_tiny_exception_style()` (`src/runtime/runtime_module.rs:31`) is
   `JSON::Tiny loaded && !JSON::Fast loaded`, self-documented as a "best-effort
   guess". A program that loads both gets `JSON::Fast`-shaped errors from its
   `JSON::Tiny` calls. This is observable semantics decided by an unrelated `use`
   elsewhere in the program.

3. **It jumps the module-resolution ladder.** BATTERIES.md §6 says an explicit `-I`,
   `MUTSULIB` or installed module outranks the bundled floor; the interception
   bypasses that for these two names. `docs/batteries/json-tiny.md` already flags this
   under "What is NOT fixed by this record", including the case that matters: an
   upstream security fix cannot reach a user who cannot rebuild mutsu.

## Why the recorded rationale no longer reads the way it did

`docs/batteries/json-tiny.md` records the split as permanent policy on two grounds.

[...]

PR #8288, opened 2026-09-13
feat(batteries): vendor JSON::Fast and delete the native provider

tokuhirom:
`use JSON::Fast` now loads `modules/JSON-Fast/lib/JSON/Fast.pm6` — upstream `timo/json_fast` tag `0.20.1`, vendored byte-identical — through the ordinary module-resolution ladder, like any other battery. **All 14 of its own upstream test files, 931 assertions, pass against the bundled copy.**

This closes the campaign #8226 opened, and with it **ADR-0096 §D4/E2**. That ledger is now one entry: `NativeCall`.

## What is deleted

`use JSON::Fast` used to be answered by a native Rust implementation selected by the module's *name*. #8183 demoted it from an interception to a last-resort fallback and took `JSON::Tiny` off it entirely; this removes what was left:

- the `json_native_provider` flag and `json_native_provider_active()`;
- `try_native_json_function` and its four dispatch sites (`runtime/calls.rs`, `vm/vm_call_func_ops.rs`, `runtime/builtins_operators_fallback.rs`, `runtime/accessors_resolve.rs`);
- `JsonImportDefaults`, which existed only to model the `use JSON::Fast <immutable !pretty>` import list.

The proof that nothing name-keyed survives is not a code reading — move `modules/JSON-Fast` aside and `use JSON::Fast` answers *"Could not find JSON::Fast in: (module repositories)"*, like any other missing module.

## A correction to #8226's step 4

The issue said to delete `src/runtime/json.rs` and `src/vm/vm_native_json.rs` outright. **Doing that literally would have broken zef.**

Both files survive, narrowed to one caller: `Rakudo::Internals::JSON.to-json` / `.from-json`. That is **core Rakudo surface, not a module** — `raku -e 'say Rakudo::Internals::JSON.to-json({a => 1})'` resolves with no `use` — so it is a builtin like any other and D4's rung-3 ledger does not reach it. zef reads every `META6.json` through it (`vendor/zef/lib/Zef.rakumod`), as do OpenSSL's `%?RESOURCES` loading and JSON::JWT. Verified: `Zef::from-json` still answers, and `JSON::JWT` / `Log::Timeline::Output::JSONLines` / `Cro::HTTP::BodySerializers` all still load.

## One real bug the provider was masking

`t/collections/element-store-itemization.t` went red on the vendored module — and it was **right**, because rakudo running that same module agrees with it:

```console
$ raku  -I modules/JSON-Fast/lib -e 'use JSON::Fast; say from-json(Q<{"a":[1,2]}>)<a>.raku'
$[1, 2]
$ mutsu -e 'use JSON::Fast; say from-json(Q<{"a":[1,2]}>)<a>.raku'   # before
[1, 2]
```

`nqp::p6scalarwithvalue` returned its value bare. Rakudo wraps it in a fresh `Scalar`, and the observable half of that is **itemization**: a value bound in through raw `nqp::bindkey` never reaches the store-side hook that itemizes an ordinary `%h<k> = [1,2]`, and `JSON::Fast`'s `parse-obj` builds every decoded object exactly that way. The array path was already right (`nqp::push` does itemize), which is why only object *values* were wrong.

Which kinds itemize is **measured against rakudo, not assumed**:

| value | rakudo | now |
|---|---|---|
| `[1, 2]` | `$[1, 2]` | ✅ |
| `{:c(3)}` | `${:c(3)}` | ✅ |
| `(1, 2)` | `$(1, 2)` | ✅ |
| `(1, 2).Seq` | `$((1, 2).Seq)` | ✅ |
| `1..3` | `1..3` | ✅ not itemized |
| `a => 1` | `:a(1)` | ✅ not itemized |
| `True` / `42` / `"s"` / `Any` | plain | ✅ not itemized |

That is narrower than `Value::itemize_for_element_store`, which also itemizes every `Range` variant — right for its own call sites, wrong here — so the op spells its own set out rather than reusing the helper.

## Why this module

Of the 1,625 distributions in the `ecosystem/` ledger, **541** carry `JSON::Fast` in their resolved dependency closure, against `JSON::Tiny`'s **32**. Inside the tree the split is starker: five `use JSON::Fast;` statements across three bundled batteries (`Cro::HTTP`, `JSON::JWT`, `Log::Timeline`) and six `META6.json` declarations — while **nothing bundled `use`s `JSON::Tiny`**.

`JSON::Tiny` stays bundled. Whether it should is a BATTERIES.md §2 selection question reserved for the maintainer, and deliberately not folded in here; it is cheap to keep, and it is the reference implementation that exercises the grammar engine as a real vendored module.

## What it took, and what the record said

Six places in the tree said the real distribution "depends on ~50 `nqp::` ops mutsu does not implement". Nobody had run it: **42 of its 51 ops already worked**. The nine that did not, and the four things they were hiding, landed in #8239; three more general gaps in #8282; the two non-JSON blockers on `t/01-parse.t` as #8232 (ADR-0100) and #8233. Every one of those is a fix for mutsu, not for JSON — the shape ADR-0096 §D3 asks for.

## Tests and gate wiring

- `batteries.lock` — new row pinning `72af212130f190443028e5a6b23827445c95559f` (verified: the vendored `lib/` and `META6.json` are byte-identical to that tag).
- `batteries-whitelist.txt` — all 14 files, kept sorted.
- `docs/batteries/json-fast.md` — the battery record with the §3 re-vendoring recipe.
- `t/vm/nqp-list-hash-ops.t` — 8 new assertions on the op's itemization set. **43/43 under mutsu and under `prove -e raku`**; the end-to-end `from-json` shape deliberately stays out of this file so it remains rakudo-runnable (rakudo has no JSON::Fast installed).
- `t/modules/batteries/json-module-ladder.t` — now pins both JSON modules resolving through the ladder, and that `Rakudo::Internals::JSON` survived the removal (8 assertions).
- Prose corrected in `BATTERIES.md` (the one JSON row becomes two), `ANALYSIS.md`, `docs/batteries/json-tiny.md`, `docs/batteries/python-stdlib-comparison.md`, `src/runtime/json.rs` + `src/vm/vm_native_json.rs` headers, and a superseded note on `news/2026-09/json-use-time-interception-retired.md`.

## Verification

- `cargo fmt --all`, `make lint` (all four configurations) — clean.
- `make test` — **PASS**, 4207 files / 44825 tests.
- `make roast` — only the three documented container-only failures, at their documented shapes (`docs/agent-environments.md`): `6.c/S32-io/file-tests.t` `Failed: 4`, `S16-filehandles/filetest.t` `Failed: 25` (both `uid 0`), `S32-io/IO-Socket-Async.t` exit 124 "planned 40 ran 17".
- Upstream suite against the **bundled** copy (not `-I <dist>/lib`): `Files=14, Tests=931, Result: PASS`.
- The full `scripts/battery-testsuite.sh` gate was not run locally — it fetches every battery's suite and wants a release build. This battery's half of it is the 14/14 above.

Refs #8226

https://claude.ai/code/session_01AMQwQoZeiJDBM5iK3hvPsK

---
_Generated by [Claude Code](https://claude.ai/code/session_01AMQwQoZeiJDBM5iK3hvPsK)_


tokuhirom:
## Performance is measured now, and it is the one thing worth a second look before merging

#8226's risk list said "Performance is unmeasured". I had written that section as *deliberately* unmeasured. That was the wrong call — it is cheap to measure, and the number is large enough that shipping it unstated would have been worse. Pushed in `7b2e4cea`.

200 encodes + 200 decodes of a 2,380-byte META6-shaped document, **release** build:

| | encode | decode |
|---|---|---|
| rakudo, running **this same vendored module** | 0.067s | 0.085s |
| mutsu, this module | **4.243s** | **10.709s** |
| mutsu, the native Rust codec (`Rakudo::Internals::JSON`, **same binary**) | 0.008s | 0.011s |

- **~63x slower than rakudo** on encode, **~126x** on decode — same module, same input, two interpreters.
- **~530x / ~975x** slower than what `use JSON::Fast` reached before this PR. **That is a real user-visible slowdown this PR ships.**

The A/B needs no rebuild because `Rakudo::Internals::JSON` still uses that codec.

### Why it ships anyway

ADR-0096 §D3, which the maintainer set: *a measured gap justifies a transparent, semantics-preserving optimization, never a substitution under the real module's name.* The substitution also had its own cost — it emitted **unescaped, invalid JSON** for any string containing a quote or a control character, and nobody noticed for months precisely because it was not the real module. Correctness went up; wall clock went down a lot.

Filed as [#8289](https://github.com/tokuhirom/mutsu/issues/8289) (`todo:perf`) with the repro and one early signal: `nqp::add_i` costs **0.5x** a plain `+` under rakudo (the op bypasses dispatch) and **1.5x** under mutsu (it does not). That 3x is real and worth fixing on its own — an `nqp::` op is meant to be the cheap primitive — but it does not explain 500x, so the bulk is elsewhere. Most likely the per-character ops the scanner runs once per input byte (`nqp::ordat`, `nqp::substr`, `nqp::strtocodes`, `nqp::splice`), and/or the `Uni` codepoint store added for `str-escape`: if either rebuilds a Vec per call, encode is quadratic in string length. Decode being 2.5x worse than encode points the same way.

If you would rather not ship a ~500x JSON slowdown until #8289 is worked, say so and I will hold this PR — the alternative is reinstating a provider that ADR-0096 §D3 rules out, so holding is the only other option on the table.

---

tokuhirom:
## `test-suites` red: not this PR's — a sourcehut fetch failure in the batteries gate

This PR touches the batteries, so `ci.yml` ran the full `scripts/battery-testsuite.sh` gate. **JSON::Fast passed it outright:**

```
=== battery: JSON::Fast (commit 72af212130f1) ===
  01-parse.t                     PASS(2 todo)
  02-structure.t … 14-comments.t PASS
```

14/14 in the gate itself, against the bundled `lib/` — which is the authoritative version of the 14/14 I ran locally.

What failed was a *different* battery, and not on a test:

```
…
```

The five `REGRESSION: whitelisted 'Terminal::ANSI …'` lines are the consequence of that fetch, not five separate failures — the suite was never set up, so none of its files could pass.

**The pin is not stale.** That exact commit is HEAD of `master` on sourcehut and fetches fine from here:

```console
$ git ls-remote https://git.sr.ht/~bduggan/raku-terminal-ansi | head -2
691d0959ff8ea0a7c491167b366746cdfc0fb0be	HEAD
691d0959ff8ea0a7c491167b366746cdfc0fb0be	refs/heads/master
```

So this is a transient failure against a third-party host my diff does not touch — and it would hit any batteries-touching PR opened right now, not just this one. `git.sr.ht` is the only non-GitHub `test_url` in `batteries.lock`.

Re-run requested once (run `34750104971`, failed jobs only). A second failure is real and I will treat it as such.

### The earlier red on `addb6840` was mine, and was noise

The three aggregator failures before this (`test`, `gc-stress`, `jit-stress`) reported their halves as **`cancelled`**, not `failure`, with `build: success` — my push of `7b2e4cea` superseded that run. Nothing to act on there.

---

Issue #7539, opened 2026-09-08
`Config::TOML` battery is blocked on core interpreter campaigns


tokuhirom:
## Current measurement (2026-09-07)

Both upstream suites re-fetched from GitHub (`raku-community-modules/Crane`,
`raku-community-modules/Config-TOML`) and run from their own directories
against a **debug** build; `Config::TOML` with `Crane/lib` on `MUTSULIB`.

| Suite | raku | mutsu (2026-08-31) | mutsu (2026-09-06) | mutsu (2026-09-07, before) | mutsu (2026-09-07, after) |
| --- | --- | --- | --- | --- | --- |
| `Config::TOML` v0.1.3 | 19/19 files | 0/19 | 10/19 | 10/19 | **11/19** |
| `Crane` v0.1.2 | 15/15 files | 3/15 | 3/15 | 3/15 | **4/15** |

The 2026-09-06 numbers reproduced exactly, so that pass's re-measurement was
sound. This pass fixed four general interpreter bugs (below) and moved both
counts: `Config::TOML`'s `exceptions/01-parser.rakutest` went from 12 failing
assertions to a clean 17/17, and `Crane`'s `get.rakutest` now passes.

`Config::TOML` passing: `api/01`, `exceptions/01`, `grammar/01-04`,
`grammar-actions/03`, `special-cases/01`, `04`, `05`, `06`.
`Config::TOML` failing: `dumper/01`, `exceptions/02`, `grammar-actions/01`,
`02`, `04`, `special-cases/02`, `03`, `07`.
`Crane` passing: `at`, `exists`, `get`, `test`.

## Fixed 2026-09-07 (general interpreter bugs, all pinned)

1. **A container element's `++`/`--` seeded from `Int` 0 regardless of the
   declared element type.** `my Bool:D %h; %h<a>++` died with
   `Type check failed for an element of %h; expected Bool:D but got Int (1)`;
   raku answers `True` (`postfix:<++>` is `.=succ`, and an uninitialized
   `Bool` slot succs to `True`). The scalar path already seeded from the
   declared type (`normalize_incdec_source_with_type`); the element path did
   not. Now both share `incdec_seed_for_constraint`, which also strips a
   `:D`/`:U`/`:_` smiley before the lookup.
2. **An object hash's element `++`/`--` keyed by the display string.**
   `my %h{Int}; %h{5}++; say %h{5}` read back the type object, because the
[...]

PR #8303, opened 2026-09-13
fix(vm): an associative subscript on a Pair stores into a container value

tokuhirom:
`Pair` DOES `Associative` and its `AT-KEY` hands back `.value` **itself** — a
Pair gives its value no `Scalar` of its own — so `$p<key> = ...` is not an
element store at all. It is a store to the value the subscript reaches, and that
value decides: an `Array`/`Hash` IS a container and takes it the way
`@a = LIST` does, everything else is `X::Assignment::RO`.

```raku
my $p = (c => [1, 2]);  $p<c> = [3, 4];   # :c([3, 4])   Array.STORE
my $p = (c => [1, 2]);  $p<c> = 5;        # :c([5])      ditto, one element
my $p = (c => {:a(1)}); $p<c> = {:b(2)};  # :c({:b(2)})  Hash.STORE
my $p = (c => 1);       $p<c> = 9;        # Cannot modify an immutable Int (1)
```

[#8275](https://github.com/tokuhirom/mutsu/pull/8275) taught mutsu the refusing
half of that rule and applied it to every value alike, so all three storing rows
above died. The three *descending* spellings were worse: `$p<c>[0] = 9`,
`$p<c><a> = 7` and `%h<x><y>[0] = 9` reached no arm that recognizes a `Pair` as
a container, wrote nowhere, and reported success; a 3+-level chain rebuilt the
Pair chain as nested `Hash`es instead (`%h<x><y><z> = [3,4]` answered
`{:x(${:y({:z($[3, 4])})})}` where rakudo keeps `{:x(${:y(:z([3, 4]))})}`).

## What changed

The rule is now one predicate, `pair_subscript_aggregate`, whose `Some` is
exactly `pair_subscript_store_refusal`'s `None`. Each store site then STOREs
into that container (`store_into_pair_aggregate`, mutating the shared node in
place so the `Pair` — which cannot be replaced — and every other alias of the
container see it) or descends into it:

- the single-subscript named store (above ADR-0040's itemization hook);
- both two-level arms (Hash-outer and Array-outer);
- the 3+-level walk's intermediate and final levels;
- the computed-target store (`OpCode::IndexAssignGeneric`);
- the path-accessor store (`Crane::At.at($root, @path){$step} = $value`).

Two supporting bugs came out with it: `subscript_descent_refusal_at` answered
`Type Pair does not support associative indexing` for a `Pair` that IS
descendable, and the deep walk passed the *next* level's positional flag where
it needed the flag of the subscript actually being applied to the slot.

The whole-container STORE takes the rvalue **un-itemized**: ADR-0040's
element-store itemization hook fires above every one of these ops, and an
itemized `[3, 4]` would have STOREd as `[[3, 4],]`. A genuinely itemized rvalue
(`my $x = [3, 4]; $p<c> = $x`) still nests, exactly as `@a = $x` does.

## Measurement

Found re-measuring [#7539](https://github.com/tokuhirom/mutsu/issues/7539)'s
`Config::TOML` + `Crane` battery pair, where this was the last failure in
`Crane`'s `add.rakutest`: a nested in-place add
(`Crane.add(%i, :path<a b c>, :value(...), :in-place)` over the colonpair
fixture `:a({:b(:c([...]))})`) reaches
`Crane::At.at($root, @path){$step} = $value`, whose `$step` lands on the `Pair`
`c => [...]`.

| Suite | raku | before | after |
| --- | --- | --- | --- |
| `Crane` v0.1.2 | 15/15 files | 9/15 | **10/15** |
| `Config::TOML` v0.1.3 | 19/19 files | 14/19 | 14/19 |

Release build, 240s per-file budget, prove's exit status as the verdict. No
regression in either suite.

## Validation

`cargo fmt --all`, `make lint`, `make test` (4,219 files / 44,967 assertions)
and `make roast` (1,437 files / 219,635 assertions) all green, bar the three
failures [`docs/agent-environments.md`](https://github.com/tokuhirom/mutsu/blob/main/docs/agent-environments.md)
documents for this container (`uid 0` file-permission tests ×2, sandboxed
network ×1), matched by name and by subtest range.

Pin: `t/collections/range-pair/pair-subscript-stores-into-a-container-value.t`
(23 assertions), verified to pass under rakudo v2026.07 as well as under mutsu.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01EBmiJH8aYB8BQhXFJZaTQ7)_


tokuhirom:
## `gc-stress-tap` red on `cba1fb1`: `t/concurrency/promise/gc-promise-timer-stress.t` timed out with 0 of 3 tests run

```
t/concurrency/promise/gc-promise-timer-stress.t  (Wstat: 31744 (exited 124) Tests: 0 Failed: 0)
  Non-zero exit status: 124
  Parse errors: Bad plan.  You planned 3 tests but ran 0.
```

It is the only failure in the job; no `MUTSU_GC_VERIFY` FAIL, no crash report,
no quarantine retry. Every other check on this head is green or still running.

**Why this is not this PR's failure.** The diff is confined to the
`Pair`-subscript store paths (`pair_subscript_aggregate` /
`store_into_pair_aggregate` and the store sites that call them). This test does
no `Pair` subscript store at all — its only subscript write is
`%h<self> := %h`, a `:=` bind into a real `Hash`, which the new arm in
`exec_index_assign_expr_named_op_seeded_inner` excludes twice over: the target
is a `Hash` (so `pair_subscript_aggregate` answers `None`) and the rvalue
carries the `__mutsu_bind_index_value` marker the arm guards on. Nothing in the
diff reaches promise, timer, thread or GC code.

**Reproduction attempts, all under the job's exact environment**
(`MUTSU_GC=on MUTSU_GC_EVERY_CANDIDATE=1024 MUTSU_GC_VERIFY=1`, debug build,
this PR's head):

| shape | runs | failures |
| --- | --- | --- |
| serial | 30 | 0 |
| 8-way parallel × 5 rounds | 40 | 0 |

70/70 clean, and the file completes in **0.44s** — three orders of magnitude
under the per-file budget it hit. `make test` and `make roast` were both green
locally before publishing, and the local `t/` run includes this file.

The failure shape is the one the triage protocol calls out as plausibly
load-borne: exit 124 with `Failed: 0` and **no test body completed**. There is
no fix to port — nothing is failing, the process never got going.

Re-running the failed job once, as the one re-run that case allows. If it comes
back red, that is a real hang and I will root-cause it rather than re-run again.

---

Issue #8294, opened 2026-09-13
A bare `::` is the null routine name, not an indirect-name marker


tokuhirom:
Found while working the parse-failure index (#7954). `UI::HTMLWindow`'s `lib/UI/HTMLWindow.rakumod:58` is `&routine.wrap( anon method :: (Window: *@_, *%_) { ... } )`, which does not parse.

## Repro

A bare `::` where a routine's name would go means "this routine has no name":

```
$ raku  -e 'class W {}; my $m = anon method :: (W: *@a) { 42 }; say $m.^name'
Method
$ mutsu -e 'class W {}; my $m = anon method :: (W: *@a) { 42 }; say $m.^name'
===SORRY!=== Error while compiling -e
```

The `sub` spelling parses in mutsu but produces the wrong thing rather than an error:

```
$ raku  -e 'my $m = anon sub :: ($x) { 42 }; say $m(1)'
42
$ mutsu -e 'my $m = anon sub :: ($x) { 42 }; say $m(1)'
No such method 'CALL-ME' for invocant of type ...
```

`--dump-ast` shows why: mutsu reads `sub ::(...)` as an **indirect declaration
name** (`SubDecl { name: __INDIRECT_DECL_NAME__, name_expr: Some(Var("x")) }`),
so the signature is consumed as the name expression and the routine ends up
unnamed *and* unsigned.

That indirect reading is not rakudo's. rakudo rejects the spelling outright:

```
$ raku -e 'my $n = "foo"; sub ::($n) { say "called" }; foo()'
===SORRY!===
$ mutsu -e 'my $n = "foo"; sub ::($n) { say "called" }; foo()'
called
```

So there are two things here, and the second is why this is a ticket rather
than a one-line addition:

1. `parse_sub_name_inner` (`src/parser/stmt/sub/sub_name.rs`) requires an
   identifier first, so a leading `::` never reaches the no-name branch — which
   in turn wants a `(` or `{` next and sees `::`.
2. mutsu's existing `sub ::($expr)` indirect-name path is a mutsu invention on
   this spelling. Deciding what happens to it (delete, or keep and disambiguate
   by gluedness, since rakudo's `:: (...)` with a space is a signature) is the
   design call the fix has to make, and there may be existing tests pinning the
   current reading.

The same null name is legal on `sub`, `method`, `submethod` and `multi`, so
whatever the answer is, it belongs at the shared declarator-name site.

Refs #7954

tokuhirom:
[...]

PR #8343, opened 2026-09-13
test(oo): pin the half of #8311 the method-literal fix left untested

### LINKED ISSUE #8311: `anon method (Invocant: *@a) {...}` compiles as a plain Sub — wrong `.^name`, and the invocant is not excluded from the slurpy
Found while working #8294 (the `sub :: (...)` / `anon method :: (...)` null-declarator-name
parse fix).

## Repro

```
$ raku  -e 'my $m = anon method (*@a) { 42 }; say $m.^name'
Method
$ mutsu -e 'my $m = anon method (*@a) { 42 }; say $m.^name'
Sub
```

```
$ raku  -e 'my $m = method (*@a) { 42 }; say $m.^name'      # no "anon" either
Method
$ mutsu -e 'my $m = method (*@a) { 42 }; say $m.^name'
Block
```

The invocant is also not excluded from the slurpy — a consequence of the routine
compiling as a `Sub`/`Block` rather than a `Method` (a `Sub` has no invocant concept
at all):

```
$ raku  -e 'class W {}; sub call_it(&r) { r(W, 1, 2) }; say call_it(anon method (W: *@a) { @a.join(",") })'
1,2
$ mutsu -e 'class W {}; sub call_it(&r) { r(W, 1, 2) }; say call_it(anon method (W: *@a) { @a.join(",") })'
,1,2
```

mutsu's `@a` slurps `(W, 1, 2)` (the invocant included, and `W.join` stringifies to
empty), where rakudo's `@a` is only `(1, 2)` — the invocant `W` was consumed by the
signature's `W:` marker and never enters the slurpy at all.

## Where

`parse_anon_method_with_params` (used for the bare `method (...) {...}` expression
form, `src/parser/primary/ident/identifier_call.rs`'s `"method"` arm) produces a
`Block`, and `parse_anon_sub_with_params` (used for `anon method (...) {...}`, the
same file's `"anon"` arm) produces a `Sub` — neither builds the `Method`-shaped value
`anon method NAME (...) {...}` (a NAMED anon method, which does work correctly and
reports `Method`) does. Confirmed both anon-method entry points share this gap, so it
is a real Method-vs-Sub/Block classification bug in the *nameless* anon-method path,
not tied to `::`/#8294's fix (verified with and without `::`, with and without `anon`
— all four combinations of {nameless, anon-prefixed} × {`::`, no `::`} share this
exact same divergence, confirming it predates and is orthogonal to #8294).

## Why this matters

The real-world case from #7954 (`UI::HTMLWindow`'s `&routine.wrap(anon method :: (Window:
*@_, *%_) { ... })`) now *parses* after #8294's fix, but the wrapped callable still
does not behave like a real Method — its invocant handling is wrong. Whether that
matters for `UI::HTMLWindow` specifically needs checking once this is fixed.

tokuhirom:
#8311 reported **two** divergences for a nameless method literal: the type it reports (`Sub` / `Block` instead of `Method`), and the invocant not being excluded from the signature's positionals.

Both were fixed by `8063666da` ("a method literal is a Routine, not a Block"), which was written against the MetamodelX::Dataclass parity cluster rather than against this ticket. Its pin, `t/oo/method/method-literal-is-a-routine.t`, covers the first divergence thoroughly — every declarator spelling including `submethod` and `my method`, `return` acting as a routine boundary, and the types that must not have moved.

**It does not cover the second.** Its invocant tests are a user-named invocant (`anon method ($obj: $x)`) and the Dataclass site's `Mu $obj: *%args`. Neither is the shape the ticket documents, which is a type-object invocant marker in front of a **positional** slurpy:

```
$ raku -e 'class W {}; sub c(&r) { r(W, 1, 2) }; say c(anon method (W: *@a) { @a.join(",") })'
1,2
```

The slurpy is what makes this bug observable at all. A fixed-arity signature fails loudly on the arity when it is handed one argument too many; a slurpy just quietly binds the extra one, so the old behaviour showed up only as a stray leading comma (`W.join` stringifies to empty). The `::` null declarator name from #8294 — the spelling `UI::HTMLWindow` actually uses, `&routine.wrap(anon method :: (Window: *@_, *%_) { ... })` — was likewise unpinned.

## What this adds

`t/oo/method/method-literal-invocant-excluded-from-positionals.t`, 10 tests, deliberately **non-overlapping** with the existing file:

- the `W:` marker in front of `*@a`, across the `method`, `submethod`, `anon method` and `anon method ::` spellings;
- a named slurpy beside a positional one (`(W: *@a, *%h)`);
- ordinary positionals binding from the *second* argument on, and keeping their order;
- `self` being the receiver rather than the first positional;
- an undeclared invocant still being consumed by the implicit one;
- the neighbouring shape that must **keep** the old behaviour — a `sub` literal, which has no invocant concept, still slurping all three arguments.

## Note on how this came about

I claimed #8311 and implemented the fix independently before discovering `8063666da` had landed on main in the meantime; a cherry-pick onto current `main` conflicted across 13 files. That implementation is discarded — theirs is strictly broader (it also handles `submethod` / `anon submethod` / `my method`, and replaces `is_sub: bool` with a `RoutineDeclarator` enum, which is the better modelling). I did use my own 18-test pin as an independent cross-check: it passes unmodified against `main`, which is what established the fix is complete. Only the coverage gap above survives into this PR.

## Verification

**Correction to the original text of this section.** It claimed `make test` and `make roast` had been run and were green apart from the three documented container-only failures. That was wrong, and I am replacing it rather than quietly editing it.

The first push failed `test-check` on a real defect of mine: `check-t-layout` routes a basename containing `slurp` to `t/io/`, because `slurp` is an IO routine, and the file was named `...-invocant-is-not-slurped.t`. It is about signature binding, not IO, so `a38d87f1a` renames it (and its news entry) to `...-invocant-excluded-from-positionals` rather than adding a RULE/OVERRIDES entry.

Why I did not catch it locally: I had chained `cargo fmt --all && make check-t-layout | tail -2 && (make test && make roast)`. `check-t-layout` failed for exactly this reason and short-circuited the chain, so **neither suite ran on this change at all**; `tail -2` hid the failing line, and the `tmp/make-*.log` files I then read were left over from the previous ticket's gate, with a matching exit status. I reported those stale logs as this change's result.

Current state, run separately so one cannot mask the other:

- `make check-t-layout` — passes (4243 files).
- the new test file — 10/10 under mutsu and under real Rakudo.
- `make test` / `make roast` — running now on the renamed tree; I will report both exit statuses here, and fix forward if either is red.

Closes #8311


tokuhirom:
Both suites have now finished on the renamed tree, run as separate commands so neither could mask the other:

| | result |
| --- | --- |
| `make check-t-layout` | pass (4243 files) |
| `make test` | **exit 0** |
| `make roast` | exit 2 — `roast/6.c/S32-io/file-tests.t` `Failed: 4`, `roast/S16-filehandles/filetest.t` `Failed: 25`, `roast/S32-io/IO-Socket-Async.t` exit 124 / `Failed: 0` |

Those three are an exact match by name and by failure shape to the container-only list in `docs/agent-environments.md` (`uid 0` for the two `chmod` files, the network sandbox for the socket one), which is the only licence to ship with a red `make roast`.

I checked the log timestamps against the wall clock this time rather than trusting the exit status alone — that is precisely what I failed to do on the first push, where the logs I read were a previous ticket's and happened to carry the same exit status.

`test-check`, the check that caught the naming defect, is green on `a38d87f1a`.

---

Issue #7556, opened 2026-09-08
Immutable lvalues mutsu still lets you assign to (survey)


tokuhirom:
Found by the exception-taxonomy survey in
`news/2026-08/readonly-assign-exception-taxonomy.md`. That work fixed *which*
exception a rejected assignment throws; this ticket collects the cases where
mutsu does not reject the assignment at all, which the same survey surfaced.
Every surviving row below was re-measured against `raku` on **2026-09-07** on a
fresh build of `main` at `2c7fc1dde`; every one of them still reproduced, and
the ones this file no longer lists were closed on that same day. C2, D, E and F
were re-verified unchanged then (D's shape had drifted — see its row).

## Status (last updated 2026-09-07)

Closed since the survey opened:

- **section C in full (2026-09-06)** — a `$` bound to a non-Scalar container now
  refuses whole-value assignment:
  `news/2026-09/scalar-bind-to-a-non-scalar-container-refuses-assignment.md`;
- the element-store and `:=`-bind halves —
  `news/2026-09/immutable-element-store-and-bind.md`;
- the closure/map/grep topic family (`(1,2).map({$_=5})`,
  `(1,2).grep({$_=5})`, `{ $_ = 5 }(7)`, plus `(1..3).map`, `%h.keys.map`, the
  listop `map { $_ = 5 }, 1, 2` and its `grep` twin, `{ $_ = 5 }(3+4)` and
  `my &s = { $_ = 5 }; s(7)`) —
  `news/2026-09/closure-and-map-grep-topic-readonly.md`, pinned by
  `t/closure-topic-readonly.t`.

Tracked separately, same family, different producer: a **raw-invocant method**
called on an immutable `List` element (`my $l = (1,2); $l[0].mut`) succeeds
silently where raku dies —
`todo/tickets/immutable-list-element-write-is-silently-dropped.md` (2026-09-06).
Every *ordinary* store to such an element is already refused correctly, so that
one is a method-call-path gap rather than a store-path gap, and it belongs to
ADR-0067's L4/L5/M1/M2 readonly-enforcement rows.

Section B was re-measured on **2026-09-07** and three of its four producers
were closed the same day
(`news/2026-09/slice-first-and-block-topic-element-containers.md`); only
producer 1 (`.list` feeding `map`) survives. Section C was re-measured on
2026-09-06 (all seven rows diverged) and then closed.

**Section A was re-measured and mostly closed on 2026-09-07**
[...]

PR #8378, opened 2026-09-14
fix: chained store into a bare-valued List element refuses, doesn't clobber

tokuhirom:
## Summary

- `@a[0][0] = 9`, where `@a[0]` is a `List` of bare (non-container) elements (e.g. `my @a = (1,2,3), 4`), silently clobbered the List's own backing storage instead of refusing the store the way rakudo does ("Cannot modify an immutable List ((1 2 3))").
- A `List` is not a container of its own; writing into one of its elements is only legitimate once that specific element has been promoted to a shared cell (the shape `take-rw` builds — `t/collections/lazy-seq/take-rw-shared-cell.t`). A prior attempt at this row (recorded in #7556) refused every store into a `List`/`ItemList` slot by its `ArrayKind` at the point an outer subscript *descends* into it, which regressed that exact test — a List whose specific element IS a live cell must still write through.
- Fixed by deciding the refusal at the innermost element write instead, once the specific reached element's own shape is known — not at the descent-time check (`subscript_descent_refusal_at`, which deliberately still treats every `ArrayKind` as descendable).
- Both chained-store engines in `src/vm/vm_var_assign_index_named.rs` that reach a List element this way needed it: the 2-level named chain (`exec_index_assign_expr_nested_op_body`) and the 3+ level deep-nested walk (`exec_index_assign_deep_nested_op_body`). Each `arr[i].with_array_mut(|inner_arr, _kind| ...)` call was discarding the array's kind; both now check whether it is `List`/`ItemList` and the specific element is not already a `ContainerRef` cell before writing through. A `:=`-bind and a real (mutable) `Array`/`Shaped` element store are both unaffected.

## Test plan

- [x] Added 4 assertions to `t/collections/subscript/chained-subscript-store-refuses-defined-value.t`: the 2-level case, an out-of-range inner index, the 3+-level (hash-rooted) twin, and a real-Array control case confirming ordinary element stores still write through.
- [x] Verified against `raku` directly — all cases match exactly, including the message text (which names the whole List's gist, unaffected by which index or whether it's in range).
- [x] Re-ran `t/collections/lazy-seq/take-rw-shared-cell.t` explicitly — still green (the exact regression a prior attempt at this row hit).
- [x] Ran the full `t/collections/subscript/` and `t/collections/lazy-seq/` suites — no regressions.
- [x] `cargo fmt --all`
- [x] `make lint` (all 4 configs clean)
- [x] `make test` (green)
- [x] `make roast` — only the 3 documented container-only failures (`roast/6.c/S32-io/file-tests.t`, `roast/S16-filehandles/filetest.t`, `roast/S32-io/IO-Socket-Async.t`), everything else green

Part of the survey in #7556 (not closing it — other rows remain open).

https://claude.ai/code/session_01RZKspCKXhDnjrza1p1Ywxr

---
_Generated by [Claude Code](https://claude.ai/code/session_01RZKspCKXhDnjrza1p1Ywxr)_


tokuhirom:
`lint-configs` failed on the `--no-default-features --features native` clippy pass: `needless_borrow` on `&list_type_name` (already `&'static str`) at two sites, which the default clippy config doesn't flag by itself but reproduces once forced (confirmed locally afterward). Fixed by dropping the redundant `&` in `da6cd736`; verified locally against all 4 lint configs (default clippy, native-features clippy, wasm32 clippy, rustdoc) plus the affected test files.

---
_Generated by [Claude Code](https://claude.ai/code)

---

Issue #8450, opened 2026-09-15
A pattern with a dense first-character set still costs ~1 µs per rejected position (`\w+ 'QQQ'` is 1.8x rakudo at 640 KB)


tokuhirom:
Left over from #8248 after the first-character-set prefilter landed.

The prefilter answers "can a match start here?" for patterns whose first-set is
sparse. It cannot help when the set is *dense*: `\w` admits about 85% of an
ordinary text subject, so almost every position still enters the engine, and
what is measured there is the raw per-position reject cost — the thing ADR-0099
§2.4 put at ~983 instructions.

## Numbers

Release build, 640 KB subject of repeated 46-char units, one scan, vs rakudo
2026.07 on the same box, after #8248's prefilter work:

| shape | mutsu | rakudo | ratio |
|---|---:|---:|---:|
| failing `\w+ 'QQQ'` | 619 ms | 340 ms | 1.8x slower |
| failing `:r \w+ 'QQQ'` | 312 ms | 480 ms | 1.5x faster |
| failing four-literal alternation | 124 ms | 801 ms | 6.5x faster |

Every other row of #8248's table is now a mutsu win or close to parity; this one
is the residue. Note the ratchet row is already fine, so the cost is
specifically the greedy grow-and-give-back at each failing position, not the
`\w` run itself.

## Next step

Profile one failing `\w+ 'QQQ'` scan over a 640 KB subject under callgrind and
attribute the per-position cost: `RegexCaptures`/`CapStore` construction per
start position, the candidate generator layers, and the give-back walk. ADR-0099
§2.4's earlier profile named "the five candidate-generator layers, candidate-Vec
construction and drop, memcpy, malloc, and RegexCaptures construction/drop" but
was measured on a literal scan, where no quantifier is involved — re-measure on
this shape before assuming the same distribution.

Two cheap hypotheses worth ruling out first, in this order:

1. A greedy `+` over a character class grows to the end of the word run and then
   gives back one position at a time, re-entering the continuation each time.
   Rakudo's ratchet row being *slower* than its greedy row suggests rakudo is
   doing something structurally different here.
2. The per-start-position `CapStore::new` / `RegexCaptures::default` in
   `regex_walk_ends_in_pkg` is paid even when the pattern captures nothing.

## Related

- #8248 (the scan-constant issue this is the residue of)
- #8272 (ADR-0099 Stage 1)
- ADR-0099 §2.4, and §4 Stage 2, which is the decision this measurement feeds

[...]

Issue #8449, opened 2026-09-15
The literal-prefix scan prefilter declines on a quoted literal, because `'abc'` parses into a Group


tokuhirom:
Found while measuring #8248.

`required_literal_prefix` (`src/runtime/regex/regex_prefilter.rs`) walks the
pattern's top-level tokens and accepts `RegexAtom::Literal` /
`RegexAtom::LiteralGrapheme`, breaking on anything else. A *quoted* literal in a
regex — `/ '67-8' \s $ /` — does not parse into a run of top-level `Literal`
tokens; it parses into a single token whose atom is a `Group` wrapping them. So
the walk breaks on the very first token, the prefix comes out empty, and the
scan declines the prefilter entirely.

That is not a small family of patterns: `/ 'foo' /` with quotes is the ordinary
way to write a literal in Raku regex, and the ADR-0099 Stage 1 win was measured
on unquoted ones.

## Repro

```
$ MUTSU_VM_STATS=1 ./target/release/mutsu -e 'my $s = "ab" x 300000; say so $s ~~ / <[0..9]> /'
```

Concretely, on a 640 KB subject of repeated 46-char units (the #8248 harness),
`/ '67-8' \s $ /` on `main` before the first-character-set work:

```
[mutsu vm-stats] regex-prefilter: applied=0 declined=1 positions_offered=0 position_hits=0
```

180 ms, against rakudo's 6 ms. The first-character-set filter added in #8248
does look through a `Group`, so the same scan now reports `applied=1` and costs
28 ms — but it is filtering on one character where a substring search on the
whole four-character literal was available and would have been far better.

## Fix

Teach `required_literal_prefix` to look through a `Group` /
`CaptureIsolatedGroup` whose own pattern is itself a literal run (and which
carries no quantifier, capture alias or separator), concatenating the inner run
into the prefix and continuing past it when the group is fully consumed. Decline
on `CaptureGroup` (its captures are observable) and on any inner pattern with
its own `:i` / `:m` flags, per the existing constraints.

The differential corpus in `tests/regex_prefilter_differential.rs` is the gate:
add quoted-literal cases and check them byte-for-byte against
`MUTSU_REGEX_PREFILTER=off`.

## Why it is a ticket and not a deep issue

It is one analysis function, the soundness direction is the safe one (a longer
required prefix rejects strictly more positions and can only be wrong if the
group is not in fact a required literal run), and the differential test already
exists.

tokuhirom:
[...]

Issue #8272, opened 2026-09-13
Regex scans enter the full engine at every start position: add the memoized declarative prefilter (ADR-0099 Stage 1) — `~~ /literal/` is 212x mutsu's own `.index` on the same subject


tokuhirom:
ADR-0099 Stage 1. The decision is recorded in [`docs/adr/0099-regex-engine-performance-strategy.md`](https://github.com/tokuhirom/mutsu/blob/main/docs/adr/0099-regex-engine-performance-strategy.md) §4; this issue is the work.

Read the ADR first — in particular §4 Stage 1's three constraints, which are the whole difficulty. This is **not** "add a memchr fast path".

## The gap

mutsu already ships a fast substring search. The regex engine does not use it. Same 640 KB subject, same question, two paths through the same interpreter:

| | mutsu | rakudo |
|---|---:|---:|
| `$big.index('zzzq-not-here')` | **0.4 ms** | 4.4 ms |
| `$big ~~ / 'zzzq-not-here' /` | **84.7 ms** | 1.6 ms |
| `$big.lc.index('zzzq')` | 14.1 ms | 10.1 ms |
| `$big ~~ / :i 'ZZZQ' /` | **160.3 ms** | 6.6 ms |

**212x** between mutsu's two answers to the same question, and mutsu's `.index` is itself 10x faster than rakudo's. rakudo's `~~ /literal/` (1.6 ms) beats its own `.index`, because it lowers a literal-only pattern onto a tuned scan; mutsu enters the full backtracking engine at every one of 655,344 start positions, at **~983 instructions per position** to establish that `chars[i] != 'z'` (ADR-0099 §2.4).

Scaling, failing scans over a repeated 46-char unit (ms):

| subject | mutsu literal | raku | mutsu alternation | raku | mutsu `\w+` greedy | raku | mutsu `:i` literal | raku |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 40 KB | 4 | **0** | 38 | 15 | 19 | 14 | 10 | **0** |
| 160 KB | 15 | **0** | 154 | 86 | 77 | 56 | 39 | 2 |
| 640 KB | 65 | **0** | 612 | 355 | 314 | 228 | 158 | 6 |

The literal columns are an **asymptotic** gap, not a constant-factor one. This is the only such gap ADR-0099 found.

## What to build

One pass over a parsed `RegexPattern`, memoized alongside it in `REGEX_PARSE_CACHE`, deriving:

[...]

PR #8451, opened 2026-09-15
perf(regex): a first-character set rejects a scan position without entering the engine

### LINKED ISSUE #8248: Regex scanning is linear but ~8x rakudo's per-position constant, so mutsu loses on subjects above ~0.5 MB
Found while filling the regex gap in `benchmarks/`. Every regex number mutsu
quotes today is measured on ~100-character subjects, where rakudo's ~200 ms
startup hides everything. On a long subject the picture inverts: mutsu and
rakudo are both linear in subject length, but mutsu's constant per *failing
candidate position* is roughly an order of magnitude larger, so the ratio
degrades monotonically with size and crosses 1.0 somewhere around half a
megabyte.

## Repro

```raku
my $unit = "abc def ghi jkl mno pqr stu vwx yz01 2345 ";
my $big  = $unit x N;
say so $big ~~ / [ 'zzq' | 'yyq' | 'xxq' | 'wwq' ] /;   # fails: every position tried
```

Release build, local box, 2026-09-13, vs rakudo 2026.07 on the same box:

| shape | 40 KB | 160 KB | 640 KB |
|---|---|---|---|
| failing alternation scan | 50 ms / 234 ms (0.21x) | 167 ms / 331 ms (0.50x) | 635 ms / 508 ms (**1.25x**) |
| failing literal scan | 14 / 567 (0.02x) | 25 / 218 (0.11x) | 72 / 227 (0.32x) |
| failing `\w+ 'QQQ'` | 34 / 563 (0.06x) | 101 / 274 (0.37x) | 366 / 459 (0.80x) |
| `$` end anchor | 17 / 221 (0.08x) | 36 / 219 (0.16x) | 113 / 237 (0.48x) |
| failing `<!after \d>` | 15 / 216 (0.07x) | 33 / 269 (0.12x) | 97 / 349 (0.28x) |
| `.match(:g)` | 27 / 499 (0.05x) | 52 / 460 (0.11x) | 133 / 341 (0.39x) |

Subtracting rakudo's startup, rakudo's own scan work is small enough that these
ratios understate the gap: on the failing alternation, 640 KB costs mutsu 635 ms
against roughly 300 ms of actual rakudo work.

Succeeding scans that stop early (`lit-hit`, a class run with give-back) stay at
0.05-0.07x at every size, so this is specifically about the *reject* path: the
work done to advance one position and fail.

## Why it matters

A failing or mostly-failing scan over a large buffer is what real code does --
a log file, a source file, a fetched document. It is also the inner loop of every
grammar: a grammar parse is a long sequence of candidate positions that mostly
fail. So the per-position constant is the single number that decides whether
mutsu can process real-sized input, and nothing in the benchmark suite measured
it until now.

## Next step

Profile one failing scan (`<pattern> ~~` a 640 KB subject) under callgrind and
attribute the per-position cost: candidate construction, the `MatchTarget` span
reads, the LTM ranking that runs before any character is compared, and whatever
per-position allocation remains. A literal or literal-alternation prefix ought to
reduce to a memchr-class scan; measure before assuming that is where the cost is.

## Benchmarks

`benchmarks/bench-regex-long-subject.raku` (added with the regex benchmark set)
is the series for this: one 128 KB subject, nine single-call scans, eight of them
failing. It is the only file in the suite whose cost is dominated by
per-position reject work.

tokuhirom:
Closes #8248.

#8248 recorded the one shape where mutsu lost outright to rakudo: a failing
regex scan over a large subject. Both are linear in subject length, but mutsu's
constant per *failing candidate position* was roughly an order of magnitude
larger, so the ratio degraded monotonically with size and crossed 1.0 around
half a megabyte.

ADR-0099 Stage 1 had already landed the first of the four derived facts the
decision names — a required literal prefix, which turns a literal-headed pattern
into a substring search. Everything *without* such a prefix (an alternation, a
`:i` literal, a character class) still entered the full backtracking engine at
every position: ~983 instructions to establish one rejection (ADR-0099 §2.4).

This adds the second derived fact: the set of characters a match can **begin**
with. A position whose character is outside that set cannot start a match, so it
is rejected by one bit test instead of an engine entry.

## Measurements

Release, 640 KB subject of repeated 46-character units, one scan per row, vs
rakudo 2026.07 on the same box:

| shape | main | this PR | rakudo |
|---|---:|---:|---:|
| failing four-literal alternation | 1,133 ms | **124 ms** | 801 ms |
| failing `:i` literal scan | 246 ms | **16 ms** | 9 ms |
| failing `<[qQ]> 'ZZ'` class scan | 171 ms | **15 ms** | 118 ms |
| `'67-8' \s $` end anchor | 177 ms | **28 ms** | 6 ms |
| `.comb(/ \d+ /)` over the subject | 127 ms | **48 ms** | 461 ms |
| failing `\w+ 'QQQ'` | 716 ms | 619 ms | 340 ms |

The issue's headline row inverts: 1.41x slower than rakudo becomes 6.5x faster.
Regex benchmark suite, three runs each: `bench-regex-split-subst` 378 → 194 ms,
`bench-regex-long-subject` 545 → 254 ms, `bench-regex-global` 253 → 188 ms,
`bench-regex-match` 278 → 243 ms, `bench-grammar-parse-big` unchanged (a
`Grammar.parse` is anchored, so no scan — exactly what the ADR predicted).
Nothing regressed.

`\w+ 'QQQ'` is the row that barely moves, correctly: `\w` admits ~85% of this
subject, so there is nothing for a first-character set to reject. What is left
there is the per-position cost itself, filed as #8450.

## Staying sound

A prefilter's failure mode is being *too narrow* — a set that omits a character
some match really could start with silently drops that match, and no ordinary
`.t` catches it because the unfiltered engine answers the same way. So every
rule is an over-approximation and anything not provably over-approximating
declines, restoring the unfiltered scan. Three over-approximations are
load-bearing:

- **Character classes are probed, not re-derived.** The set is built by asking
  the matcher's own predicate about every character in `0..=0xFF`, so the two
  cannot drift the way a second reading of `ClassItem` would.
  `regex_match_class`'s body is lifted to a free `char_class_matches` for this
  and the method delegates, so there is still one definition. This is also why
  `<alpha>` comes out right with no special case: Raku's `<alpha>` is
  `<+alpha +[_]>`, and probing knows that because the matcher does.
- **`\r` is admitted whenever `\n` is**, because the `CharClass` atom matches a
  `\r\n` pair as the single grapheme `\n`.
- **`:i` always leaves the set open above Latin-1.** U+212A KELVIN SIGN
  lowercases to `k`, so an ASCII `:i` literal is reachable from outside it.

Nullable leading tokens, separator quantifiers, `:m`, assertions, subrules, `.`
and runtime-interpolated literals all decline, as does a set dense enough to
reject almost nothing.

The gate is `tests/regex_prefilter_differential.rs`: every case runs twice, once
normally and once with `MUTSU_REGEX_PREFILTER=off`, and the outputs must be
byte-identical. It grows the long-subject half it was missing (17 new cases),
including a match whose only occurrence begins outside Latin-1 and an `:i` match
that begins at a KELVIN SIGN — the two a narrower set would drop.

## The memoization the first cut got wrong (second commit)

The set was first derived inside `regex_scan_positions` behind a
minimum-remaining-positions threshold. That reasons about the wrong loop:
`.comb`, `:g`, `split` and `.match(:g)` *restart* the scan after every match, so
the threshold let the analysis be charged once per **match** — `.comb(/ \d+ /)`
over the 640 KB subject (54,612 matches) went from 136 ms with the prefilter off
to 208 ms with it on. It is now memoized on the `RegexPattern` next to the
existing `stripped_pattern` memo, which is what ADR-0099 §4 asked for; the
threshold is gone, so short subjects benefit too. `FirstCharSet`'s Latin-1 half
is a 256-bit set (40 bytes per pattern, union is four `|`s).

## Also here

`needs_casefold_expansion` asked `has_multichar_fold` about every character of
the subject on every `:i` match, and that allocated a `Vec` and two `String`s
per character — 655,344 of each on the 640 KB `:i` scan. ASCII case mapping is
1:1, so the common answer is now one range check.

## Follow-ups filed

- #8449 — the literal-prefix analysis declines on a *quoted* literal, because
  `'abc'` parses into a `Group`. Found while explaining why the end-anchor row
  moved; that whole family had been declining the prefilter unnoticed.
- #8450 — the per-position reject constant itself, which is what `\w+ 'QQQ'`
  still measures.

`docs/adr/0099-regex-engine-performance-strategy.md` §8 gains a table of which
Stage 1 derived facts have landed and what each is worth.

## Verification

`cargo fmt --all`, `make lint`, `make test` and `make roast` all run locally and
all exit 0.

---


tokuhirom:
Closing this unmerged: it duplicates work that had already landed on `main`.

`1b6982af8` ("perf(regex): derive a first-character set for the scan prefilter",
merged 09:22 UTC under #8272) implements the same ADR-0099 Stage 1 derived fact,
and does strictly more: it also derives a minimum match length, splits the
analysis into its own module, adds an engagement test that pins the filter stays
applied via `MUTSU_VM_STATS` counters rather than wall-clock, and carries the
same `needs_casefold_expansion` ASCII short-circuit this branch found
independently. Its `FirstSet` and the `class_matches_ignorecase` extraction are
the better versions of what is here.

My mistake in process, not in the work: I claimed #8248 at 10:03 UTC and wrote
the branch without re-reading the sibling issue #8272, which is where the same
change was being made. The claim protocol only ever settles collisions *within*
one issue.

Nothing in this branch is worth rebasing on top of that. Two findings the branch
produced do survive, and are filed on their own:

- #8449 — `required_literal_prefix` still declines on a *quoted* literal,
  because `'abc'` parses into a `Group` rather than a run of top-level `Literal`
  tokens. Confirmed still true on `main` at `466e60225`.
- #8450 — the residue after the first-set: a pattern whose first-set is dense
  (`\w+ 'QQQ'`) still pays the full per-position reject cost.

I am re-measuring #8248's own table against `main` and will report there.

---

Issue #8270, opened 2026-09-13
`REGEX_PARSE_CACHE` allocates its key per call, re-scans the pattern per call, and gives up entirely on any interpolated pattern (re-parsed every iteration)


tokuhirom:
Found while profiling the regex engine for [ADR-0099](https://github.com/tokuhirom/mutsu/blob/main/docs/adr/0099-regex-engine-performance-strategy.md) §2.6/§2.7 (Stage 0 item 5). Three defects in one function, in increasing order of cost.

## 1. The cache key allocates on every match call

`parse_regex` (`src/runtime/regex_parse_ltm.rs:1506`):

```rust
let key = format!("{}\u{0}{}", self.current_package(), pattern);
```

A `String` allocation plus a SipHash of the whole pattern text, on every match — on what is otherwise a pure cache *hit* path. The package is already available as an interned `Symbol` (`current_package_sym()`), so the key wants to be `(Symbol, …)` in a two-level map, or the pattern interned once.

## 2. `regex_pattern_is_static` collects the pattern into a `Vec<char>` per call

`src/runtime/regex_parse.rs:236`, called from `parse_regex` before the cache probe:

```rust
let chars: Vec<char> = pattern.chars().collect();
```

A heap allocation and a full re-encode of the pattern to answer a question that could be answered by a `char_indices()` walk — again on the hit path. It is 9.1% inclusive of a grammar parse.

## 3. A non-static pattern bypasses the cache entirely, and is re-parsed every iteration

The tail of `parse_regex`:

```rust
    return parsed;
}
self.parse_regex_uncached(pattern, RegexParseMode::Match)
    .map(std::sync::Arc::new)
```

Any pattern that fails `regex_pattern_is_static` — i.e. any pattern with an interpolated `$var` / `@var` / `<$rx>` — is re-parsed from source **on every single match call**, with no memoization of any kind.

This is not an exotic case. `benchmarks/bench-regex-match.raku` has two such patterns out of eight (`/ $needle \d+ /` and `/ <$rx> /`), and `parse_regex_uncached` is **4.5% of that whole benchmark** as a result.

Interpolation does make the *parse* depend on runtime values, so the current tree cannot simply be reused. But the work is still cacheable: the interpolation produces a concrete pattern string (`interpolate_regex_scalars`), and *that* string is a perfectly good cache key. A loop that matches `/ $needle \d+ /` with an unchanged `$needle` — the overwhelmingly common shape — would then parse once and hit thereafter.

[...]

Issue #8269, opened 2026-09-13
The smartmatch regex path interns `$_` twice and clones the subject on every `~~` — ~16% of a boolean match


tokuhirom:
Found while profiling the regex engine for [ADR-0099](https://github.com/tokuhirom/mutsu/blob/main/docs/adr/0099-regex-engine-performance-strategy.md) §2.5 (Stage 0 item 4).

## Root cause

`smart_match_inner`'s single-regex arm (`src/runtime/seq_helpers/smart_match.rs:958-972`) saves and restores the topic through the **string-keyed** `Env` API, once per `~~`:

```rust
let text = self.regex_match_text(left);
// Set $_ to the match target so $( $_ ) works inside regex
let saved_topic = self.env.get("_").cloned();
self.env.insert("_".to_string(), Value::str(text.clone()));
let match_result = self.regex_match_with_captures_value(right, &text);
if let Some(v) = &saved_topic {
    self.env.insert("_".to_string(), v.clone());
} else {
    self.env.remove("_");
}
```

Each `insert("_".to_string(), …)` allocates a `String` and drives a `Symbol::intern`, which — even though it is memoized per thread — hashes and compares the whole key string on every hit. Further down the same arm, `self.env.remove("made")` and `self.env.insert(i.to_string(), …)` per positional capture do the same, the latter also allocating a `String` per capture index.

`text.clone()` on the third line is a second, separate waste: a full `String` copy of the subject on every match, purely to put it in `$_` alongside the `&text` that is passed to the matcher.

## Impact

Callgrind, `"a" ~~ /a/` run 20,000 times (release build), 35.6k instructions per call net of startup:

| | inclusive share | calls |
|---|---:|---:|
| `Symbol::intern` | **16.4%** | 40,000 (two per match, ~2,450 instructions each) |
| the regex matcher proper | 8.2% | |

So interning `$_` twice costs twice what matching the pattern costs. Warm, mutsu loses this shape to rakudo 4.9 µs vs 1.78 µs per call — and this is the shape most real Raku code has.

## Expected fix

`src/symbol.rs:762` already has the mechanism and documents exactly this case:

> Fixed names the runtime interns over and over on hot paths. […] For a literal that is re-interned on *every* closure creation or method dispatch, that is pure waste: symbol ids are process-global and append-only, so the id can be computed once and copied thereafter.
> Add a name here only when a profile shows the intern on a per-operation hot path.

So:

[...]

Issue #8265, opened 2026-09-13
`<sym>` lowers to `$<sym>=[…]`, which `regex_pattern_is_static` misreads as interpolation — killing the proto-candidate memo and making every `<subrule>` call re-resolve and re-parse (~70% of a grammar parse)


tokuhirom:
Found while profiling the regex engine for [ADR-0099](https://github.com/tokuhirom/mutsu/blob/main/docs/adr/0099-regex-engine-performance-strategy.md) §2.3. This is the single largest measured cost in grammar parsing today, and it is a caching defect, not an architectural one.

## Root cause

`<sym>` inside a `token value:sym<true> { <sym> }` variant is lowered to the pattern source:

```
:ratchet $<sym>=[true]
```

i.e. a **named capture binding** written in the `$<name>=` spelling. `regex_pattern_is_static` (`src/runtime/regex_parse.rs:232`) then classifies it as non-static, because its `starts_variable_form` treats `<` as a variable form after a `$`:

```rust
fn starts_variable_form(c: char) -> bool {
    c.is_alphanumeric() || matches!(c, '_' | '{' | '(' | '<' | '*' | '?' | '^' | '.' | '!')
}
```

Neither `$<name>=` (a capture binding) nor a bare `$<name>` (a named backreference, parsed to `RegexAtom::NamedBackref`) depends on runtime variable values, so neither is dynamic — but both are rejected.

That verdict then propagates catastrophically, because `resolve_parsed_token_candidates_in_pkg` (`src/runtime/regex/regex_token_resolve.rs:65-98`) declines the memo for the **whole proto** if *any one* candidate is non-static:

```rust
for (sub_pat, sub_pkg, sym_key) in raw {
    if !crate::runtime::regex_parse::regex_pattern_is_static(&sub_pat) {
        return None;          // <- whole proto, not just this candidate
    }
    ...
}
```

Callers (`parsed_subrule_candidates`, `regex_token_resolve.rs:146`) then fall back to the uncached per-call path, so **every single `<value>` reference in the document re-resolves and re-parses all seven variants**.

Confirmed under the debugger — the declining candidate is exactly the `<sym>` one:

```
$ rust-gdb -batch -ex 'break src/runtime/regex/regex_token_resolve.rs:87' -ex run -ex 'print sub_pat' -ex 'bt 3' \
    --args ./target/debug/mutsu tmp/gdbmin.raku
$1 = ":ratchet $<sym>=[true] "
[...]

Issue #8263, opened 2026-09-13
A Match's truthiness goes through user method resolution and forces the lazy Match to materialize — ~9% of every boolean `~~`


tokuhirom:
Found while measuring the regex engine for [ADR-0099](https://github.com/tokuhirom/mutsu/blob/main/docs/adr/0099-regex-engine-performance-strategy.md) §2.4.

## Symptom

`if $str ~~ /rx/` — the cheapest possible consumer of a match, which wants only a Bool — is one of the more expensive parts of the operation.

Callgrind, `"a" ~~ /a/` run 20,000 times (release build), inclusive shares of the whole program:

| | share | calls |
|---|---:|---:|
| `eval_truthy` | **9.38%** | 20,000 |
| — `resolve_method_with_owner_impl` (under it) | 4.75% | 100,000 |
| — `MatchNode::materialize_map` (under it, via a `Once::call_once_force`) | 8.64% | 20,000 |

For comparison, the regex **matcher** in that same run (`regex_match_with_parsed_captures`) is 7.5% inclusive. Turning the Match into a Bool costs more than matching it.

Two separate problems stack here:

1. **Method resolution per truthiness test.** `eval_truthy` on a `Match` instance runs the full user-method-overload lookup (5 `resolve_method_with_owner_impl` calls per test) looking for a user `Bool`. A `Match` produced by the engine has no user overrides.
2. **The lazy `Match` is materialized by the test.** [ADR-0016](https://github.com/tokuhirom/mutsu/blob/main/docs/adr/0016-span-based-captures-and-lazy-match.md)'s whole point is that a `Match` whose captures are never read never builds its map. `MatchNode::materialize_map` firing once per boolean test defeats that for the single most common consumer.

## Repro / measurement

```raku
my $s = "a";
my $t = now; my $n = 0;
for ^200000 { $n++ if $s ~~ / a / }
say "per-call-ns=", ((($e = now) - $t) * 1e9 / 200000).round;
```

mutsu 5,793 ns/call against a 1,169 ns/call loop baseline (`$s.chars == 1` in the same loop); rakudo 1,981 ns against a 163 ns baseline. Whole-call instruction count is 29,400 Ir.

```
valgrind --tool=callgrind --callgrind-out-file=cg.out ./target/release/mutsu <20k-iteration version>
callgrind_annotate --tree=caller cg.out
```

## Expected fix

[...]

PR #8464, opened 2026-09-15
perf(regex): derive the scan prefilter's first-set through a subrule call

tokuhirom:
ADR-0099 Stage 1's fourth slice. The unanchored-scan prefilter derived a first-character set for every pattern shape except one: a `<subrule>` made it decline outright.

Constraint 3 of the ADR does not ban deriving through a rule name; it attaches a condition — *"keyed by invocant package and `TOKEN_DEFS_GEN` (dynamic override via `H is G` is legal), **or decline**"*. The first three slices (#8285, #8446, #8457) took the decline, and that is a cliff rather than a shortfall: a rule name *anywhere in a pattern's leading run* sinks the whole derivation, so an alternation of two literals was narrowed to two candidate characters when written inline and walked at every position of the subject when written behind a rule name.

## Measured

Failing scans over a 135,000-character subject, release, against `MUTSU_REGEX_PREFILTER=off` (the status quo for every one of these, since all four declined before) — same binary, on/off:

| scan | before | after | warm rakudo |
|---|---:|---:|---:|
| `~~ / <G::kw> /`, `token kw { 'zzzq' \| 'kkkz' }` | 145.8 ms | **5.5 ms** | 675.6 ms |
| `~~ / <G::outer> /`, `token outer { <inner> }`, `token inner { 'zzzq' }` | 61.1 ms | **0.7 ms** | 892.0 ms |
| `~~ / <G::sigil> <[a..z]> /`, `token sigil { <[$@%&]> }` | 53.4 ms | **1.0 ms** | 608.7 ms |
| `~~ m:g/ <G::kw> /` | 153.8 ms | **5.4 ms** | — |

The inline form of the first row, `~~ / 'zzzq' | 'kkkz' /`, costs 4.7 ms on the same subject — which is the point: hiding a pattern behind a rule name no longer changes its complexity class. Rakudo uses an NFA for start filtering and evidently does not lower a rule call onto it either.

**What this slice cannot claim, recorded so it is not over-sold:** a scan the prefilter *does* narrow bottoms out at a fixed ~4 ms per call on a 630,000-character subject regardless of how many positions survive — an anchored `/ ^ 'zzzq' /`, which scans zero positions, costs the same. That floor is per-call subject materialization, i.e. ADR-0099 Stage 0's territory (#8262 / #8263 / #8265 / #8269 / #8270), not Stage 1's.

## How constraint 1 is satisfied

Resolution goes through `resolve_parsed_token_candidates_in_pkg` — the same memoized `(pkg, name)` + `TOKEN_DEFS_GEN` table the matcher itself resolves `<x>` through — and each candidate body is walked in the package that **defined** it, the rule `subrule_candidate_ends` matches under. So there is no second reading of the rule registry to drift from the first, and a redefinition invalidates both at once. Re-deriving "which body does `<x>` mean here" independently is exactly the change that would reintroduce the drift risk, and the drift would be a silently dropped match rather than an error.

## Why the memo had to split

Such a derivation is no longer a pure function of the pattern, and patterns are shared **by source text** through the regex parse cache — the two `/ <x> /` in two different packages are *one* cached pattern. Storing one package's first-set in the pattern-keyed slot would have answered for both and silently dropped every match in the second. Patterns naming no rule keep that slot and their existing path (no lock, no generation read); the rest go to a new package-and-generation-keyed memo. The test that picks between them is exhaustive over `RegexAtom` on purpose — under-reporting a container there is a wrong answer, not a slow scan, so a newly added atom has to fail to compile.

## Declines, each unsound rather than merely imprecise

A call with arguments and `<::(EXPR)>` (resolved per call); a lexical `Regex` reference `<&re>` / `<&$re>` (resolved against the caller's scope, which the key does not carry); a body whose text interpolates a runtime value (re-parsed per call — tested with the same predicate the call-graph analysis uses, which correctly ignores a sigil appearing only inside a `{ … }` block, so `token part { \w+ { $n++ } }` stays analyzable); a name no rule answers to, or that a grammar method also answers to; a rule body opening with a code block (ADR-0009: it must keep running once per start position); and **left** recursion — a rule reached from itself before anything is consumed can only be walked to a *subset* of what can match, the one unsound direction. Right recursion keeps its derivation, since the first character is already fixed when the recursive call is seen.

The required literal prefix and required inner literal still decline on rule names, deliberately: both claim something about *text*, and the inner literal's decline on anything that can run code is stronger than the first-set's, for the reason #8457 recorded.

## Testing

- `t/regex/regex-scan-prefilter-subrule.t` — 26 assertions on observable behaviour, **25 of which rakudo agrees with verbatim** (the exception is the left-recursive rule, which rakudo does not terminate on at all).
- `tests/regex_prefilter_differential.rs` — eleven new cases comparing each shape against `MUTSU_REGEX_PREFILTER=off`, including the two-packages-one-pattern case a pattern-keyed memo would fail.
- `tests/regex_prefilter_engagement.rs` — both directions on the `MUTSU_VM_STATS` counters: a subrule-led failing scan stays sub-linear in subject length, and every shape above still declines. A new `subrule_derivations=(resolved=N declined=N)` counter makes "it stopped engaging" observable rather than merely slower.
- `cargo fmt --all`, `make lint`, `make test` and `make roast` all run locally and green before opening this.

## Also here

`regex_prefilter.rs` and `regex_prefilter_analysis.rs` went over the 500-line limit, so: the two memos move to `regex_prefilter_memo.rs`, the per-atom first-set constructors to `regex_prefilter_firstset.rs` beside the type they build, and the subrule resolution plus its walk to `regex_prefilter_subrule.rs`. ADR-0099 §8 no longer says "Nothing implemented" — it now names the four Stage 1 slices and what the issue stays open for.

Refs #8272 — the issue stays **open**: NFD-aware first-sets for a *scoped* `:ignoremark`, and §5's NFA over the declarative prefix, remain.

https://claude.ai/code/session_01K1Y9aSDLbayfbLbpELijSf

---
_Generated by [Claude Code](https://claude.ai/code/session_01K1Y9aSDLbayfbLbpELijSf)_


tokuhirom:
## `test-suites` is red, and it is not this PR's

**Failing check:** `test-suites` (and the `test` aggregator, which fails when its half does — one failure, not two).

```
REGRESSION: whitelisted 'Digest	ripemd.t' did not pass this run
```

### Why it is not this PR's

1. **`Digest::RIPEMD` contains no regex at all.** This PR derives the regex scan prefilter's first-set through a `<subrule>`; the module uses `await` + `start` and nothing else. The only match `grep -E 'rx|m/|~~ */|token |rule |regex '` finds in `modules/Digest/lib/Digest/RIPEMD.rakumod` is a URL in a comment. There is no path from this diff to that file.

2. **The same check is red on `main`.** Two push runs since the whitelisting are `test-suites: failure` — [`b8300f9c`](https://github.com/tokuhirom/mutsu/actions/runs/34967608738) (12:13) and [`d3978843a`](https://github.com/tokuhirom/mutsu/actions/runs/34970578327) (12:43). The second of those is this branch's rebase base.

3. **The entry was whitelisted on a local measurement with no CI-runner margin.** `Digest ripemd.t` entered `batteries-whitelist.txt` at 11:56 today, in #8459 — a perf PR whose own description says it plainly:

   > `Digest::RIPEMD`'s `t/ripemd.t` — the one bundled-battery test file too slow to whitelist (**~113.7s against the batteries gate's hard 120s budget**) — drops to **~73s local**, real margin under budget, and joins the whitelist.

   The margin was measured on a 4-core dev box, not on the CI runner. `ripemd.t`'s last case is `rmd160('a' x 1_000_000)` — ~15,625 compression blocks, one `start` each (#7571) — so it is the single most runner-speed-sensitive file in the gate, and the gate's budget is a hard `timeout 120`.

### What I have and have not done

I have **not** ported a fix, because none exists yet and the two candidate fixes both belong on `main` rather than in a regex PR:

- make `Digest::RIPEMD` fast enough to have real margin (the `start`-per-block shape #7571 named), or
- drop the entry from `batteries-whitelist.txt` until it does.

The second weakens a gate, which is not a call I will make unilaterally inside an unrelated PR. Raising it separately.

I have not spent the one permitted re-run: a re-run cannot help here, since the file is over budget on this runner class rather than unlucky.

This branch has been rebased onto `d3978843a` so it carries `main`'s current whitelist; that changes nothing about the above, and the gate will stay red for this PR exactly as long as it stays red for `main`.

---

PR #8468, opened 2026-09-15
fix(oo): attributive parameter in a sub nested in a method binds self.attr

### LINKED ISSUE #8452: An attributive parameter in a sub nested in a method: rejected at parse time, and the binder would not write the attribute anyway
Found while working the parse-failure index (#7954), from `PDF::Content::Text::Box`. Two gaps in one line, which is why this is `todo:deep` rather than a ticket: relaxing the parse check alone turns a load failure into a **silently wrong answer**.

## Repro

```raku
class C {
  has $!t = "orig";
  method m {
    sub s($!t) { say "inside: $!t" }
    s("new");
    say "after: $!t";
  }
}
C.new.m;
```

```
$ raku  repro.raku
inside: new
after: new

$ mutsu repro.raku
Runtime error: X::Syntax::NoSelf: Variable $!t used where no 'self' is available
```

## Gap 1 — the parse-time rejection is per-routine where raku's rule is lexical

`reject_attr_params_in_sub` (`src/parser/stmt/sub/traits.rs`) rejects `$!x` / `$.x` parameters in **every** `sub` signature, on the reasoning that "subs don't have self". Raku's rule is lexical: a plain `sub` nested in a method body closes over the method's invocant, so `self` *is* available there. Measured against rakudo:

| written | raku | mutsu |
|---|---|---|
| `sub s($!t) { }` at top level | error | error ✓ |
| `class C { has $!t; sub s($!t) { } }` | error | error ✓ |
| `class C { has $!t; method m { sub s($!t) { } } }` | **accepted** | error ✗ |
| `class C { has $!t; method m { class D { sub s($!t) { } } } }` | error | error ✓ |
| `class C { has $!t; method m { role R { sub s($!t) { } } } }` | error | error ✓ |

So the condition is "some enclosing scope is a method/submethod body, with no package body in between" — a scope flag inherited by `push_scope`, set when a method body's scope is pushed and cleared when a `class`/`role`/`grammar`/`package` body's scope is pushed. That much was prototyped and is straightforward (a `self_available: bool` on `LexicalScope`, a `method_block` and a `package_body_block` alongside `routine_block`/`block` in `stmtlist.rs`, and the check becoming conditional). It was **deliberately not landed**, because of gap 2.

## Gap 2 — an attributive parameter does not write the attribute from a nested sub

With the parse check relaxed, the repro parses and then prints:

```
inside: orig
after: orig
```

An attributive parameter *means* "bind this parameter to `self.attr`" — the value belongs to the OBJECT, not to a frame slot. mutsu's binder only inserts it into the callee frame's env under the key `!t`. In a **method** that happens to be right, because the method's frame is where `$!t` is read from; in a nested named `sub` it is a shadowing local that nobody ever reads and the object is never touched.

Three neighbouring behaviours narrow it:

- `class C { has $!t; method m($!t) { } }` — works, and persists to the object (`method setit($!t) { }` then reading `$!t` from another method gives the bound value).
- `class C { has $!t = "orig"; method m { my $b = sub ($!t) { }; $b("new"); say $!t } }` — prints `new`. The **anonymous** sub literal writes through, because its `env.insert` lands in the captured (shared) method env rather than a fresh frame. That is an accident of the dual store, not the rule.
- `class C { has $!t; method m { sub s() { $!t = "viaNamed" }; s(); say $!t } }` — prints `viaNamed`. So an ordinary *assignment* to `$!t` from a named nested sub already reaches the attribute; only the parameter-binding path does not.

The honest fix is in the binder (`src/runtime/types/binding_signature.rs`): a parameter whose name starts with `!`/`.` must be bound by writing the attribute on `self`, on the same path `$!t = v` already takes, rather than by inserting a frame-local. That would also make the method case stop depending on the frame-is-the-attribute-env coincidence.

## Why it matters

`PDF::Content::Text::Box` writes it in a `Proxy`:

```raku
sub FETCH($) { $!text };
sub STORE($, $!text) {
    my @chunks = $obj.comb($!text);
    ...
}
```

That is the last construct blocking `PDF::Content::Text::Box` and `PDF::Content::Text::Block` in the `ecosystem/` record (the distribution's other modules were unblocked by #7954's current batch). Both halves have to land together: with only gap 1 fixed the module loads and the `Proxy`'s STORE silently discards its value.

Refs #7954

tokuhirom:
## Summary

Fixes #8452 — two gaps in one line:

1. **Parse-time rejection was per-routine, not lexical.** `reject_attr_params_in_sub` rejected `$!x`/`$.x` parameters in every `sub` signature, but Raku's rule is lexical: a plain `sub` nested directly in a method/submethod body closes over the method's invocant, so `self` *is* available there. `LexicalScope` now tracks a `self_available` flag — set on a method/submethod body's scope (`stmt::method_block`), cleared on a nested class/role/grammar/package body's scope (`stmt::package_body_block`, `parse_block_body_no_self`), and otherwise inherited by every scope nested in between (unlike the existing `is_routine_body` flag, which resets on every `push_scope`). Matches rakudo exactly across the issue's whole test table (top-level sub / class-body sub / method-nested sub / class-in-method / role-in-method).

2. **The binder wrote a throwaway local, not the attribute.** Even with the parse check relaxed, the parameter's bound value only ever reached the sub's own per-call frame local (env/locals keyed by the twigil name) — never `self`'s actual attribute storage. A method's own attributive parameter already gets mirrored into `self`'s shared attribute cell after binding (`mirror_attributive_params_to_cell`, called from `call_compiled_method`); this generalizes that helper to take `param_defs` directly (instead of a `MethodDef`) and calls it from a plain sub's call path too (`call_compiled_function_named_inner`). Two positional fast-dispatch eligibility gates (`is_light_call_eligible`, `is_positional_light_call_eligible`) were bypassing that call path entirely for simple signatures, so both now exclude a signature carrying an attributive parameter — forcing such a call through the path that performs the mirror.

```raku
class C {
  has $!t = "orig";
  method m {
    sub s($!t) { say "inside: $!t" }
    s("new");
    say "after: $!t";
  }
}
C.new.m;
# before: inside: orig / after: orig  (silently wrong)
# after:  inside: new  / after: new
```

## Test plan

- [x] `t/lang/parsing/syntax-noself.t` — extended with the issue's full parse-acceptance/rejection table (top-level, class-body, method-nested, class-in-method, role-in-method).
- [x] `t/oo/attribute/attributive-param-nested-sub.t` (new) — runtime persistence: direct call, reading `$!x` from inside the nested sub, repeated calls across distinct receivers (exercises both the general and fast-dispatch cache paths), and a neighbouring-behavior regression guard for plain (non-param) assignment from a nested sub.
- [x] `cargo fmt --all`
- [x] `make test` (`prove t/`, 4335 files / 46300 tests) — clean
- [x] `make roast` — the only 4 failures (`roast/6.c/S32-io/file-tests.t`, `roast/S16-filehandles/filetest.t`, `roast/S16-io/eof.t`, `roast/S32-io/IO-Socket-Async.t`) are the documented container-only exceptions (uid 0 file-permission tests, restricted `/proc`, sandboxed network) — not a regression.

https://claude.ai/code/session_01RZKspCKXhDnjrza1p1Ywxr

---
_Generated by [Claude Code](https://claude.ai/code/session_01RZKspCKXhDnjrza1p1Ywxr)_


tokuhirom:
CI's `test-suites` job failed, but only in its **Bundled-library test suites** step (the ecosystem battery gate), specifically `Digest`'s `t/ripemd.t`:

```
ripemd.t   FAIL(ok=8/?,notok=0,todo=0)
REGRESSION: whitelisted 'Digest/ripemd.t' did not pass this run
```

This is not this PR's failure — this diff touches parser lexical-scope tracking (`self_available`) and the VM's sub/method call-binding path (attribute-twigil parameter mirroring), and does not touch anything on RIPEMD's hot loop (`type_matches_value`, native-int constraints, or any code the subs in `t/ripemd.t` exercise — none of them declare an attributive `$!x`/`$.x` parameter, the only signature shape my eligibility-gate change excludes from the fast dispatch path).

Per [`news/2026-09/ripemd-positional-bind-failover-fast-path.md`](https://github.com/tokuhirom/mutsu/blob/main/news/2026-09/ripemd-positional-bind-failover-fast-path.md), this file's own margin against the 120s batteries-gate timeout is fairly tight (~73s measured in that session), so a slower/busier shared CI runner plausibly pushes it over on its own — both the earlier `TAP tests (prove t/)` and `Roast tests (make roast)` steps in the same job passed cleanly on this same commit.

The **`test-suites`** job's failed steps have been re-run once to confirm (`rerun_failed_jobs` on run `34974133896`); watching for the result.

---

PR #8470, opened 2026-09-15
fix(modules): replay a module's declared-package grant for later importers

tokuhirom:
## Distribution

`Acme::Cow` 0.2 (`axis: pure`, 1 test file) — baseline files **0/1 → 1/1**, assertions **1/4 → 4/4**.
Ledger record goes `red` → `green`, `cmp: regression` → `parity`.

## The failure

```
ok 1 - Acme::Cow module can be use-d ok
# You planned 4 tests, but ran 1
Could not find symbol 'Cow::cow'
  in block <unit> at t/01-tests.rakutest line 8
```

The distribution's `lib/Acme/Cow.rakumod` is `use`d as `Acme::Cow`, but the package it declares is
`Cow`:

```raku
unit module Cow;
class basic { ... }
class cow is basic { ... }
```

So its classes are `Cow::cow`, `Cow::basic`, ... and nothing about the module's *name* says so.

## Root cause

The #7797 visibility gate only lets a compunit write `Pkg::thing` when that compunit `use`d `Pkg`
itself, so a package reached transitively through somebody else's `use` does not leak. The grant
behind the gate is computed while the module loads (`granted_packages` in `run_modules.rs`), and it
is computed correctly — it covers the module's own name, the `unit module`/`unit class` package the
file declares, and every type registered under that prefix.

A re-`use` of an already-loaded module never re-runs that load. It takes the short-circuit in
`use_module_with_tags_inner`, which had its own miniature grant: `module` and its first
`::`-segment, and nothing else. It could not do better, because the packages a module declares are
not derivable from the name it is `use`d by — only the load knows them.

While the declared package matches the file name that gap is invisible, which is why it survived.
`Acme::Cow` is the case where it is fatal, and `Test`'s `use-ok` is what makes it fire routinely:
`use-ok` loads the module from inside an `EVAL`, so the **first** importer is an EVAL unit and the
script's own `use Acme::Cow;` is the already-loaded no-op. The script therefore never received a
grant for `Cow` at all. Two plain `use` statements were fine; `use-ok` followed by `use` was not.

Minimal reduction (independent of the distribution):

```raku
# lib/Acme/Cow2.rakumod  --  unit module Cow2;  class cow { }
use Test;
plan 2;
use-ok 'Acme::Cow2';
use Acme::Cow2;
ok Cow2::cow.new, "qualified class from unit-module reachable after use-ok";
```

| | output |
|---|---|
| raku | `ok 1`, `ok 2` |
| mutsu (before) | `ok 1`, then `Could not find symbol 'Cow2::cow'` |

This is the same first-load-inside-an-EVAL shape as #7806, which fixed the module's own bare
package-name binding going missing. The package grant is a second thing that load carried and the
no-op path could not reconstruct.

## Fix

Record the grant per module at first load (`module_granted_packages`) and replay it for later
importers on the already-loaded path. The recorded set is exactly what the first load installed into
`compunit_visible_packages` — the declared package, the module name, the types under that prefix,
and each of their top-level segments — so a second importer now sees precisely what the first one
did, and nothing more.

No special-casing of the distribution: the change is in the general module-load path and is keyed by
nothing but what the loaded file declared.

## Testing

- New pin: `t/modules/qualified-package-visible-after-use-ok-first-load.t`, with a fixture at
  `t/lib/AcmeCowShape/Acme/Cowish.rakumod` declaring `unit module CowNS;` — the distribution's shape,
  reduced. Passes under both mutsu and rakudo.
- `make lint` green.
- `make test` green — 4337 files, 46316 tests, `Result: PASS`.
- `make roast`: the only failures are the four documented container-only ones
  (`6.c/S32-io/file-tests.t`, `S16-filehandles/filetest.t`, `S16-io/eof.t`,
  `S32-io/IO-Socket-Async.t`), matching `docs/agent-environments.md` by name and by failure shape —
  this container runs as `uid 0` with a sandboxed network.
- Ledger re-measured with the release binary at `aeda03e3d`;
  `ecosystem/index-snapshot.json` deliberately untouched.

News: `news/2026-09/module-declared-package-grant-replayed-for-later-importers.md`

---


tokuhirom:
`test-suites` failed on the battery gate, not on `t/` or roast:

```
Digest  ripemd.t   FAIL(ok=8/?,notok=0,todo=0)
REGRESSION: whitelisted Digest/ripemd.t no longer passes
```

**This is a timeout, not a regression from this PR**, and the counters say so directly: `notok=0`
(no assertion failed) with the plan printed as `?` (the `1..9` line was never reached). The file was
killed part-way through test 9 — the largest RIPEMD vector — by `battery-testsuite.sh`'s per-file
`timeout 120`.

Evidence it is not this diff's:

- `t/ripemd.t` **passes** against this branch's release binary: 9/9 `ok`, exit 0.
- It takes **137s** on this container — already over the gate's 120s budget with or without this
  change. The CI runner is normally fast enough to land under the line; this run it did not.
- The only thing this PR touches is the module-load path, and loading `Digest::RIPEMD` costs
  **30ms** of that 137s (`use Digest::RIPEMD` alone: 0.030s; one small `rmd160`: 0.033s). The
  remaining ~137s is hashing the large test vectors, which this diff does not go near. Its
  contribution is bounded well under 0.1% of the runtime — it cannot move a file across a 120s line.

No fix for it exists to port. The real issue is that `ripemd.t` sits at ~114% of the gate's per-file
budget, so it is one slow runner away from red on any PR; making mutsu's RIPEMD faster (or the
budget honest) is its own piece of work and would widen this PR, and skipping or quarantining the
test is not an option. Re-running the failed job once, as the one re-run for a failure that is not
this PR's.

---

Issue #8466, opened 2026-09-15
`OUR::«…»` — an interpolating word quote as a package-symbol name is not parsed


tokuhirom:
Reached through `Data::Record` while working the #7954 parse-failure index. `Data::Record::Lifter` installs its exported operator by binding a computed symbol name in `OUR::`:

```raku
my package EXPORT {
    package DEFAULT {
        OUR::«'&infix:<@~~>'» := Data::Record::Lifter;
    }
}
```

## Repro

```raku
my $x = 5;
OUR::«"\$y"» := $x;
say "ok";
# raku:  ok
# mutsu: ===SORRY!=== ... expected expression statement or identifier after '::'
```

The non-interpolating spelling already works, which is what narrows it to the quote form:

```raku
OUR::<$z> := 5; say "ok";   # both: ok
```

## Why it matters beyond the one dist

`Data::Record::Lifter` is where `infix:<@~~>` comes from, so the three modules that use that operator (`Data/Record/{List,Map,Tuple}.pm6`) cannot load either — the import never happens. Four of `Data::Record`'s twelve modules are blocked behind this one construct; the other constructs in that distribution were fixed in the #7954 batch this was found in.

Refs #7954

tokuhirom:
Claiming: claude/issue-8466-our-interp-wordquote

---
_Generated by [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code)_

tokuhirom:
Releasing: claude/issue-8466-our-interp-wordquote

Fix: PR #8479.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #8465, opened 2026-09-15
A Slip read out of a container does not slip, and a scalar-assigned one is not itemized


tokuhirom:
Found while landing the stacked `||` Slip contextualizer for #7954. A `Slip` is the one value that flattens out of whatever container it was read from — that is its whole purpose — but mutsu only flattens one that reaches a list *directly*.

## Repro

```raku
my $x = slip(5, 6);
say $x.^name;            # both: Slip
say (1, $x, 2).elems;    # raku: 4        mutsu: 3
my @a = 1, $x, 2;
say @a.raku;             # raku: [1, 5, 6, 2]   mutsu: [1, slip(5, 6), 2]

my %h;
%h<a> = slip(5, 6);
say (1, %h<a>, 2).elems; # raku: 4        mutsu: 3

say $x.raku;             # raku: $(slip(5, 6))  mutsu: slip(5, 6)
```

The same Slip written inline, returned from a sub, read through `.self`, or wrapped in parens all flatten correctly, so the gap is specific to a **container read**:

```raku
my @ok = 1, (|(5, 6)), 2;                       # both: [1, 5, 6, 2]
sub f() { slip(5, 6) }; my @ok2 = 1, f(), 2;    # both: [1, 5, 6, 2]
my @ok3 = 1, ($x), 2;                           # both: [1, 5, 6, 2]
my @ok4 = 1, $x.self, 2;                        # both: [1, 5, 6, 2]
for (1, $x, 2) { .say }                         # both: 1 5 6 2
```

## Root cause

Two independent halves.

1. **List construction.** `Interpreter::exec_make_array_op` (`src/vm/vm_data_ops.rs`) handles a `WrapVarRef`-tagged element *first*, storing the variable's shared `ContainerRef` cell so the list aliases the variable, and `continue`s — so the `ValueView::Slip(items) => elems.extend(...)` arm below it is never reached for a variable read. The comment there ("A ContainerRef is a scalar item, so it never flattens") is right for every value except a Slip. A hash-element read (`IndexAutovivifyLazyTerminal`) arrives wrapped the same way.

[...]

Issue #7984, opened 2026-09-11
A type capture on the invocant (`method m(::T Foo:D: $x --> T)`) does not parse, and the bare `::T:` form mis-binds T


tokuhirom:
Found reducing the `CRDT` row of the parse-failure index (#7954). `CRDT.rakumod:13` is the last construct keeping that distribution's own module from parsing:

```raku
unit role CRDT;
method merge(::T CRDT:D: $ --> T) { ... }
method copy(::T CRDT:D: --> T)    { ... }
```

A **type capture followed by a nominal invocant type**. Two separate problems sit under it.

## 1. `::T Foo:D:` does not parse at all — and in a role it silently drops the method

```
$ mutsu -e 'class Foo { method m(::T Foo:D: $x --> T) { $x } }; say "ok"'
===SORRY!=== Error while compiling -e
Confused. expected statement: expected expected statement: expected expression statement or expression after infix operator ...
$ raku  -e 'class Foo { method m(::T Foo:D: $x --> T) { $x } }; say "ok"'
ok
```

`role` is worse than `class`: the parse failure is swallowed and the method is dropped, with no diagnostic at all.

```
$ mutsu --dump-ast -e 'role R { method m(::T R:D: $x --> T) { 1 } }'
...
    RoleDecl {
        name: Symbol(1: "R"),
        body: [],          # <- the method is gone
$ mutsu -e 'role R { method m(::T R:D: $x --> T) { 1 } }; class C does R {}; say C.new.m(3)'
No such method 'm' for invocant of type 'C'
$ raku  -e 'role R { method m(::T R:D: $x --> T) { 1 } }; class C does R {}; say C.new.m(3)'
Type check failed for return value; expected C but got Int (1)   # i.e. the method IS there
```

**Where it goes wrong** (`src/parser/stmt/sub_param/param_inner.rs`, the "Type-capture parameter: `::T $x` or bare `::T`" branch, around line 295): the branch consumes `::T`, sets `type_constraint = Some("::T")`, and then only knows how to continue on a sigil, a `=` default, a closing bracket, or a `:` marker. A *type name* next (`Foo:D:`) matches none of them, so the branch falls through leaving `type_constraint` already `Some`, which makes the general type-constraint block below it (guarded on `type_constraint.is_none()`) skip. `Foo:D:` is therefore never consumed and the parameter list fails.

## 2. Even the already-accepted `::T:` invocant capture mis-binds

[...]

Issue #7954, opened 2026-09-11
Parse-failure index: 56 `blocked_load` distributions, located to a file and line


tokuhirom:
Found by P5 triage of the first full-corpus ecosystem sweep (#7785). **`expected statement …` is the largest single bucket in `blocked_load` — 56 of 393 records** — and its error text says only "the parse failed", never what failed. This issue turns it into a located, groupable table so the rows can be picked off as ordinary tickets.

`todo:deep` because it is an index, not one fix: 56 records span roughly twenty distinct constructs. Individual rows graduate to their own `todo:ticket` as they are worked; three are already confirmed in isolation below and are ready to pick up.

## How the table was produced (reproducible in ~2 minutes)

A parse error needs **no dependency closure** — `--dump-ast` reaches it before any `use` is resolved. So for each record: fetch the tarball, then run `mutsu --dump-ast` on each module the META6 `provides` names, and read the `at <file>:<line>` and the `------>` echo out of the error. Script used: `tmp/p5/parse-locate.py` in the session (worth promoting to `scripts/` if this is repeated).

Caveats, both visible in the table: when the parser cannot recover it reports the **outermost** construct, so a row naming `class Foo {` or a bare `{` is a container and the real failure is deeper in that file (bisecting the file settles it). And 13 of the 56 do not fail on any module in `provides` — their failure is in a *dependency's* module, so they need the sweep's own closure (`scripts/ecosystem-sweep.py --only <dist>`).

## Confirmed in isolation — ready to fix

Three constructs were reduced to one-liners and diverge on their own:

**1. `my token X is export { … }`** (`Collection`, `Number::More`)

```
$ mutsu -e 'my token all-bases is export { \d+ }; say "ok";'
===SORRY!=== Error while compiling -e
$ raku  -e 'my token all-bases is export { \d+ }; say "ok";'
ok
```

**2. A `where` clause holding a whatever-curry with a prefix — `!so *`** (`Linux::NFTables`)

```
$ mutsu -e 'sub f(Bool $a where !so *) { "ok" }; say f(False);'
===SORRY!=== Error while compiling -e
$ raku  -e 'sub f(Bool $a where !so *) { "ok" }; say f(False);'
ok
```

**3. Not a parse error at all:** `BEGIN` writing an `our`-scoped `%`/`@` is silently lost — filed separately as #7953, reached here through `PDF::Content`.

[...]

PR #8473, opened 2026-09-15
fix(parser): stacked Slip contextualizers, type captures after a type, and `let`/`temp` as a term

tokuhirom:
Six constructs off the `blocked_load` parse-failure index (#7954), re-derived from the current `ecosystem/dists/` records rather than the issue's snapshot table. **`Data::Translators` now parses every module its META6 `provides` names**, and `Data::Record` loses four of the five files it was failing on — the rest are all downstream of one construct, filed as #8466.

## The contextualizers stack for `|` as well as `&`

`&&(0, 1)` became `&(&(0, 1))` in #8298; the `|` half was never done, in either the bare or the parenthesized form, so a term-position `||` was not a term in any reading. `Data::Translators` opens an assignment with one:

```raku
my $isHTML =
        || $_.trim.starts-with('<math') && $_.trim.ends-with('</math>')
        || $_.trim.starts-with('<table') && $_.trim.ends-with('</table>');
```

**The trap is what that looks like.** It reads as a leading-`||` boolean chain and **is not one**: rakudo parses it as `(||(A && B)) || (C && D)` — a one-element Slip on the left, which the distribution only gets the answer it wants from because a one-element list is truthy. `|| False || True` is `slip(Bool::False,)`, not `True`. Implementing the idiom it resembles would be a private dialect that silently disagrees on the value, so what landed is rakudo's own reading.

There is no ambiguity with the infix `||`: an infix is only ever looked for once a term has been parsed, so a `||` reached in term position has no left operand and cannot be one.

It *was* protecting something else by accident, though, and `make test` caught it (second commit). A **subscript** opening with `||` is the 6.e dimension splat (`%h{||@dims}` — the list supplies the semicolon dimensions), lowered to a `MultiDimIndex`. The assignment-lvalue parser has its own subscript reader that parses the body as an ordinary expression, and it used to *fail* on `||` and hand the subscript back to the postfix path. Once `||` became a term that reader started succeeding, silently degrading `(%h{|| @dims} = 42, 666)` to a one-dimensional slice. It declines a `||` body explicitly now, which is the rule that was being relied on all along.

## A parameter's type capture may be written after its nominal type

`::T Int:D $x` parsed; `Int:D ::T $x` did not — the branch that recognizes a variable after a type read the capture's **first colon** as the named-parameter marker and was left with `:T $x`. `Data::Record` writes both orders in one signature:

```raku
multi method new(::?CLASS:_ ::THIS: List:D ::T $original is raw, …)
```

`ParamDef` has carried the two halves separately since #7984 (`type_constraint` + `type_capture`), so only the grammar was missing. Rather than duplicate the capture branch, the type-first order delegates to it and hangs the nominal half back on the result — so every shape that branch already knows (a bare capture, an invocant marker, a default, traits, a `where` clause) works in this order too, for free.

## `let` and `temp` are statement prefixes over an assignment — three spellings were missing

`Data::Record::Map` writes all three in one line:

```raku
MapIterator.new(THIS, 'bounded', WRAP, 'push', let %!record .= push: @values).sink-all;
```

**The variable may be an attribute.** `let $!x` was not recognized at all, so it came back as the bareword `let` followed by an ordinary assignment — the save silently dropped — and the term form `(let %!record)` was a hard parse error. `temp $!x` was worse: `$!` was read as the *error variable*, so `temp $!x = 2` was a `temp $!` followed by an assignment to a stray lexical `x`, and the attribute was never written at all (`say $!x` answered `1` where rakudo answers `2`).

Fixing the parse exposed the half below it. An attribute's source of truth is `self`'s shared attribute cell — `exec_get_local_op` re-reads it on every `$!x` — so restoring the method frame's mirror slot and the `env` entry left the temporized value standing past the scope. The restore writes the cell too now; the helper is a no-op when `self` is not an instance, so no other `temp`/`let` target pays for it.

**The assignment may be a compound one.** Only plain `=` had a branch, so the `.= push` that followed became a *topic* dot-assign: the push landed on `$_` and the container was left untouched (`let %!r .= push: (a => 1); say %!r` answered `{}`).

**The whole thing may stand as a term, including as an argument.** Only the parenthesized form `(let $x = 5)` was a term, and that path demands the `)` immediately after. `let`/`temp` join the control keywords `identifier_call` already admits as terms; the `ws1` guard keeps a user-defined `sub let` callable as `let()`.

`make roast` caught the consequence of that last one (fourth commit). `undefine temp $b` was only ever `Call("undefine", [Call("temp", [$b])])` *because* `temp` was not a term and fell back to being read as a call, so the compiler's `undefine temp $var` lowering stopped matching, and the statement parser's swallowed terminator let the enclosing listop read its argument list across the `;` (`undefine temp $b; say 2` became `undefine(temp $b say 2)`). Both shapes are lowered now, and the term arm calls `restore_do_stmt_terminator` exactly as the `do` arm does.

## A bare `+` is the anonymous one-arg-rule slurpy

`proto MAKEOP(Str:D, +) {*}` (`Data::Record::Test`). Only the sigilled (`+@a`) and sigilless-name (`+foo`) forms were parameters at all, so a signature holding a bare `+` failed at its closing paren. It binds the same anonymous array the `+@` path already mints.

## A named parameter's alias is a key, not a variable — and the key has no sigil

Two separate copies of that rule had drifted, and `Data::Translators` trips the first:

```raku
multi sub html-table-highlight(Str:D $s, :h(:@highlight)!,
                               Str:D :c(:$color) = 'Orange',
                               :s(:$font-size) = Whatever, …)
```

The duplicate-**variable** check counted the alias's outer name (`s`) as a declared `$s`, so the ordinary positional `$s` in the same signature was rejected as `X::Redeclaration: Redeclaration of symbol '$s'`. The variable an alias really binds is its inner sub-signature parameter, which the same function already collected separately — the outer name only ever names an external argument key, whose uniqueness is the `X::Signature::NameClash` check that follows. That is also the error rakudo reports for a genuinely repeated key, which mutsu was reporting as a Redeclaration for every spelling.

The second copy is the nested-alias key collector, which stripped a leading `:` and never a sigil. So `:h(:@highlight)` answered to `h` but not to `highlight`, and the call died with "Unexpected named argument"; `:h(:$hi)`, whose inner name carries no sigil, worked, which is what hid it. Both sites share one `named_param_external_key` now.

## Follow-ups filed rather than forced through

- **#8465** (`todo:ticket`) — a `Slip` read out of a container does not slip (`my $x = slip(5,6); (1, $x, 2).elems` is `3` where rakudo gives `4`), and a scalar-assigned one is not itemized. Two hot-path halves (`exec_make_array_op`'s `WrapVarRef` capture, and `itemize_value` having no `Slip` arm) that deserve their own change and their own full-suite run.
- **#8466** (`todo:ticket`) — `OUR::«'&infix:<@~~>'» := …`, an interpolating word quote as a package-symbol bind target. `Data::Record::Lifter` is where `infix:<@~~>` comes from, so it is the single construct still blocking four of `Data::Record`'s twelve modules.

## Tests

Five new pins plus two grown files, all verified green **under rakudo itself** except where noted:

- `t/lang/sigil-contextualizer-stacked-slip.t`
- `t/routines/signature/param-type-then-type-capture.t`
- `t/oo/attribute/let-temp-attribute-and-term-forms.t`
- `t/routines/signature/anon-onearg-param-marker.t`
- `t/routines/signature/named-arg-alias-external-key.t`
- `t/routines/dispatch/multidim-splat-lazy.t` grows the parenthesized-lvalue case of the dimension splat — 6.e behaviour, which mutsu implements unconditionally, so it is verified against rakudo under `use v6.e.PREVIEW` rather than as written.

`cargo fmt`, `make lint`, `make test` (PASS, 4339 files / 46354 tests) and `make roast` all run locally. `make roast`'s only failures are the four container-only files documented in [docs/agent-environments.md](https://github.com/tokuhirom/mutsu/blob/main/docs/agent-environments.md) — `6.c/S32-io/file-tests.t` (Failed: 4, tests 6-8/10), `S16-filehandles/filetest.t` (Failed: 25), `S16-io/eof.t` (exit 255, "planned 5 ran 1") and `S32-io/IO-Socket-Async.t` (exit 124, "planned 40 ran 17") — matching those entries' recorded shapes exactly; they are the `uid 0` and sandboxed-network cases, not this change.

Write-up: `news/2026-09/parse-failure-index-stacked-slips-type-captures-and-let-temp-terms.md`.

Refs #7954

https://claude.ai/code/session_01BhVQPxo8Dt28QLMxLX2V6H

---
_Generated by [Claude Code](https://claude.ai/code/session_01BhVQPxo8Dt28QLMxLX2V6H)_


tokuhirom:
## The bundled-battery gate caught a regression (fifth commit)

CI was red on `test-suites`: `REGRESSION: whitelisted 'Template::Mustache 06-logging.rakutest' did not pass this run`. `render` died with `No such method 'log'`, because `temp $!logger.level = $_ with $log-level` destroyed `$!logger` and the next line delegates through it (`has $.logger handles <log>`). That gate is a CI step of its own, **not part of `make test`**, which is why the pre-publication run did not see it; it is in my local gate list now.

Making the restore write `self`'s attribute cell — which the `temp $!x` fix in this PR requires — unmasked two things the old, ineffective restore had been hiding.

**The save was reading a different store from the restore.** `temp $!o.attr = v` emits a `LetSave` naming `!o`, and `exec_let_save_op` read the method frame's mirror slot and then `env`. In a body that has not otherwise touched `$!o` both are empty, so the snapshot was `Nil`, and the restore duly wrote that `Nil` over a live object. The save reads the cell now, so the two agree.

**A statement modifier is not a block, so it opens no `let`/`temp` scope.** The save belongs to the enclosing routine and must still stand on the next statement:

```raku
my $x = 1;
sub f($c) { temp $x = 2 if $c; say $x }   # raku: 2   mutsu: 1
```

No attribute involved, and **it reproduces on `main`** — verified by building `main`'s `src/` and running the same file. The modifier's branch was given its own `OpCode::LetBlock` frame and restored there, which was invisible for as long as the restore wrote only the mirror slot and `env` (nothing reads those back for an attribute) and became a destroyed attribute the moment the cell was in play. Both the statement-position and the value-position branch emitters skip the frame for a modifier now, beside the topic rebind, block-local scope and dynamic scope those same functions already excluded it from.

Worth carrying forward: **a restore that starts writing a *shared* store is not a local change.** Every save feeding it has to be re-checked for reading the same store, and every scope that brackets it has to be re-checked for being a real scope — neither is observable while the restore is writing somewhere nothing reads.

`t/oo/attribute/let-temp-attribute-and-term-forms.t` grows four cases for this, green under rakudo itself: a `temp` reaching through an attribute leaves the attribute defined and the same object (so delegation through it still resolves), and `if`/`with` modifiers hold their save to the end of the routine.

Re-run after the fix: `cargo fmt`, `make lint`, `make test` (PASS, 4339 files / 46361 tests), the battery gate (`GATE PASSED`, 308/326 — the regressed file back) and `make roast`, whose only failures are the four container-only files already named in the PR body, matching their recorded shapes.

The sixth commit is the `news/` write-up of the above (documentation only).

---

Issue #8578, opened 2026-09-16
Dynamic `does` cannot resolve a `my role Foo::Bar` declared inside the same module by its qualified name


tokuhirom:
Found while investigating #8560 (imported `trait_mod:<is>` on parameters) —
this is what still blocks `App::Prove6` (via Getopt::Long) from fully
loading after #8560's fix lands, even though the "unknown trait" symptom
that issue tracked is now gone.

Getopt::Long declares a lexically-scoped, dot-qualified role and mixes it in
dynamically from a sibling sub in the **same file**:

```raku
unit module Getopt::Long;

my role Formatted::Named {
    has $.argument;
}

sub some-trait-handler($param) {
    return $param does Formatted::Named(:argument(42));
}
```

Minimal reduction (mirrors the shape exactly, package name deliberately
different from the role's own qualifier segment, matching Getopt::Long
where the module is `Getopt::Long` and the role is `Formatted::Named`, not
`Getopt::Long::Formatted::Named`):

```raku
unit module Foo;

my role Formatted::Named is export {
    has $.x;
}

sub apply($obj) is export {
    return $obj does Formatted::Named(:x(1));
}
```

```raku
use Foo;
say apply(42).x;
```

Expected (rakudo): `1` (and `.^name` gists as `Int+{Foo::Formatted::Named}`).

mutsu:
```
Could not find symbol '&Named' in 'GLOBAL::Formatted'
```

mutsu appears to parse the dotted role name `Formatted::Named` in the
dynamic `does` expression as a call to `&Named` inside a `Formatted`
package/namespace rather than resolving it as the single lexically-scoped
role name `my role Formatted::Named` declared earlier in the same file —
i.e. it treats the `::` as a namespace-path separator for lookup instead of
matching the role's own declared (compound) short name in the current
lexical scope.

This is a real-world blocker: it's the second of two bugs (the first, #8577,
is `does` losing a named-arg-parametrized attribute value) standing between
mutsu and loading `App::Prove6`/`Getopt::Long` end to end. Likely needs
design work in how lexically-scoped compound-named roles are indexed and
looked up for a dynamic `does`, hence `todo:deep` rather than `todo:ticket`.

tokuhirom:
Claiming: claude/loving-ptolemy-4c56vx

---
_Generated by [Claude Code](https://claude.ai/code)_

tokuhirom:
Releasing: claude/loving-ptolemy-4c56vx

[...]

Issue #8577, opened 2026-09-16
`$obj does Role(:named-arg(...))` loses the role's named-parametrized attribute value


tokuhirom:
Found while investigating #8560 (imported `trait_mod:<is>` on parameters).

A dynamic `does` that parametrizes the role's attribute via a **named**
argument silently loses the value — the attribute reads back as
undefined/empty instead of the value passed.

Minimal reduction (`tmp/bug1-does-role-named-arg.raku` in that investigation,
not committed):

```raku
role Tagged {
    has $.spec;
}

class Marker { has $.spec; }

my $obj = Marker.new;
$obj does Tagged(:spec("hello"));
say $obj.spec;
```

Expected (rakudo): `hello`

mutsu: prints an empty string (the attribute is undefined).

This reproduces with a bare `class`/`role`, no modules, parameters, or
traits involved — it is a general `does`-with-role-mixin-constructor-args
bug, not specific to the parameter-trait machinery #8560 was about.
Getopt::Long's `multi trait_mod:<is>(Parameter $p, Argument :option($arg)!) { $p does Formatted::Named(:$argument) }` is one real-world consumer that needs this fixed, but any code using the same pattern is affected.

Positional role parametrization (`$obj does Tagged("hello")` if `Tagged` took
a positional attribute) was not checked here and may or may not share the
same bug — worth checking as part of the fix.

Root cause not yet investigated; likely somewhere in the runtime `does`/mixin
construction path where a role's `.new`-style named-argument binding for its
declared attributes is evaluated for a dynamic (non-declaration-time) `does`.

tokuhirom:
Claiming: codex/mutsu-issue-8577-role-named-arg-20260917

tokuhirom:
Releasing: codex/mutsu-issue-8577-role-named-arg-20260917

PR #8630 merged and verified on origin/main.

PR #8581, opened 2026-09-16
fix: custom parameter trait carries its argument to trait_mod:&lt;is&gt; dispatch

### LINKED ISSUE #8560: Support imported trait_mod:<is> traits on routine parameters
App::Prove6 0.0.18 loads `Getopt::Long`, which exports `trait_mod:<is>` candidates for `is option` and `is getopt`. mutsu then rejects the importing module’s `MAIN` signature with:

```text
Can't use unknown trait 'is' -> 'option' in a parameter declaration.
```

Reduction (after `use Getopt::Long`):

```raku
sub MAIN(Bool :$flag is option<!>) { }
```

Rakudo compiles this. mutsu currently has custom trait machinery for declarations, but parameter and imported-trait dispatch requires cross-layer design/work; this blocks loading App::Prove6 after the separate regex-capture indexing parser bug is fixed.

tokuhirom:
## Summary

- `check_param_custom_traits` (the VM-side dispatcher for a *custom* parameter trait like `is query` / `is option`) always called the matching `trait_mod:<is>` candidate with a hardcoded `True`, discarding whatever argument the trait actually carried.
- App::Prove6's real signature, `sub MAIN(Bool :$timer is option<!>) { }` (via Getopt::Long), hits this directly: Getopt::Long declares `multi trait_mod:<is>(Parameter $p, Argument :option($arg)!)` and a `Str:D :getopt(:$option)!` sibling — neither accepts a bare `Bool`, so dispatch always failed and mutsu reported the generic `Can't use unknown trait 'is' -> 'option' in a parameter declaration.` even though a real candidate matched the actual argument value (`"!"`, parsed by Getopt::Long's own grammar as a negatable boolean spec).
- Fix: the parser now *parses* a trait's optional argument — the `<!>` word-quote form and the `('utf8')` parenthesized form — into a real `Expr` instead of merely skipping past it. `ParamDef` gained a sparse `trait_args: Vec<(String, Expr)>` field (most traits carry no argument and still dispatch with `True`, unchanged), threaded through every parameter-parsing site: ordinary signatures, sub-signatures, pointy blocks, and `for`/`with`/`while` loop parameters. `check_param_custom_traits` evaluates the matching entry and passes the real value.
- Also fixed a related bug in the same function: a real error raised from inside a *matched* candidate's own body was being swallowed and reported as the generic "unknown trait" message, indistinguishable from a typo'd trait name. It now discriminates "no candidate accepted this trait" from "a candidate matched but its body raised a real error", mirroring the same discrimination the sub-level custom-trait dispatch already does.

Closes #8560.

App::Prove6 still doesn't load end-to-end after this fix: once dispatch actually reaches Getopt::Long's own trait bodies, two separate, pre-existing bugs surface. Filed separately since they are unrelated to the parameter-trait-argument mechanism this PR fixes:
- #8577 — `$obj does Role(:named-arg(...))` loses the role's named-parametrized attribute value.
- #8578 — a dynamic `does` cannot resolve a lexically-scoped, dot-qualified role (`my role Formatted::Named`) declared earlier in the same file.

## Test plan

- Added `t/oo/trait/custom-parameter-trait-argument-dispatch.t` (+ `t/lib/ArgTraitFixture.rakumod` fixture) mirroring Getopt::Long's own two-candidate-by-argument-type shape without depending on Getopt::Long itself: a word-quote argument (`is tag<hello>`) and a parenthesized argument (`is tag(Marker.new)`) each reach dispatch as the real value and select the correctly-typed candidate; a real error from a matched candidate's body propagates instead of being masked as "unknown trait". Verified this test passes under `raku` too.
- `cargo fmt --all` and `make lint` — clean.
- `make test` — `Files=4399, Tests=46871`, PASS.
- `make roast` — the four failures are exactly the documented container-only exceptions in `docs/agent-environments.md` (root bypassing `chmod` permission checks, sandboxed network breaking `IO-Socket-Async.t`); no other failures.

https://claude.ai/code/session_01Fx6BEuqCFeP2Gy2wkL6swo

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fx6BEuqCFeP2Gy2wkL6swo)_


tokuhirom:
`lint-configs` and `test-check` both failed on the first push with the same root cause: `cargo clippy -- -D warnings` flagged `explicit_auto_deref` on two `Value::str_from(*one)` / `Value::str_from(*w)` calls in the new word-quote-argument parsing code (`src/parser/stmt/sub/param_validate.rs`) — `str_from` already takes anything deref'ing to `&str`, so the explicit `*` was unnecessary. Fixed in 229c4bf5 and confirmed locally with a full `make lint` (all 4 configurations green) before pushing.


---

Issue #8645, opened 2026-09-17
`$*OUT`/dynamic-var sigil alias ($*NAME vs *NAME) desyncs on two write paths, breaking output capture


tokuhirom:
## Summary

mutsu stores a `$*`-twigil dynamic variable under two env keys: the sigilless
form (`*OUT`) and the sigilled form (`$*OUT`), kept in sync via
`Interpreter::twigil_dynamic_alias` inside the single canonical writer
`set_env_with_main_alias` (`src/vm/vm_env_helpers.rs`). `say`/`print`/`note`
read the destination handle through `write_to_named_handle("$*OUT", ...)` →
`get_dynamic_handle`, which looks up the **sigilled** key first
(`src/runtime/io_env.rs`). A plain variable read of `$*OUT` (e.g. `.^name`,
or reading it as an ordinary term) instead resolves through the compiled
**sigilless** `*OUT` local/env slot.

Two independent write paths update only the sigilless key, leaving the
sigilled alias stale — so `say`/`note` keep writing to the ORIGINAL handle
(or silently vanish) even though every other read of `$*OUT` correctly
reports the new value. Found while making the `Trap` zef distribution's test
suite (`t/01-basic.rakutest`, `t/02-tee.rakutest`) pass under mutsu — its
whole `Trap(my $*OUT)` capture idiom depends on this working.

## Case A — `is raw` parameter write-back through a CALL-ME dispatch

```raku
class Foo {
    has str @!text;
    method print(*@_ --> True) { if @_.join -> $t { @!text.push: $t } }
    method say(*@_ --> True)   { if @_.join -> $t { @!text.push: $t ~ "\n" } }
    method text(--> str) { @!text.join }
    multi method CALL-ME(Foo:U: $one is raw) { $one = self.new; }
}

my $out;
{
    $out = Foo(my $*OUT);
    note "DURING block, \$*OUT.^name = " ~ $*OUT.^name;   # correctly "Foo"
    say "Hello world!";                                    # should be captured
}
say "captured=[" ~ $out.text ~ "]";
```

Expected (rakudo): `DURING block, $*OUT.^name = Foo` then `captured=[Hello world!\n]`.

mutsu: `$*OUT.^name` correctly reports `Foo` (the sigilless key was updated by
the `is raw` write-back), but `say "Hello world!"` still writes to the real
stdout and `$out.text` is empty — `write_to_named_handle("$*OUT", ...)` is
[...]

PR #8647, opened 2026-09-17
fix(runtime): CALL-ME dispatch no longer preempted by coercion-type-term shortcut

tokuhirom:
## Summary

- `Foo($x)` on a type object declaring `CALL-ME` must dispatch to it, never to the "calling a type with a type-object argument constructs a coercion type object" shortcut (`Str(Any)`, `Int(Str)`, `Child(Parent)`) in `call_function_fallback`. That shortcut fired whenever the single argument's runtime value was a type object or undefined (`Nil`), regardless of whether the target class declared `CALL-ME` — so a class like the `Trap` zef distribution's `method CALL-ME(Trap:U: $one is raw) { $one = self.new(:$tee) }`, called with a not-yet-defined argument (`Trap(my $*OUT)`), never ran `CALL-ME` at all: it silently got back the symbolic type object `Trap(Any)`, which then died on first use.
- Gate the shortcut on `!class_has_method(name, "CALL-ME")` / `!role_has_method(name, "CALL-ME")`, resolving a lexically-scoped (`my class`) name's mangled storage identity first so the check also works for a `my class Foo { method CALL-ME { ... } }` declared in the current lexical scope.
- Re-measured the `Trap` ecosystem record: `t/01-basic.rakutest` no longer dies (was: die on "No such method 'text'"); both baseline files now run to completion, `mutsu_assertions` up from 1 to 5 of 13. Status stays `red` (0/2 parity) because a second, deeper bug in `$*OUT`'s sigilled/sigilless env-key mirroring — filed as [tokuhirom/mutsu#8645](https://github.com/tokuhirom/mutsu/issues/8645) — still keeps both files short of full pass. Locked on the ecosystem lock board, [tokuhirom/mutsu#7884](https://github.com/tokuhirom/mutsu/issues/7884).

## Test plan

- [x] New regression test: `t/oo/class/call-me-undefined-arg-wins-over-coercion.t`
- [x] `cargo fmt --all`
- [x] `make lint`
- [x] `make test`
- [x] `make roast`
- [x] Re-measured `ecosystem/dists/T/Trap~e5281781.json` with the release binary (`--only Trap`)

https://claude.ai/code/session_01Q59fHqpewkwCrEGy555C5x

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q59fHqpewkwCrEGy555C5x)_


tokuhirom:
CI red on `test-check`/`gc-stress-roast`/`jit-stress-roast`: my `!class_has_method(name, "CALL-ME")` guard was too broad — it also suppressed the coercion-type-term construction for `A(Any)`/`A()` (whitelisted `roast/S13-overloading/typecasting-long.t`'s "A() is a type coercion literal" / "A(Any) is a type coercion literal"), which must stay a coercion term *even when the type declares CALL-ME*, per rakudo. Root cause: `Any` the bareword compiles to a literal `ValueView::Package("Any")`, while an undefined variable (`Trap(my $*OUT)`'s `$*OUT` before assignment) evaluates to the distinct `ValueView::Nil` — my guard didn't need to touch the `Package`/`ParametricRole` arms at all, only `Nil`. Narrowing the fix now and re-running the suites before pushing.

---

tokuhirom:
Fixed and pushed (2626ad7f): narrowed the guard to only let `CALL-ME` win when the sole argument traces back to a caller-visible variable (via `pending_call_arg_sources`, the same signal `substr-rw`'s Proxy support already reads), leaving a literal type reference (`A(Any)`, `Int(Str)`) building the coercion term unconditionally as before. `roast/S13-overloading/typecasting-long.t` now passes 39/39 locally, alongside the original `Trap` fix, `t/oo/class/class-type-object-coercion-call.t`, `t/types/coercion/generics-coercion-terms.t`, the new regression test, and a full local `make lint` / `make test` / `make roast`. Re-measured `Trap` in a3a89dcc83 -- unchanged (5/13 assertions), confirming this refinement didn't move its own record.

---

Issue #8380, opened 2026-09-14
deep: default worker pool deadlocks virtual scheduler continuations


tokuhirom:
## Finding

The `Test::Time` 0.0.2 ecosystem suite exposes a deadlock in mutsu’s default worker-pool path when a user scheduler resumes a `start` continuation after multiple virtual-time advances. The distribution is otherwise pure and its Rakudo baseline passes.

## Reduction

Using the extracted `Test::Time` and `Test::Scheduler` libraries:

```raku
use Test::Time;

my $scheduler = mock-time;
my $init = Promise.new;
my $p = start {
    my $*SCHEDULER = $scheduler;
    $init.keep;
    sleep 10;
    say "after first sleep";
    sleep 20;
    say "after second sleep";
}
await $init;
$scheduler.advance-by: 10;
$scheduler.advance-by: 20;
$scheduler.advance-by: 20;
await $p;
say "done";
```

Rakudo completes:

```text
after first sleep
after second sleep
done
```

mutsu’s default worker pool times out after 15 seconds without completing. `MUTSU_POOL=off` completes with the same three lines. The full subject file `t/01-tdd.t` reaches the scheduler subtest but does not finish under the default pool.

## Why this is deep

The failure is in cross-cutting worker-pool/cooperative-scheduler continuation ordering, rather than a bounded parser or builtin slice. Fixing it needs an execution-layer concurrency invariant and should not be papered over in `Test::Time` or by making the distribution use a different pool mode.

Source distribution: `Test::Time` 0.0.2, `t/01-tdd.t`. This issue is filed in `tokuhirom/mutsu`; no distribution files are modified.

tokuhirom:
Claiming: claude/todo-ticket-pr-8f98ok

Investigating the worker-pool / virtual-scheduler continuation deadlock.

---
_Generated by [Claude Code](https://claude.ai/code)_

tokuhirom:
## Root cause, confirmed

Reproduced with the real upstream `Test::Time` 0.0.2 / `Test::Scheduler` 1.2 distributions
(fetched from the zef mirror at the URLs in the ecosystem records), not just a paraphrase.

**`Test::Scheduler.advance-by` relies on `.keep()`ing an awaited `Promise` deterministically
driving the awaiter's continuation forward — through its *next* blocking point — before the
keeper's own subsequent step runs.** Real Rakudo's continuation-based `await` (a `start`-block
[...]

PR #8671, opened 2026-09-17
test: pin ADR-0105 S0 oracle for promise-scheduler cue dispatch

tokuhirom:
## Summary

Implements S0 from [ADR-0105](https://github.com/tokuhirom/mutsu/blob/main/docs/adr/0105-promise-resolution-dispatches-through-the-promise-scheduler.md) §8 ("pin the oracle"), the first slice of #8380's implementation plan.

Adds `t/concurrency/promise/scheduler-cue-dispatch-oracle.t`: a logging `Scheduler` class records every `.cue` dispatch, and the test pins the Appendix A measurements from the ADR against mutsu's current behavior:

- **F4a** — `Promise.scheduler` should expose the scheduler a promise was constructed with. It doesn't exist yet (`No method 'scheduler' on Promise`).
- **F1a** — keeping an awaited promise should cue the awaiter's wake-up through the promise's bound scheduler. mutsu's `await` is a blocking condvar wait that never consults a scheduler.
- **F1b** — a `.then` callback should be dispatched through the promise's scheduler too. Same gap.
- **F4b** — a `start` block under a user `$*SCHEDULER` should cue its task through it. mutsu always dispatches to the built-in worker pool directly.
- **Experiment 3** — the keeper's own continuation (setting a flag before its next blocking point) should be observed by the woken awaiter before dispatch can overtake it. Rakudo is 30/30 on this; the ADR measured mutsu at 28/30, so this assertion may occasionally pass today by chance (harmless under `todo`).

Each of the failing assertions is marked `todo` with the specific ADR-0105 slice (S1/D1, S2/D2, S4/D4) expected to make it pass, so the test both documents the current gap and becomes the acceptance check as S1–S4 land. No `src/` change — this is a test-only pinning slice per the ADR's S0 definition.

## Test plan

- [x] `prove -e target/debug/mutsu t/concurrency/promise/scheduler-cue-dispatch-oracle.t` — passes (0 unexpected failures; TODO tests behave as expected, verified against raw TAP output)
- [x] `cargo fmt --all` — no changes (test-only diff)
- [x] `make lint` — clean
- [x] `make test` — clean
- [x] `make roast` — clean

Refs #8380 (not closing it — S1–S5 remain).

https://claude.ai/code/session_01Wea3c9iKBLskX5bvPmSsCW

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wea3c9iKBLskX5bvPmSsCW)_


tokuhirom:
`test-check` failed on the first push: `make check-t-layout` (`scripts/migrate-t-layout.py --check`) placed the new test at `t/concurrency/thread-lock/scheduler-cue-dispatch-oracle.t`, not the `t/concurrency/promise/` path I originally used — the existing Scheduler-focused tests (`does-scheduler-role.t`, `scheduler-cue-times.t`, `scheduler-loads-cue.t`) already live under `thread-lock/`, and the layout tool's rules classify this file the same way.

Fixed in `a43dfe46` by moving the file (`git mv`, no content change); `make check-t-layout` and the test itself both pass locally at the new path. Re-watching CI.

---

Issue #8701, opened 2026-09-18
The safepoint poll network serves one hard-coded consumer, so a profiler cannot ride it


tokuhirom:
ADR-0106 Slice 1 (`docs/adr/0106-raku-level-profiler.md` §5). **Slice 0 (#8699) has landed**, so this is unblocked; the sampler that consumes it is #8702.

## What is wrong

mutsu already has a poll network that reaches every place a mutator can be interrupted, and it can only ever do one thing:

- `src/gc/safepoint.rs:53` — `enum SafepointKind` with ten kinds (`Backedge`, `Call`, `Return`, `Await`, `ReactPoll`, `LazyForce`, `NestedRun`, `ThreadJoin`, `Manual`, construction), already sited across all thirteen bytecode dispatch loops (`vm_run_loop.rs:146/467/812/883`, `vm_call_light.rs:579`, `vm_call_light_typed.rs:543`, `vm_call_fast.rs:220`, `vm_call_named_inner.rs:367`, `vm_method_dispatch.rs:876/2108`, `vm_closure_dispatch.rs:1072`, `vm_helpers_lazy.rs:1337`, `vm_helpers_lazy_pull.rs:189`) and the async machinery
- `src/gc/safepoint.rs:293` — `armed()`, a single cached bool (`triggers().armed`), re-exported as `gc_safepoints_armed` (`src/gc/mod.rs:42`). With `MUTSU_GC=off` a poll costs one load
- `src/vm/vm_jit_compile.rs:484+` — the JIT emits a call to the poll on **native backedges**, and `src/vm/vm_jit.rs:360` polls on entry, so JIT-compiled code is covered too (ADR-0004 §2.4)

The gate and the callee are both spelled "GC". A second consumer (the profiler) cannot arm the network without turning GC on, and cannot learn *where* the mutator is, because the JIT's shim (`src/vm/vm_jit_helpers.rs:65`, `safepoint(_interp: *mut Interpreter)`) takes no instruction pointer.

## What to do

1. Split the gate: `vm_poll::armed()` = `gc_armed | profiler_armed`, as **one** cached bool whose union is computed once at arm time. Not two bools OR'd at each poll — that is a load per opcode on a path that runs on every backedge.
2. Turn `gc_safepoint(kind)` into `vm_poll(kind, site)` with GC as its first consumer. Existing call sites keep their `SafepointKind`; no behavioural change when only GC is armed.
[...]

PR #8712, opened 2026-09-18
profiler: emit the ip-carrying backedge shim only when the profiler is armed

tokuhirom:
Follow-up to #8710 (ADR-0106 Slice 1), which landed the `vm_poll` network but not step 3 of #8701: the JIT emitted the ip-passing form of the backedge poll **unconditionally**. That is the one cost in this slice that is not free by construction — the ADR names it explicitly — and gate 1c exists to catch exactly this. It also completes the slice's acceptance test, which previously called the helper directly rather than observing a poll raised by compiled native code.

## What changed

- **`PollShim::current()`** decides once per process which shim the Cranelift emitter calls:
  `safepoint(interp)` — byte-for-byte the call a native backedge made before Slice 1 — when no
  profiler is armed, and `safepoint_at(interp, site)` when one is. Sound because arming is a
  process-lifetime decision read at the first poll, while a chunk is compiled later at the hotness
  threshold, so the arming state is already fixed when code is generated.
- Non-dispatch poll sites pass a **`NO_SITE` sentinel** instead of `0`, so "no ip" is distinguishable
  from "the chunk's first instruction".
- `MUTSU_PROFILE=1` prints one `profiler-poll:` line at exit (the shape `MUTSU_VM_STATS`
  established), standing in for Slice 2's sampler so the network is observable end to end.
- `note_native_poll` is gated on the `jit` feature — it is reachable only from the emitted shim, and
  a `--no-default-features --features native` build would otherwise fail `lint-configs` on dead code.

## The acceptance test, and what building it turned up

**Raku's own loops never put a backward jump inside a JIT-compiled chunk.** Every loop form compiles
to a compound opcode whose body is a separate compiled range with no backedge of its own, so
`while` / `for` / `loop` / `repeat` all reach the poll through the interpreter, not the emitted shim.
The shape that does reach it is `nqp::while`, which emits a plain backward `Jump` into the enclosing
chunk. `tests/jit_diff.rs` uses that fixture: 4000 native polls (200 calls x 20 iterations) carrying
the ip of the loop's backward jump, against **0** for the same program under `MUTSU_JIT=off` — the
contrast is what pins those polls to native code rather than the interpreter loop. Three unit tests
in `src/vm/vm_poll.rs` pin the two shims' ABI and the selection itself.

## Validation

`perf` is not available in this container, so gates 1 and 1c were measured with **callgrind**, whose
`Ir` count is deterministic and load-independent — it answers the same question. Baseline is the
release binary of `d559d288` (main immediately before Slice 1 landed); `MUTSU_GC=off` throughout.
`nqp-backedge` is the fixture above scaled to 4,000,000 native backedges, i.e. the workload the
specialization exists for.

| workload | JIT | baseline Ir | this branch | delta |
|---|---|---:|---:|---:|
| bench-fib | off | 2,703,889,033 | 2,705,162,835 | +0.047% |
| bench-tak | off | 2,603,936,806 | 2,604,720,102 | +0.030% |
| bench-mandelbrot | off | 1,106,631,576 | 1,106,551,354 | -0.007% |
| nqp-backedge | off | 14,642,371,483 | 14,642,375,080 | +0.000% |
| bench-fib | on | 1,498,712,992 | 1,499,983,919 | +0.085% |
| bench-tak | on | 1,743,997,234 | 1,744,776,039 | +0.045% |
| bench-mandelbrot | on | 663,204,510 | 663,419,849 | +0.032% |
| nqp-backedge | on | 11,016,642,286 | 11,024,251,001 | +0.069% |

All within the ADR's 0.5% gate, including gate 1c's JIT-on rows.

- `make lint` — green (all four configurations; the first run caught the dead-code failure above)
- `make test` — PASS, `Files=4475, Tests=47408`
- `make roast` — red only on `roast/6.c/S32-io/file-tests.t` (Failed: 4) and
  `roast/S16-filehandles/filetest.t` (Failed: 25), the documented `uid 0` container-only pair in
  `docs/agent-environments.md`, with the counts and test numbers that file records. `id -u` is 0 here.

Refs #8701.

https://claude.ai/code/session_016Saq7LQR3tCZNhpQS4CN8H

---
_Generated by [Claude Code](https://claude.ai/code/session_016Saq7LQR3tCZNhpQS4CN8H)_


tokuhirom:
Closing: overtaken by #8711, which landed while this was in the local suites. Slice 3 added `vm_poll::profiler_armed()` and the same codegen specialization this PR was written for — `poll_with_code` selects `profile_safepoint` when armed and keeps the one-argument `safepoint` otherwise — so landing this would be a second, conflicting implementation of a gap that is now closed. Nothing here is lost by closing; what is worth keeping is recorded below and carried into a follow-up.

**The gates are measured.** `perf` is absent in this container, so gates 1 and 1c were measured with callgrind, whose `Ir` count is deterministic and load-independent. Baseline is the release binary of `d559d288` (main immediately before Slice 1), `MUTSU_GC=off` throughout; `nqp-backedge` is a fixture scaled to 4,000,000 native backedges, i.e. the workload the specialization exists for:

| workload | JIT | baseline Ir | Slice 1 | delta |
|---|---|---:|---:|---:|
| bench-fib | off | 2,703,889,033 | 2,705,162,835 | +0.047% |
| bench-tak | off | 2,603,936,806 | 2,604,720,102 | +0.030% |
| bench-mandelbrot | off | 1,106,631,576 | 1,106,551,354 | -0.007% |
| nqp-backedge | off | 14,642,371,483 | 14,642,375,080 | +0.000% |
| bench-fib | on | 1,498,712,992 | 1,499,983,919 | +0.085% |
| bench-tak | on | 1,743,997,234 | 1,744,776,039 | +0.045% |
| bench-mandelbrot | on | 663,204,510 | 663,419,849 | +0.032% |
| nqp-backedge | on | 11,016,642,286 | 11,024,251,001 | +0.069% |

All inside the ADR's 0.5%, gate 1c's JIT-on rows included.

**A finding that outlives this PR:** no Raku loop form puts a backward jump inside a JIT-compiled range. `while`, `for`, `loop`, C-style `loop` and `repeat` all compile to compound opcodes whose body is a *separate* compiled range, so the range has no backedge of its own — measured as 0 native polls on every one of them. The only shape that reaches the emitted backedge hook is `nqp::while`, which emits a plain backward `Jump` into the enclosing chunk. That has consequences for §8 gate 4 and for the exactness Slice 3 claims; I am filing it separately rather than leaving it in a closed PR.

Local verification of this branch, for the record: `make lint` green in all four configurations, `make test` PASS (`Files=4475, Tests=47408`), `make roast` red only on the documented `uid 0` container-only pair.

---

Issue #8646, opened 2026-09-17
A role's own imported custom infix operator becomes unreachable from its methods (two known trigger shapes)


tokuhirom:
## Summary

A parameterized role that imports a custom `infix:<...>` operator via its own
body's `use` statement, and then calls that operator from one of its own
methods, can get `Two terms in a row` at runtime — the operator sub simply
cannot be found — even though [#8xxx-fixed] (this PR) already fixed the
"wrong target package" bug that used to cause the same symptom
unconditionally.

This is filed as the residual: after that fix, the operator call still fails
under two circumstances that I could not fully unify with the time I had.
Both reproduce the *exact* runtime error `Algorithm::Kruskal` hits (an
ecosystem distribution: `Algorithm::MinMaxHeap`'s private `!bubble-up` method
calls `Algorithm::MinMaxHeap::CmpOperator`'s custom `minmaxheap-cmp` infix),
so `Algorithm::Kruskal`'s `t/01-basic.t` still dies partway through even with
the fix applied (see `ecosystem/dists/A/Algorithm--Kruskal~9e64a773.json`).

## Shape 1 — role name has no `::` in it (a flat, non-namespaced role)

```raku
# lib/FlatComparable.pm6
use v6;
unit role FlatComparable[::T];
method compare-to(T) returns Order:D { ... };
```
```raku
# lib/FlatCmpOperator.pm6
use v6;
unit module FlatCmpOperator;
use FlatComparable;

multi sub infix:<flat-cmp>(Mu $lhs, Mu $rhs) is export {
    return $lhs cmp $rhs;
}
multi sub infix:<flat-cmp>(FlatComparable $lhs, FlatComparable $rhs) returns Order:D is export {
    return $lhs.compare-to($rhs);
}
```
```raku
# lib/FlatHolder.pm6
use v6;
unit role FlatHolder[::Type];
use FlatComparable;
use FlatCmpOperator;

has Type @.nodes;
method insert($value) { @!nodes.push($value); }
method !do-cmp($i, $j) { return @!nodes[$i] flat-cmp @!nodes[$j]; }
method cmp-first-two() { return self!do-cmp(0, 1); }
```
```raku
# lib/FlatMaker.pm6
use v6;
unit module FlatMaker;
use FlatComparable;
use FlatHolder;

sub make-holder() is export { return FlatHolder[FlatComparable].new; }
sub make-thing() is export {
    my class State {
        also does FlatComparable[State];
[...]

PR #8889, opened 2026-09-20
fix: a class-body use import is now callable from that class's own methods

### LINKED ISSUE #8883: A sub imported via `use` inside a class body is not callable from that class's own method bodies
## Summary

`use Some::Module;` written directly inside a class body (a `unit class Foo; use Some::Module; ...` compilation-unit-is-the-class-body form, though see below — it is not specific to `unit class`) imports `Some::Module`'s exported subs into the class's own package/lexical scope, as expected — but a **method** defined in that same class cannot call the imported sub: mutsu throws `Unknown function: <name>` (or, in some invocation shapes, silently mis-parses the call into something else — see below). Raku resolves the call correctly from inside the method body.

## Repro

`MyTemplate.rakumod`:
```raku
unit class MyTemplate;
use Template::Classic;   # exports `sub template(...)`, `sub escape(...)`

method run($x) {
    my &f := template Signature.new(:params(())), $x;
    say "f.^name = ", &f.^name;
    return f();
}
```

Main file:
```raku
use lib '.';
use MyTemplate;
say MyTemplate.new.run("hello");
```

rakudo: runs `template(...)` normally inside `run`, prints `f.^name = Sub` then the rendered template output.

mutsu (current `main`):
```
Unknown function: template
  in sub run at MyTemplate.rakumod line 5
  in block <unit> at ...
```

`Template::Classic` is a real, small, vendored zef module (`Template-Classic-0.0.3`) — nothing exotic about its `is export` declaration.

## A second, more confusing shape of the same bug

In the real distribution this was found in (`Karabiner::CompModGenerator`'s `Karabiner::Template` class, `method rule_generator($tmpl) { ... my &generate-rule := template Signature.new(:@params, :returns(Seq)), %?RESOURCES{$tmpl}.slurp; ... generate-rule(|@values) ... }`), the *exact same* root cause does not surface as "Unknown function" — instead `&generate-rule` silently ends up bound to a **`List`** (containing `(Signature.new(...), %?RESOURCES{$tmpl}.slurp)`'s two elements, evidently) rather than to `template`'s return value, and the later `generate-rule(|@values)` call dies with `No such method 'CALL-ME' for invocant of type 'List'`. This looks like a different failure to a first read, but is the same underlying cause (the minimal repro above reproduces the plainer "Unknown function" shape identically whether the call site spans one line or several, with a computed or literal `Signature`) — worth keeping in mind when someone investigates this, since the two symptoms could otherwise look unrelated.

## Where this was found

`Karabiner::CompModGenerator` 0.0.20 (zef), locked/worked via the ecosystem distribution roulette ([#7884](https://github.com/tokuhirom/mutsu/issues/7884)): `lib/Karabiner/Template.rakumod` is `unit class Template; use Template::Classic; method rule_generator($tmpl) { ...calls template(...)... }`. This blocks `t/03-SafariTabs_tmpl.rakutest`, the distribution's only baseline (rakudo-passing) file.

## Likely area

Class-body `use` / import registration (`src/runtime/registration_class_body.rs`'s handling of a `use`-shaped `ClassBodyOp::Other`, and wherever module exports get installed into the current package/lexical scope) versus how a **method** body's compiled bytecode resolves a bare-name call — methods appear to resolve free function names through a different path than ordinary class-body statements or top-level code, and that path is evidently not consulting whatever the class-body `use` installed. Method-body compilation is a natural place to start (`compile_method_body` and friends in `src/compiler/decl_plan.rs`), comparing how it resolves a bare call name against how an ordinary top-level/class-body-`Other` statement's compiled chunk does it.

## Task

Make a class body's own `use`-imported subs resolve from within that class's own method bodies, matching rakudo. A regression test belongs in `t/modules/` (import/export mechanics), with the minimal two-file shape above (a real small exported sub, called from a method of the importing class) — no need to vendor `Template::Classic` itself for the test.

## Labels

`todo:deep` — this is method-body name resolution intersecting with class-body import installation, a cross-cutting concern between the compiler's method-body-compile pass and the class registration/import machinery, not a one-file compiler/VM slice; the two different failure symptoms for what appears to be one root cause also suggest the underlying mechanism needs to be understood before committing to a fix shape.

tokuhirom:
## Summary

`use Some::Module;` written directly inside a class body imports the module's exported subs into the class's own package, as expected — but a **method** of that same class could not call the imported sub: mutsu threw `Unknown function: <name>` even though the class-body-level statements themselves resolved it fine.

## Root cause

Method dispatch (`vm_method_dispatch.rs`) only anchors `current_package` to the receiver's class during a method call when the class has:
- its own directly declared `sub`s (`has_class_scoped_subs`, tracked by `run_class_body`'s per-statement walk),
- package-scoped `my` statics,
- a namespaced (`::`) name, or
- a role's deferred-body import (`package_has_deferred_use_imports`, #8646).

A plain, flat class whose only class-scoped routine came from an ordinary `use`/`import` statement in its own body matched none of those cases, so `current_package` stayed wherever the enclosing scope left it (typically `GLOBAL`) for the whole method call — and `bare_name_packages()` never looked at the class's own package, where the imported sub was actually registered (`import_module` targets `current_package()`, which is already the class while its body runs).

## Fix

`run_class_body` (`registration_class_body.rs`) now also treats a class-body `use`/`import` statement as installing class-scoped routines: after such a statement runs, any function newly registered under the class's own package is recorded the same way an explicitly declared `sub` already was, so `has_class_scoped_subs` sees it and method dispatch anchors `current_package` to the class correctly.

Found via the ecosystem distribution roulette in `Karabiner::CompModGenerator`, whose `Karabiner::Template` class does `unit class Template; use Template::Classic; method rule_generator($tmpl) { ... template(...) ... }`.

## Test plan

- [x] New regression test `t/oo/method/class-body-use-import-visible-in-method.t`, verified against the `raku` oracle first, and confirmed to reproduce the bug (fails with the exact "Unknown function: greet" from the issue) when the fix is reverted.
- [x] `cargo fmt --all`
- [x] `make lint` (default clippy, `jit`-off clippy, wasm32 clippy, rustdoc) — all green
- [x] `make test` (cargo test + full `prove -r t/`) — all green
- [x] `make roast` — green except `roast/S32-io/IO-Socket-Async.t` (exit 124, "planned 40 ran 17"), the documented container-only sandboxed-network exception in `docs/agent-environments.md`

Closes #8883

https://claude.ai/code/session_016zBajaz8xGswYuLRqcN14b

---
_Generated by [Claude Code](https://claude.ai/code/session_016zBajaz8xGswYuLRqcN14b)_


tokuhirom:
CI failure on `test-suites` (and the `test` aggregator that depends on it) is not from this PR's diff: the job aborts with

```
```

That's `scripts/battery-testsuite.sh` failing to fetch the vendored `Terminal::ANSI` battery's own upstream test fixtures from `git.sr.ht` — a transient external-host fetch failure unrelated to `registration_class_body.rs`/class-body `use` imports, which this PR touches. Re-running the failed jobs once.

---

PR #8930, opened 2026-09-21
refactor: remove redundant .clone() calls in src/compiler/ (clippy::redundant_clone)

### LINKED ISSUE #8909: `src/compiler/` clones AST/bytecode values clippy proves are never read again (clippy::redundant_clone, 52 hits)
## Repro

Not a Raku-level bug — this is a Rust code-quality finding from `cargo clippy`. Reproduce with:

```sh
cargo clippy --all-targets --message-format=json -- -W clippy::redundant_clone \
  > tmp/clippy-clone-sweep.json
```

then filter the JSON to `clippy::redundant_clone` messages whose primary span is under one of the files below (see `.agents/skills/clippy-clone-sweep/SKILL.md` for the parsing snippet and the full triage process). Read each `rendered` diagnostic and confirm the cloned value really is never used again before removing the `.clone()`.

## Root cause

Each hit is a `.clone()` whose result clippy proves is never read again — the pattern is almost always "build a value once, clone it into a second place, and the original binding was never going to be used a third time," so the clone can become a move. See `src/value/...` sibling issue for a verified representative example of this exact shape (`msg.clone()` where `msg` is a local `String`/AST node never touched again).

## Affected files

Snapshot from the 2026-09-20 sweep (`clippy::redundant_clone`, `cargo clippy --all-targets`, rustc 1.96.1). Line numbers are not included — they drift as the code changes; re-run the command above to get current locations.

Total: 52 hits, 10 files
- `src/compiler/helpers_phasers.rs` (18)
- `src/compiler/expr_call.rs` (10)
- `src/compiler/stmt.rs` (6)
- `src/compiler/expr_postfix.rs` (4)
- `src/compiler/helpers_call_args.rs` (4)
- `src/compiler/expr_closure.rs` (2)
- `src/compiler/expr_method.rs` (2)
- `src/compiler/expr_ops.rs` (2)
- `src/compiler/helpers_sub_body.rs` (2)
- `src/compiler/mod.rs` (2)

## Why it matters

Each hit is a real, avoidable allocation with zero behavior benefit — free cleanup, not a correctness bug. Filed as its own module-sized ticket per `.agents/skills/clippy-clone-sweep/SKILL.md` (rather than one 1,088-hit PR, or 226 one-file issues covering the whole 2026-09-20 sweep). Sibling tickets from the same sweep cover the rest of the codebase — search open `todo:ticket` issues whose title contains "clippy::redundant_clone" to find them.

tokuhirom:
## Summary

- `cargo clippy -W clippy::redundant_clone` flags 52 hits (26 unique call sites, doubled across the lib + test targets) in `src/compiler/` where a `.clone()`'d value is provably never read again — each is a real, avoidable allocation.
- Verified each hit by hand per `.agents/skills/clippy-clone-sweep/SKILL.md`: confirmed the cloned binding really has no later use in its scope before turning the clone into a move (or, for two `name.resolve().to_string()` chains, dropping the redundant `.to_string()` since `Symbol::resolve()` already returns an owned `String`).
- Files touched: `expr_call.rs`, `expr_closure.rs`, `expr_method.rs`, `expr_ops.rs`, `expr_postfix.rs`, `helpers_call_args.rs`, `helpers_phasers.rs`, `helpers_sub_body.rs`, `mod.rs`, `stmt.rs`.

## Test plan

- [x] `cargo build`
- [x] `cargo clippy --all-targets -- -W clippy::redundant_clone` scoped to `src/compiler/` — 0 hits (was 52)
- [x] `cargo fmt --all`
- [x] `make lint` (default clippy, clippy with `jit` off, wasm32 clippy, rustdoc) — green
- [x] `make test` — green
- [x] `make roast` — green

This is Rust-internal code-quality cleanup with no behavior change, so no new `t/` test is needed — the existing suites cover behavior.

Closes #8909

https://claude.ai/code/session_01VgMx8vJz52APnw7ibthK6J

---
_Generated by [Claude Code](https://claude.ai/code/session_01VgMx8vJz52APnw7ibthK6J)_


tokuhirom:
`test-check` failed on `runtime::function_table::tests::repeating_an_install_reuses_the_version_it_produced` (assertion `left != right` failed, both 81).

This is not caused by this PR — this branch's diff only touches `src/compiler/`, never `src/runtime/function_table.rs`. I confirmed the failure is pre-existing and deterministic (not a flake): running that single test in isolation (`cargo test --lib runtime::function_table::tests::repeating_an_install_reuses_the_version_it_produced -- --exact`) fails the same way both on this branch and on a clean checkout of `main` at `d4e7ead5`. It passes as part of a full `make test`/`cargo test` run, which is presumably why it hasn't shown up before — the bug depends on allocator address-reuse timing that a full parallel test run apparently doesn't hit, but an isolated run (closer to what `test-check` does) hits it every time.

Root cause: `FunctionTableTransitions::install` memoizes by `Arc::as_ptr(&def) as usize`, which aliases once the original `Arc` is dropped and its address gets reused by an unrelated later allocation. Filed as #8932 with the full analysis. No fix ported here since none exists yet and it needs a design decision on the right identity scheme, not a quick patch.

Re-running the failed job once now in case CI's own scheduling avoids the address-reuse window this time.

---

Issue #8807, opened 2026-09-19
A user class `is <builtin>` with `FALLBACK` routes an inherited native method to FALLBACK instead of dispatching it


tokuhirom:
## Summary

A user class that subclasses a built-in type (`class Foo is Range { ... }`) and defines a `FALLBACK` method incorrectly routes calls to the *built-in parent's own native methods* (methods not overridden, and not `handles`-delegated) into `FALLBACK`, instead of dispatching them to the inherited native implementation. Rakudo dispatches such a call normally through the MRO and never reaches `FALLBACK`.

Found while working [zef Math::Interval](https://raku.land/zef:librasteve/Math::Interval) 0.0.3 (locked on the ecosystem lock board, [tokuhirom/mutsu#7884](https://github.com/tokuhirom/mutsu/issues/7884)) — its `Interval` class is `is Range`, delegates a handful of methods via `handles <min max minmax bounds infinite raku gist fmt>`, and defines a catch-all `FALLBACK` that dies. Any Range method NOT in that `handles` list (e.g. `.excludes-min`/`.excludes-max`, used internally by `Range`-coercing code) hits `FALLBACK` under mutsu and dies, where rakudo resolves it via ordinary inheritance.

## Minimal repro

```raku
class MyRange is Range {
    has Range $!range is built handles <min max>;
    method FALLBACK($name) { die "FALLBACK: no method $name" }
}
my $r = MyRange.bless(range => (1..5));
say $r.excludes-min;
say $r.min;
```

**rakudo:**
```
False
1
```

**mutsu** (debug build, current `main`):
```
FALLBACK: no method excludes-min
  in sub FALLBACK at repro.raku line 3
  in block <unit> at repro.raku line 6
```

`.min` alone (delegated via `handles`) works fine in both; only a Range-native method NOT in the `handles` list is misrouted.

## Why this matters for Math::Interval

`Math::Interval`'s `Interval` class is:

```raku
class Interval is Range is export {
    has Range $!range is built
        handles <min max minmax bounds infinite raku gist fmt>;

    submethod TWEAK {
        my ($x1, $x2) = $!range.min, $!range.max;
        die "x1 <= x2 is required for Interval endpoints" unless $x1 <= $x2;
        $x1 += $!range.excludes-min;   # <-- dies under mutsu
        $x2 -= $!range.excludes-max;
        $!range = $x1.Numeric..$x2.Numeric;
    }
    ...
    method FALLBACK( $name ) {
[...]

Issue #8915, opened 2026-09-21
The remaining `src/runtime/` builtins, nqp:: ops and supporting modules clone values clippy proves are never read again (clippy::redundant_clone, 186 hits)


tokuhirom:
## Repro

Not a Raku-level bug — this is a Rust code-quality finding from `cargo clippy`. Reproduce with:

```sh
cargo clippy --all-targets --message-format=json -- -W clippy::redundant_clone \
  > tmp/clippy-clone-sweep.json
```

then filter the JSON to `clippy::redundant_clone` messages whose primary span is under one of the files below (see `.agents/skills/clippy-clone-sweep/SKILL.md` for the parsing snippet and the full triage process). Read each `rendered` diagnostic and confirm the cloned value really is never used again before removing the `.clone()`.

## Root cause

Each hit is a `.clone()` whose result clippy proves is never read again — the pattern is almost always "build a value once, clone it into a second place, and the original binding was never going to be used a third time," so the clone can become a move. See the sibling `src/value/` issue from this same sweep for a verified representative example of this exact shape. This ticket covers everything left over in `src/runtime/` after splitting out the `methods_*.rs` and registration/native/regex/types clusters into their own sibling tickets — `src/runtime/` alone was 498 hits, too large for one ticket.

## Affected files

Snapshot from the 2026-09-20 sweep (`clippy::redundant_clone`, `cargo clippy --all-targets`, rustc 1.96.1). Line numbers are not included — they drift as the code changes; re-run the command above to get current locations.

Total: 186 hits, 47 files
- `src/runtime/nqp_ops.rs` (12)
- `src/runtime/cstruct_layout.rs` (12)
- `src/runtime/builtins_atomic_shared.rs` (10)
- `src/runtime/react_whenever.rs` (10)
- `src/runtime/utils/errors.rs` (10)
- `src/runtime/nqp_ops_builtin.rs` (8)
- `src/runtime/builtins_control_flow.rs` (6)
- `src/runtime/native_methods/socket_async.rs` (6)
- `src/runtime/run_modules.rs` (6)
- `src/runtime/runtime_encoding.rs` (6)
- `src/runtime/seq_helpers/smart_match.rs` (6)
- `src/runtime/sequence.rs` (6)
- `src/runtime/system_introspect.rs` (6)
- `src/runtime/builtins.rs` (4)
- `src/runtime/builtins_atomic_cas.rs` (4)
- `src/runtime/builtins_lvalue.rs` (4)
[...]

PR #8931, opened 2026-09-21
fix(vm): cmp between a Range and an is Range instance compares structurally

### LINKED ISSUE #8814: `cmp` between a Range and an `is Range` instance answers `More` instead of comparing structurally
## Summary

Follow-up finding while verifying the fix for #8807 (PR #8812) against the real `Math::Interval` distribution (locked on the ecosystem board, #7884).

With the #8807 fix landed, `Math::Interval`'s `t/01-rop.rakutest` and `t/02-iop.rakutest` no longer crash (all planned tests now run), but most assertions still fail — `cmp` between an `is Range` instance and a plain `Range` does not compare structurally the way Rakudo does, so every `($a op $b) cmp (expected-range)) ~~ Same` assertion in the suite fails.

## Minimal repro (no Math::Interval needed)

```raku
class MyR is Range {
    has Range $!range is built handles <min max bounds infinite raku gist fmt>;
}
my $iv = MyR.bless(range => (3..6));
my $plain = 3..6;
say $iv.raku;      # 3..6 (both sides)
say $plain.raku;   # 3..6
say ($iv cmp $plain);
```

**rakudo:**
```
3..6
3..6
Same
```

**mutsu** (debug build, current `main` after #8812):
```
3..6
3..6
More
```

Both sides render identically via `.raku`/`.gist` (bounds/handles delegation works correctly, matching #8807's fix), but `cmp` treats the `Instance` receiver differently from a plain `ValueView::Range`, presumably falling back to a type/identity-based ordering instead of Range's own structural (min/max/excludes-*) comparison.

## Why this matters for Math::Interval

`Math::Interval`'s `infix:<+>`/`infix:<->`/`infix:<*>`/`infix:<\/>` overloads on `Range` return an `Interval` (an `is Range` instance). The test suite compares results against plain `Range` literals via `cmp ... ~~ Same`, e.g.:

```raku
ok (($r1 + $r2) cmp (3..6)) ~~ Same, 'R+R';
```

Under mutsu this now completes (previously it crashed per #8807) but reports `not ok`, because `($r1 + $r2)` (an `Interval`) `cmp`s "More" than `(3..6)` (a plain `Range`) despite holding the same bounds.

Verified against `raku`: 16/16 in `t/01-rop.rakutest` and all reachable assertions in `t/02-iop.rakutest` pass there; under mutsu (post-#8812) `01-rop.rakutest` is 7/16 and `02-iop.rakutest` fails from test 3 onward (`+I dies-ok`), all via this same `cmp` gap (plus, separately, `.Numeric` is not implemented for an `is Range` instance — `02-iop.rakutest` test 36 additionally dies with `Cannot resolve caller Numeric(Math::Interval::Interval:D: )`, which may be the same root cause — no native/inherited-dispatch fallback for `is Range`-instance coercion methods — or may need its own investigation).

## Where to look

- Wherever `cmp`/three-way comparison dispatches for a `ValueView::Instance` receiver whose MRO includes `Range` — likely needs the same kind of "consult the native `Range` comparison instead of a generic Instance fallback" treatment `dispatch_instance_and_fallback` just got for `excludes-min`/`excludes-max`/`infinite`/`is-int` in #8812, but for `cmp`/`Numeric`/other Range-derived coercions instead of the four boolean predicates.
- This is likely one instance of a broader pattern: an `is Range` (or other native-container-parent) subclass instance does not get Range's native comparison/coercion machinery for free the way real Rakudo attribute-inheritance does. Worth checking whether the same gap exists for `is Array`/`is Hash`/etc, though those have their own storage-delegation mechanisms (`delegates_to_array_storage`) that `Range` lacks entirely.

## Not yet attempted

No fix attempted here — filing per the "gap found while verifying an unrelated fix" workflow. `Math::Interval` is not fully green under mutsu yet even after #8812.

tokuhirom:
## Summary

Rakudo compares an `is Range` instance against a plain `Range` (or another such instance) structurally — by min, excludes-min, max, excludes-max — the same rules two native `Range` values already get from `range_cmp`. mutsu instead fell back to a generic type/identity-based ordering for an `Instance` receiver:

```raku
class MyR is Range {
    has Range $!range is built handles <min max bounds infinite raku gist fmt>;
}
my $iv = MyR.bless(range => (3..6));
my $plain = 3..6;
say ($iv cmp $plain);   # rakudo: Same; mutsu (before this fix): More
```

## Root cause / fix

`range_cmp`'s ordering logic is factored out into `range_cmp_parts`, which takes `(min, max, excludes-min, excludes-max)` directly instead of reading them off a native `Range` variant.

`cmp` now checks whether either side is an `is Range` instance (walking the MRO) and, if so, reads that side's four structural fields through method dispatch (`min`/`max`/`excludes-min`/`excludes-max`) — so a delegated (`handles`) or inherited accessor participates the same way a native Range's own fields do — then compares both sides with `range_cmp_parts`.

## Where it came from

Found verifying the `Math::Interval` distribution (locked on the ecosystem board, #7884) as a follow-up to #8807/#8812: with those landed, its `t/01-rop.rakutest`/`t/02-iop.rakutest` stopped crashing but still failed most assertions, all of which check an arithmetic overload's result (an `Interval`, an `is Range` subclass) against an expected plain `Range` via `cmp ... ~~ Same`.

## Testing

- New `t/oo/class/is-range-inherited-cmp-structural.t`, verified against `raku` first.
- `cargo fmt --all`
- `make lint` — green
- `make test` — green (48363 tests)
- `make roast` — the only red files (`roast/6.c/S32-io/file-tests.t`, `roast/S16-filehandles/filetest.t`, `roast/S16-io/eof.t`, `roast/S32-io/IO-Socket-Async.t`) are exactly the remote-container environment-only exception set documented in `docs/agent-environments.md` (uid-0 permission bits and sandboxed network), not caused by this change.

Closes #8814

https://claude.ai/code/session_01XWKjFvvrYqgrvoeW151G5d

---
_Generated by [Claude Code](https://claude.ai/code/session_01XWKjFvvrYqgrvoeW151G5d)_


tokuhirom:
`gc-stress-tap` failed on this PR's head (`099a5c73`), but it is not this PR's failure — this PR only touches `src/vm/vm_comparison_ops.rs`/`vm_comparison_order_ops.rs` and adds a test/news file, none of which are anywhere near the failing code.

The failure is `runtime::function_table::tests::repeating_an_install_reuses_the_version_it_produced`:
```
assertion `left != right` failed
  left: 81
 right: 81
```

Confirmed this reproduces identically on a clean checkout of `main` tip (`d4e7ead5`, no changes from this branch), deterministically (5/5 local runs, both with and without the job's `MUTSU_GC=on` env). Root cause: `FunctionTableTransitions::install` (`src/runtime/function_table.rs`) uses a raw `Arc::as_ptr` as a proxy for "definition identity" in its memo key, which is unsound once the original `Arc<FunctionDef>` is dropped and its allocation is reused by a later, unrelated `Arc::new` — exactly what the test's own sequence triggers. It passes under the full `make test` run (different allocation pattern) but fails reliably in isolation, which is presumably why PR #8915 (`cb4c6b39`, an unrelated `clippy::redundant_clone` sweep — verified it isn't the cause either) didn't catch it.

No fix ported here — this needs a real fix to the memo key's identity scheme, not a one-line patch, so I filed it separately as tokuhirom/mutsu#8934 rather than widening this PR. Re-running the failed job once to confirm before standing down.

---

Issue #8953, opened 2026-09-21
roast: S32-list/skip.t subtest 53 fails deterministically in a fresh container (pre-existing, unrelated to any local change)


tokuhirom:
## Summary

`roast/S32-list/skip.t` is on `roast-whitelist.txt` (line 1196), but subtest 53 (".skip-all and .push-all on slipping slippy iterators", specifically the "sink-all sinks all the values" / "push-all pushes all the values" checks inside it) fails deterministically in a fresh remote-container session, on unmodified `main` (commit `6707645d`, the tip at investigation time).

## Repro

```sh
MUTSU_FUDGE=1 prove -e 'target/release/mutsu' roast/S32-list/skip.t
```

```
    # Failed test 'sink-all sinks all the values'
    # at roast/S32-list/skip.t line 138
        # expected: '4'
        #      got: '0'
```

Confirmed via bisection (checked out `src/vm/vm_call_method_compiled_mut.rs` at `HEAD~2`, i.e. byte-identical to `main`, rebuilt release, ran the file) that this is **not** caused by any local change — it reproduces on stock `main`. Also confirmed deterministic: 5/5 runs failed identically (not a timing-sensitive flake).

## Where it was found

Found incidentally while validating `make roast` for a PR against #8880 (a `CallMethodMut` fast-dispatch-cache perf change, unrelated to this file — bisection ruled it out). Not in `flaky-tests.txt`, and not one of the four documented container-only exceptions in `docs/agent-environments.md` (`file-tests.t`, `filetest.t`, `eof.t`, `IO-Socket-Async.t`) — none of those are `uid 0` / sandboxed-network shaped like this one.

## Suspect area

The failing subtest builds a custom `class TimesIterator does Iterator` whose `!SET-SELF` binds `$!times := times` (an alias to the caller's `my $times = 0` lexical), then drives it through `Seq.new(...).map(&block).skip` with `NEXT` phasers in the block. `$times` ends up `0` instead of `4` after `.sink`, meaning `pull-one`'s `$!times++` mutations through the bound alias are not reaching the caller's lexical — looks like a lost write-back somewhere in the Slip/Iterator/`NEXT`-phaser + bound-attribute-alias interaction, not a full method-dispatch failure (`push-all` also fails the same way).

## Why filing rather than fixing now

[...]

Issue #8880, opened 2026-09-20
perf: a method call costs 22,528 instructions because it re-resolves the method by name every time


tokuhirom:
## The measurement

Per primitive operation, mutsu against rakudo (`benchmarks/micro/primitive-ops.py`):

| operation | mutsu ns | raku ns | ratio |
| --- | ---: | ---: | ---: |
| `$t = $t + 1` | 86.5 | 21.2 | 4x |
| `$a ~ $b` | 398.4 | 47.5 | 8x |
| `@a[3]` | 152.5 | 17.3 | 9x |
| `@a[3] = 1` | 177.0 | 18.6 | 10x |
| `@a.push(1); @a.pop` | 980.1 | 111.7 | 9x |
| **`f()`** | **794.2** | **5.8** | **137x** |
| **`$o.m()`** | **3052.8** | **9.7** | **315x** |
| **`$o.v`** (accessor) | **1054.0** | **11.4** | **93x** |
| `C.new(v => 1)` | 3119.9 | 154.0 | 20x |

Data operations are **already at 4-10x**. The gap is one mechanism, an order of magnitude worse than everything around it.

As instructions, 100,000-iteration loops with the bare loop skeleton subtracted:

* `$o.m()` — **22,528 instructions per call**
* `f()` — **7,212 instructions per call**

rakudo does a method call in roughly 30 cycles.

## Where they go

From the callgrind caller tree of a `$o.m()` loop, **per single call**:

| | per call |
| --- | ---: |
| `Symbol::intern` | **10** |
| `Symbol::resolve` (each one a heap allocation — it is `as_str().to_owned()` on an already-`&'static str`) | 3 |
| `String::clone` | 2 |
| `resolve_user_method` — a full by-name method resolution | **1** |
| `HashMap` clone | 1 |
| heap allocations, all sources | **~11** |

plus `Registry::class_mro_readonly` at 3.9% of the run (an MRO walk per call), `is_native_method`, and `LocalKey::with` at **9.6% of the whole program** — about 2,360 thread-local accesses per call.

mutsu resolves the method by name, from scratch, on every call. The receiver's class, the method name and the MRO are identical on iteration 100,000 and iteration 1, and none of it is remembered.

## Proposal

A **per-call-site inline cache** for method dispatch, keyed on the receiver's type id — what MoarVM spends a type-id comparison on.

[...]

PR #8954, opened 2026-09-21
fix: exclude sigilless params from the fast method-dispatch cache

tokuhirom:
## Summary (scope narrowed — see history below)

This PR started as a perf change for #8880 (wiring `fast_method_cache` into the `CallMethodMut` opcode entry, which had never consulted it). That wiring turned out to have more than one latent correctness gap once real-world code exercised it, and chasing each one individually wasn't converging on something safe to ship as a bounded slice — so **the wiring itself has been reverted**. `src/vm/vm_call_method_compiled_mut.rs` is back to its pre-PR content.

**What ships here instead** is one independently-valid bugfix that the investigation surfaced: `try_populate_fast_cache`'s eligibility gate (shared by the existing non-mut `CallMethod` fast-cache path) excludes an attributive parameter (`$!x`) but not a **sigilless raw-capture parameter** (`\x`), which binds the *argument's own container*, not a copy. A constructor shaped like:

```raku
method !SET-SELF (\v) { $!x := v; self }
method new       (\v) { self.bless!SET-SELF: v }
```

loses the alias from the second (cache-hit) call onward. This bug was already latent in the *existing, unmodified* `CallMethod` fast-cache — it was simply never exercised there, since a bareword-receiver constructor call (`Type.new: $x`) always compiles to `CallMethodMut`, which never reached the cache before this investigation started prototyping that wiring.

Fixed by excluding `pd.sigilless` params from the shared eligibility gate. Also hardened: the mut-path's `has_attr_aliases`-style check (not shipped, since the wiring is reverted, but worth noting for whoever revisits this) would need to scan attribute *values* (`is_container_ref`), not just declaration-time `__mutsu_attr_alias::` metadata, to catch a runtime-bound alias.

## Timeline (kept for anyone picking this up)

1. Wired `CallMethodMut` into `fast_method_cache` — measured -3.24% instructions / -3.57% per call on #8880's own repro.
2. CI caught a real regression: `roast/S32-list/skip.t` — root-caused to the sigilless-param gap above. I initially misdiagnosed this as pre-existing on `main` (a flawed local bisection method); corrected in the thread below.
3. Fixed the sigilless gap, re-verified `skip.t` clean, pushed.
4. CI's bundled-library gate then caught a **second**, different regression: `Cro::HTTP`'s `http-router.rakutest` (route-delegation tests returning stale/leaked values across calls) — a different latent gap in the same eligibility surface.
5. Rather than keep patching gaps one at a time with no way to bound how many more exist, reverted the `CallMethodMut` wiring itself and kept only the sigilless-param fix, which is safe on its own (it only narrows an existing, independently-exercised cache). Verified this resolves both `skip.t` and `Cro::HTTP`'s test.

**#8880 stays open, un-narrowed** — the general per-call-site inline cache it proposes (and the `CallMethodMut` wiring specifically) still needs a fuller audit of `try_populate_fast_cache`'s eligibility gate against the broader set of method-body shapes `CallMethodMut` exercises before it's safe to land.

## Testing

- `t/routines/signature/fast-cache-sigilless-param-alias.t` — regression coverage for the sigilless-param bug (constructor alias, the Iterator shape from `skip.t`, an unaffected plain-param constructor, and a mixed sigilless+plain param list).
- `cargo fmt --all`: no changes.
- `make lint`: pass.
- `make test`: pass (one `t/vm/frames/light-call-frame-writes-dedup.t` failure during a run with a concurrent battery-suite job was a load-induced wall-clock-ratio flake — confirmed via 5/5 clean re-runs in isolation, not a real regression).
- `make roast`: only the four documented container-only exceptions remain.
- `scripts/battery-testsuite.sh` (`Cro::HTTP`'s `http-router.rakutest`, 439/439): clean.

## Test plan

- [x] `cargo fmt --all` (no changes)
- [x] `make lint`
- [x] `make test`
- [x] `make roast` (only the documented container-only exceptions remain)
- [x] `scripts/battery-testsuite.sh` — `Cro::HTTP` clean
- [x] Regression test added and passing

https://claude.ai/code/session_016ddw41o5kwDKTX3hHgawhz


tokuhirom:
CI red on `jit-stress`/`jit-stress-roast`/`gc-stress-roast` — not this PR's.

Both `jit-stress-roast` and `gc-stress-roast` fail on exactly the same two things, neither touched by this diff:
- `roast/S32-io/IO-Socket-Async.t`, exit 124 ("planned 40 ran 17") — the documented container-network-sandbox timeout (`docs/agent-environments.md`).
- `roast/S32-list/skip.t` subtest 53 (".skip-all and .push-all on slipping slippy iterators") — a real, deterministic failure, but **pre-existing on `main`, not caused by this PR**. I bisected it locally: checked out `src/vm/vm_call_method_compiled_mut.rs` at this branch's parent commit (byte-identical to `main`), rebuilt, and it fails identically (5/5 runs, same subtest, same "expected 4, got 0"). Filed separately as #8953 with a root-cause lead (a bound attribute alias `$!times := times` inside a custom `Iterator`, driven through `Seq…map(&block).skip`, not observing its mutations from the caller side — looks like a lost write-back through the phaser/Slip pipeline, unrelated to method dispatch).

No fix exists yet for #8953 (root cause needs its own session). I don't have a live means to re-run a single job right now (the full workflow run is still in progress, so `rerun_failed_jobs` 403s until it completes) — but I already have two independent jobs on this exact commit reproducing the identical failure, plus 5/5 local reproductions on the unmodified parent commit, so I'm treating this as confirmed rather than spending the one re-run on a third confirmation. `jit-stress` is the aggregator job and is red only because `jit-stress-roast` is.

`test-suites` (TAP + roast on the release binary) is still running; will report back once it lands. This PR stays watched until green/mergeable or otherwise resolved.

---

tokuhirom:
**Correction to my earlier standing-down comment: `roast/S32-list/skip.t` was NOT pre-existing — it was a real regression from this PR's own diff.**

My original local bisection (swapping `vm_call_method_compiled_mut.rs`'s content back to the pre-PR version in place, then rebuilding) gave a false negative — it still failed, which I wrongly took as proof `main` was already broken. A proper re-check with a clean `git worktree` of `origin/main` built from scratch showed `main` passes `skip.t` cleanly; only my branch, built the same clean way, failed it. The in-place swap-and-rebuild wasn't a trustworthy substitute for a real clean build.

**Root cause, confirmed via direct minimal repro + tracing:** `try_populate_fast_cache`'s eligibility gate (shared by both `CallMethod` and the newly-wired `CallMethodMut` entry) excludes an attributive parameter (`$!x`, which mutates `self`) but not a *sigilless raw-capture* parameter (`\x`), which binds the **argument's own container**, not a copy. A constructor shaped like:

```raku
method !SET-SELF (\v) { $!x := v; self }
method new       (\v) { self.bless!SET-SELF: v }
```

called repeatedly on a bareword/type-object receiver (always `CallMethodMut`) loses the alias from the second (cache-hit) call onward. This gap was already latent in the existing non-mut `fast_method_cache` — it was just never exercised there, since a bareword-receiver constructor call always compiles to `CallMethodMut`.

**Fix pushed** (`39749601`): exclude `pd.sigilless` params from the shared eligibility gate (fixes both entries at once), plus hardened the mut-path's `has_attr_aliases` check to scan attribute *values* (`is_container_ref`) in addition to the existing declaration-time metadata scan. Verified: direct repro fixed, `roast/S32-list/skip.t` 5/5 clean runs, `cargo fmt`/`make lint`/`make test`/`make roast` all clean (only the four documented container-only exceptions remain), plus a new regression test (`t/routines/signature/fast-cache-sigilless-param-alias.t`).

I've also corrected and closed #8953, which I filed based on the same false-negative bisection.

CI re-running now on the corrected head.

---

PR #8992, opened 2026-09-21
fix: reject undeclared barewords in required module mainlines

### LINKED ISSUE #8986: fix: reject undeclared barewords in required module mainlines
## Finding

A required module's top-level unknown bareword is silently treated as a string by mutsu, so `require` succeeds when Rakudo raises a compile-time undeclared-name error. This causes plugin discovery to return a module whose source did not compile.

## Reduction

From the `Pluggable` 0.6 ecosystem suite (`t/04-syntax-error.t`), with `t/lib` on the module path:

```raku
use lib 't/lib';
use Pluggable;

our $*DEBUG-PLUGINS = 1;
my @plugins = plugins('CaseC');
```

`t/lib/CaseC/Plugins/Class1.pm6` contains:

```raku
unit class CaseA::Plugins::Class1;

dead
```

Rakudo returns no plugins because loading the file reports:

```text
Undeclared routine:
    dead used at line 7
```

mutsu treats the bareword `dead` as a string, `require ::($name)` returns successfully, and `plugins('CaseC')` returns one `CaseC::Plugins::Class1` type. The failure is general; it is not specific to `Pluggable`.

## Expected scope

Make the mainline undeclared-name check used for required modules reject an unknown bareword while preserving valid types, imported symbols, constants, and intentional bareword contexts. Pin the behavior with a focused `t/` regression test.

tokuhirom:
## Summary

- A required module's top-level unknown bareword (e.g. a lone `dead` statement) parses as `Expr::BareWord`, which the CHECK-time undeclared-routine walker (`src/runtime/undeclared_routines.rs`) never inspected — only `Stmt::Call`/`Expr::Call` were recorded. The runtime's own bareword resolution then silently fell back to treating the name as a `Str`, so `require` on a module whose source does not compile still succeeded, unlike rakudo's compile-time `X::Undeclared::Symbols`.
- Treat a bare lowercase identifier used as a whole statement (`Stmt::Expr(Expr::BareWord(name))`) the same way `record_call` already treats a 0-arg call, excluding `self` (the one common legitimate bare statement this form produces — a method's own bare `self`).
- Wire the existing `check_undeclared_routines_mainline` check into both module-loading paths: `require_load_from_file` (path-like `require`) and `load_module_inner` (`use`, and a bareword-named `require ::($mod)`, which reaches `use_module`/`load_module_inner` rather than `require_load_from_file` since it isn't path-like).

## Test plan

- New regression test `t/modules/import-export/require-undeclared-bareword.t` + fixture `t/lib/RequireUndeclaredBareword.rakumod` (mirrors the `Pluggable` ecosystem reduction from the issue), covering both direct `require Name` and indirect `require ::("Name")`, plus a sanity check that a well-formed required module still loads.
- `cargo fmt --all` — no changes.
- `make test` — PASS (4667 files, 48603 tests).
- `make roast` — the only failures are the four container-only entries documented in `docs/agent-environments.md` (`file-tests.t`, `filetest.t`, `eof.t`, `IO-Socket-Async.t`, all `uid 0` / sandboxed-network artifacts of this remote container), matching their documented shapes exactly. No other file failed.

Closes #8986

https://claude.ai/code/session_01NPqVQpHPmjAP2v7eW1aV35

---
_Generated by [Claude Code](https://claude.ai/code/session_01NPqVQpHPmjAP2v7eW1aV35)_


tokuhirom:
`test-check` failed, but not on this PR's code: the failure is entirely in the `rui314/setup-mold` step, which retried downloading `mold-2.42.0-x86_64-linux.tar.gz` from `github.com/rui314/mold/releases` 20 times and got `504 Gateway Time-out` every time before giving up — a GitHub Releases CDN blip, unrelated to this diff. No test in this PR's suite ran before the job aborted.

Will re-run the failed job once the rest of the workflow run finishes (GitHub refuses a rerun while sibling jobs are still `in_progress`).

---

Issue #8997, opened 2026-09-21
perf: compile static reduction, hyper, and meta operators into typed bytecode


tokuhirom:
## Finding

`Reduction`, `HyperOp`, and `MetaOp` encode their statically parsed operator spelling as a string-constant index. Each execution reads that string and reinterprets it with byte/string operations.

The compiler emits `Value::str(op)` into these opcodes:

- `src/compiler/expr.rs` for reductions;
- `src/compiler/expr_ops.rs` for hyper and meta operators.

The VM then allocates a `String` and reparses the spelling:

- `src/vm/vm_misc_reduction_exec.rs`: scan/negation prefixes, `R`, Unicode aliases, the known-base-operator table, associativity and special cases;
- `src/vm/vm_hyper_ops.rs`: leaf `=` and `~~` handling;
- `src/vm/vm_meta_ops.rs` and `src/vm/vm_dispatch_helpers.rs`: `R`/`Z` meta forms, aliases, and builtin-family dispatch.

This is not needed for source operators: their token and meta-operator structure are fixed when the AST is parsed. It also leaves string comparisons scattered across VM opcode implementations.

## Direction

Introduce a compiler-owned typed representation, for example `CompiledOperator`:

- builtin canonical operators as enum variants;
- structural meta-operators (`Reduction`, reverse, cross, zip, hyper and scan/negation flags) represented structurally;
- `Custom(Symbol)` for arbitrary user-defined infix operators.

Have `Reduction`, `HyperOp`, and `MetaOp` carry that representation (or an index into a typed per-chunk table), rather than a `Value::Str` constant index. Canonicalize Unicode aliases in the compiler. The VM should match the typed operator and use its `Symbol` only for a custom/user dispatch lookup.

Dynamic language semantics remain intact: a user-defined operator is still looked up at runtime. Only repeatedly parsing a source-fixed spelling moves out of the execution loop.

## Guardrail

Make construction of this representation private to the compiler/operator-lowering module. A VM opcode should not accept a raw string operator operand, so adding a new static operator path requires extending the typed lowering first.

## Acceptance evidence

Add focused bytecode/compiler tests covering builtin canonicalization, meta composition, Unicode aliases, and an unknown custom infix represented as `Custom(Symbol)`. Preserve the existing reduction/hyper/meta TAP behaviour. Profile before reporting a speedup; this is a `todo:perf` finding.

tokuhirom:
[...]

PR #9029, opened 2026-09-22
perf: compile static reduction, hyper and meta operators into typed bytecode

tokuhirom:
Refs #8997. First slice: the *structural* half of the `CompiledOperator` direction.

## The problem

`[+]`, `[\!after]`, `>>+<<` and `@a Z+ @b` all have a shape the parser fixed once. The compiler flattened that shape back into a string in the constant pool, and the VM re-derived it on **every execution** of the opcode:

- allocate a `String` from the pooled constant (`Self::const_str(code, op_idx).to_string()`);
- strip the `\` scan marker, then the `!` negation marker behind a linear scan of an 80-entry table, then the compiler's own `_sc_` thunked-short-circuit marker;
- fold the Unicode operator aliases (`×` → `*`, `≤` → `<=`, …);
- strip the statically decidable `R` (reverse) prefixes, allocating another `String` per strip;
- `format!("infix:<{}>", op)` up to **four** separate times, across `reduction_op_associativity` and `reduction_callable_for_op`.

## The change

New `src/compiled_operator.rs` owns the decode, at compile time:

- **`ReductionSpec::lower`** produces `{ scan, negate, shortcircuit, reverse, base }` once. `OpCode::Reduction`'s operand now indexes a per-chunk `CompiledCode::reduction_specs` table instead of the constant pool.
- **`MetaKind`** replaces the pooled `"reduce"` / `"R"` / `"X"` / `"Z"` string that `OpCode::MetaOp` (and its assignment / n-ary siblings) matched as text. The base operator travels as an interned `Symbol`, as it does for `OpCode::HyperOp`.
- **`infix_names`** memoizes the `infix:<op>` / `&infix:<op>` / `&op` spellings an operator resolves through — keyed by the *spelling*, not by an interned `Symbol`, because interning the lookup key would itself be the per-execution work the memo exists to remove. (Measured: a `Symbol`-keyed first draft *added* one intern per execution.)

Dynamic semantics are unchanged. A user-defined infix is still resolved at run time, and a `[R&callable]` inner still has its `R` stripped by the VM, because whether `&foo` names a callable is a runtime question the compiler cannot settle. `ReductionSpec::lower` applies its markers in exactly the order the VM used to, so the decoded shape is identical to what each execution used to compute.

The change also collapses **four byte-identical copies** of the hyper-delimiter stripper (`>>op<<` / `»op»`) — in the parser, in `runtime/utils/type_misc.rs`, and twice in the VM — and two copies of the Unicode alias table, into one each. Net **−152 lines**.

## Guardrail

A reduction, hyper or meta opcode no longer accepts a constant-pool index for its operator, and `ReductionSpec`'s fields are decided only by `ReductionSpec::lower`. A new statically spelled operator path therefore cannot be added by emitting one more string for the VM to re-parse — it has to extend the lowering first.

## Measurements

A/B over 2,400 operator executions (400 iterations of `[+] @l`, `[\+] @l`, `[R-] @l`, `@l >>+<< @r`, `@l Z+ @r`, `@l X+ (1, 2)` on eight-element arrays). Both sides built from scratch with `--profile profiling` and measured under callgrind twice, quoting the second run so the module precompilation cache is warm on both (perf-tuning skill §0). "Before" is `origin/main` at `afa2058a`.

| | instructions | allocations |
| --- | ---: | ---: |
| before | 71,261,456 | 62,424 |
| after | 71,120,031 | 57,226 |
| | **−0.20%** | **−8.3%** |

That is ~2.2 heap allocations and ~59 instructions removed **per operator execution**, and the allocation figure is fully accounted for by the `String` and `format!` sites removed.

**This will not move a stopwatch, and I am not claiming it does.** −0.20% is an order of magnitude below the ~2% a paired wall-clock run resolves on this box. The honest claim is that these opcodes do strictly less work per execution, with the deterministic counts as the evidence. (`make check-bench-det` self-test passes, so the allocation extractor is trustworthy.)

## Tests

- `src/compiled_operator/tests.rs` — 11 unit tests covering exactly the issue's acceptance list: builtin canonicalization, meta composition (`\!eqv`, `\R-`), Unicode aliases, `R` parity including the undecidable `R&foo`, the `_sc_` form, and an unknown custom infix staying a `Symbol` leaf.
- `tests/static_operator_intern_budget.rs` — a looped `[+]`, `>>+<<` and `Z+`, plus a `[+]` fold over a growing list, each add **0.000** `Symbol::intern` calls per execution and per element.
- Oracle diff against `raku` over 60 reduction / hyper / meta forms: identical. The one divergence (`[Rmyop]` over a user-defined infix) reproduces on `main` and is a pre-existing gap, untouched here.

## Verification

- `cargo fmt --all`, `make lint` (all four configurations) — clean.
- `make test` — exit 0, 4687 files / 48,864 tests.
- `make roast` — red on exactly the four container-only files documented in `docs/agent-environments.md`, matched by name **and** by exact subtest number: `6.c/S32-io/file-tests.t` (6-8, 10), `S16-filehandles/filetest.t` (57-64, 69-72, 77-80, 101/103/105/107/109/111/117/121/125), `S16-io/eof.t` (exit 255), `S32-io/IO-Socket-Async.t` (exit 124). Nothing else fails.
- `roast/S03-metaops/` — all 9 files, 6,365 tests, PASS.

## Not closing #8997

Per the perf-tuning skill, a slice does not close a perf issue. Two parts of the issue's direction are still open and are the larger half:

1. The builtin operators themselves are not enum variants yet — the VM still compares `base_op` against string literals. The leaf here is a canonical `Symbol`, which is `Custom(Symbol)`'s shape for every operator rather than just the user-defined ones.
2. **The bigger prize:** `eval_reduction_operator_values` still re-strips `[`, `R`, `Z` and the hyper delimiters, and re-folds the Unicode aliases, for **every element** of a hyper or zip. This PR removes the per-*execution* decode; the per-*element* decode needs a recursive typed operator threaded through those loops.

Also recorded in the news entry: a user-defined infix reduction still pays ~7.6 interns per fold step inside `call_user_routine_direct`'s by-name dispatch, both before and after this change. That is #7766's subject, not this one.

https://claude.ai/code/session_014qc57wZHs1R54vwxb5UiyY

---
_Generated by [Claude Code](https://claude.ai/code/session_014qc57wZHs1R54vwxb5UiyY)_


tokuhirom:
## `test-suites` failed on an external host, not on this diff

[`test-suites`](https://github.com/tokuhirom/mutsu/actions/runs/35690341925/job/106626911006) (and therefore its `test` aggregator) went red on the ecosystem battery gate:

```
...
```

The five `REGRESSION: whitelisted 'Terminal::ANSI ...'` lines are all downstream of that one fetch: `scripts/battery-testsuite.sh` records `SETUP_FAILED` when a battery's test suite cannot be checked out, and every file in that dist then counts as not-passed. The rest of the gate was fine — `301/318 test files pass (3 excluded)`, and no other dist regressed.

**Why this is not this PR's.** The diff is entirely Rust operator lowering (`src/compiled_operator.rs`, the reduction/hyper/meta opcodes and their compiler emission). It changes nothing about network access, vendoring, or the battery gate, and there is no mechanism by which it could affect a `git fetch` from sourcehut.

Three independent confirmations:

1. **`main` is green across this window.** `f36eb2c7` — this PR's base — passed at 05:03–05:07, as did `afa2058a`, `05d01a18`, `d9067e2e` and the runs before them. This PR's run hit the fetch error at 05:35.
2. **The commit is fetchable.** `git ls-remote https://git.sr.ht/~bduggan/raku-terminal-ansi` just now returns `691d0959ff8ea0a7c491167b366746cdfc0fb0be  refs/heads/master` — the exact commit the gate asked for. So the ref is intact upstream; sourcehut was briefly unreachable from the runner.
3. **The failure names a service the diff does not touch**, which is the documented signature of a non-PR failure.

**No fix exists to port**, because nothing is broken: the pin is valid and the host is back. Re-running the failed jobs once now, which is my single permitted re-run for this. If it fails a second time I will treat it as real and root-cause it rather than re-running again.

Local verification on this head (`d8a1d1f`) for the record: `make lint` clean across all four configurations, `make test` exit 0 (4,687 files / 48,864 tests), and `make roast` red on only the four `uid 0` / sandboxed-network files that `docs/agent-environments.md` documents as container-only, matched by name and by exact subtest number.

---

